PR #879, opened 2026-06-06
Wire feed-processing path through plugin feed capability (#875)

### LINKED ISSUE #875: Plugin system's `feed` capability is dead code; feed paths hardcode LessWrong
## Problem

The plugin system declares a `feed` capability, but nothing in the feed-processing path actually uses it — the real code branches on LessWrong directly. Only the `savedArticle` capability is wired up.

Flagged during architecture review; verify against current `master`.

### Evidence

- `src/server/plugins/lesswrong.ts` (~L31–75) defines `transformToFeedUrl` / `cleanEntryContent` / `transformFeedTitle`, but nothing calls them.
- The real feed paths hardcode LessWrong instead:
  - `isLessWrongFeed` check in `content-utils.ts` (~L72)
  - inline title append in `handlers.ts` (~L377–382)
  - `feeds.ts` (~L133, L253–410)
- The `savedArticle` capability **is** wired (`full-content.ts` ~L45, `saved.ts` ~L320), so the plugin abstraction works for that half.

## Why it matters

The registry/capability abstraction is half-built. Adding a new feed source means editing core files rather than writing a plugin — defeating the plugin system's purpose — and the LessWrong-specific logic is duplicated across several core modules that must be kept in sync by hand.

## Suggested fix

Either:
- Wire the feed-processing path through the plugin `feed` capability (resolve the plugin from the registry and call `transformToFeedUrl` / `cleanEntryContent` / `transformFeedTitle`), removing the hardcoded LessWrong branches; or
- If a full plugin abstraction for feeds isn't wanted, delete the unused `feed`-capability methods and document that feed-source customization is intentionally hardcoded.

Found during a multi-agent architecture review.

— Claude

brendanlong:
## Summary
Fixes #875. The plugin system declared a `feed` capability (`transformToFeedUrl` / `cleanEntryContent` / `transformFeedTitle`) but nothing called it — the real feed paths branched on LessWrong directly, duplicating logic across `content-utils.ts`, `feeds.ts`, and `handlers.ts`. Only the `savedArticle` half was wired.

This wires the feed path through the registry so adding a feed source means writing a plugin, not editing core modules.

- Add `getFeedPlugin(url)` to `src/server/plugins/index.ts` — resolves the feed-capable plugin from a URL string (null on invalid/unhandled).
- Move the full URL→feed mapping (frontpage / shortform / user / post) into the LessWrong plugin's `transformToFeedUrl`; the feeds router now calls the plugin.
- `matchUrl` returns `true` — the host index already scopes the plugin to LessWrong hosts, and each capability validates its own URL shape.
- Change `transformFeedTitle` from an async GraphQL lookup to a synchronous function using `FeedTitleContext.firstAuthor` (first author from already-parsed feed items). This matches the existing production behavior in `handlers.ts` and avoids a network call in the feed-processing hot path.
- Route `content-utils.ts`, `feeds.ts` (cleaning + URL transform + preview title), and `handlers.ts` (title append) through the registry, removing the hardcoded LessWrong branches.
- Remove the now-dead `isLessWrongFeed` and its unit-test block.
- Update `docs/plugin-system-design.md` to reflect the completed wiring.

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm knip`
- [x] `pnpm vitest run tests/unit` (1703 passed; added `tests/unit/lesswrong-plugin.test.ts`, 12 tests)
- [ ] Integration tests (`entry-processor`, `feeds`) require a Docker DB — not run here; changes are behavior-preserving

— Claude

### DIFF
--- a/docs/plugin-system-design.md
+++ b/docs/plugin-system-design.md
@@ -1,477 +1,84 @@
 # URL Plugin System Design
 
+This document explains _why_ the plugin system is shaped the way it is. For the
+actual interfaces and implementations, follow the links to the source — the code
+is the source of truth, this doc is the rationale.
+
 ## Overview
 
-A plugin system to consolidate custom parsing logic for feeds, entries, and saved articles. Plugins register URL patterns and provide capabilities for different use cases.
-
-## Goals
-
-1. **Consolidate scattered URL detection** - Replace `isLessWrongUrl()`, `isGoogleDocsUrl()` checks throughout the codebase
-2. **Unified interface** - All plugins implement the same interface
-3. **Easy to add new plugins** - Adding ArXiv, Twitter, etc. should be self-contained
-4. **Fallback behavior** - If plugin fetch fails, fall back to normal fetching
-5. **Capability-based lookup** - Find first plugin that matches URL AND has the requested capability
-
-## Plugin Interface
-
-```typescript
-// src/server/plugins/types.ts
-
-export interface UrlPlugin {
-  /** Unique identifier for the plugin */
-  name: string;
-
-  /** Hostnames this plugin handles (for O(1) lookup) */
-  hosts: string[];
-
-  /**
-   * More specific URL matching after hostname match.
-   * Return true if this plugin should handle the URL.
-   */
-  matchUrl(url: URL): boolean;
-
-  /** Plugin capabilities */
-  capabilities: PluginCapabilities;
-}
-
-export interface PluginCapabilities {
-  feed?: FeedCapability;
-  entry?: EntryCapability;
-  savedArticle?: SavedArticleCapability;
-}
-
-// ============ Feed Capability ============
-
-export interface FeedCapability {
-  /**
-   * Transform a URL into a feed URL.
-   * E.g., LessWrong user profile → RSS feed URL
-   * Return null if URL can't be transformed.
-   */
-  transformToFeedUrl?(url: URL): Promise<URL | null>;
-
-  /**
-   * Clean entry content from this feed.
-   * E.g., strip "Published on..." prefix from LessWrong.
-   * Runs synchronously during feed processing.
-   */
-  cleanEntryContent?(html: string): string;
-
-  /**
-   * Transform feed title after parsing.
-   * E.g., append author name to LessWrong user feeds.
-   * Synchronous: uses already-parsed feed data (context.firstAuthor) to avoid
-   * extra network calls during feed processing.
-   */
-  transformFeedTitle?(title: string, feedUrl: URL, context: FeedTitleContext): string;
-
-  /**
-   * Site name to use for entries from this feed.
-   * Falls back to feed's own site name if not provided.
-   */
-  siteName?: string;
-}
-
-export interface FeedTitleContext {
-  /** First author found among the feed's parsed entries. */
-  firstAuthor?: string | null;
-}
-
-// ============ Entry Capability ============
-
-export interface EntryCapability {
-  /**
-   * Fetch full content for an entry URL.
-   * Used when RSS provides only excerpts.
-   * Return null to fall back to normal fetching.
-   */
-  fetchFullContent?(url: URL): Promise<EntryContent | null>;
-}
-
-export interface EntryContent {
-  html: string;
-  title?: string;
-  author?: string;
-  publishedAt?: Date;
-}
-
-// ============ Saved Article Capability ============
-
-export interface SavedArticleCapability {
-  /**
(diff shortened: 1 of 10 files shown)


docs/plugin-system-design.md
```diff
@@ -60,8 +60,10 @@ export interface FeedCapability {
   /**
```
brendanlong: It looks like this file has a full copy of the plugin code. We should probably remove that and just leave any references to why we wrote the code this way + links to the actual code.

src/server/plugins/lesswrong.ts
```diff
@@ -20,55 +31,85 @@ export const lessWrongPlugin: UrlPlugin = {
...
 
-  matchUrl(url: URL): boolean {
-    // Match posts (/posts/[id]), comments (?commentId=), and users (/users/[slug])
-    return /^\/(posts|users)\//.test(url.pathname) || url.searchParams.has("commentId");
+  // The host index already restricts this plugin to LessWrong hosts, and it
+  // handles every LessWrong URL: feed.xml feeds (clean/title), pages (transform
+  // to feed), and posts/comments (saved-article fetch). Each capability validates
+  // the specific URL shape it cares about, so match all LessWrong URLs here.
+  matchUrl(): boolean {
+    return true;
```
brendanlong: I think the assumption here is incorrect. There are pages on LessWrong that we don't know how to map to the GraphQL API. Any unknown pages should skip the plugin and be fetched normally.

src/server/plugins/index.ts
```diff
@@ -22,6 +22,26 @@ logger.info("Plugins registered", {
...
+ */
+export function getFeedPlugin(url: string | URL | null | undefined) {
+  if (!url) return null;
+
+  let parsed: URL;
+  try {
+    parsed = typeof url === "string" ? new URL(url) : url;
+  } catch {
+    return null;
+  }
```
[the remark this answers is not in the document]
brendanlong: Done in 2b0d6c0a — switched to `url instanceof URL ? url : new URL(url)` so unexpected non-string inputs go through `new URL()` and are caught.

— Claude

src/server/plugins/types.ts
```diff
@@ -55,8 +55,10 @@ export interface FeedCapability {
   /**
    * Transform feed title after parsing.
    * E.g., append author name to LessWrong user feeds.
+   * Runs synchronously during feed processing, using already-parsed feed data
+   * (passed via context) to avoid extra network calls.
    */
-  transformFeedTitle?(title: string, feedUrl: URL): Promise<string>;
+  transformFeedTitle?(title: string, feedUrl: URL, context: FeedTitleContext): string;
```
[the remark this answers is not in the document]
brendanlong: Done in 2b0d6c0a — `context` is now optional in the interface, and the LessWrong implementation defaults it to `{}`.

— Claude

src/server/plugins/lesswrong.ts
```diff
@@ -20,55 +31,85 @@ export const lessWrongPlugin: UrlPlugin = {
...
+        if (!isLessWrongUserFeedUrl(feedUrl.href)) return title;
 
-          return `${title} - ${user.displayName}`;
-        } catch (error) {
-          logger.warn("Failed to transform LessWrong feed title", {
-            feedUrl: feedUrl.href,
-            error: error instanceof Error ? error.message : String(error),
-          });
-          return title;
+        const { firstAuthor } = context;
```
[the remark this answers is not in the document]
brendanlong: Done in 2b0d6c0a — `context: FeedTitleContext = {}`.

— Claude

src/server/plugins/lesswrong.ts
brendanlong: Good catch, you're right. Fixed in 2b0d6c0a — `matchUrl` now only returns `true` for URLs the plugin actually knows how to handle: `feed.xml` feeds, the front page, quicktakes, user profiles, and posts/comments. Unknown pages (e.g. `/tag/...`, `/library`) return `false` and get fetched normally. Added a unit test asserting unknown pages don't match.

— Claude

docs/plugin-system-design.md
brendanlong: Done in 2b0d6c0a — rewrote the doc to drop the full code copies. It's now design rationale (why `matchUrl` must be selective, why `transformFeedTitle` is sync, the fallback contract) plus a table of integration points with links to the actual source files. Cut ~465 lines of duplicated code.

— Claude

PR #901, opened 2026-06-10
Block SSRF to private networks in server-side fetches (#826)

### LINKED ISSUE #826: feeds.preview/discover endpoints allow SSRF to internal networks
## Problem

The `feeds.preview` and `feeds.discover` tRPC endpoints accept user-supplied URLs and fetch them server-side without blocking private/internal IP addresses. This could be used to:

- Probe internal infrastructure (e.g., `http://169.254.169.254/` for cloud metadata)
- Scan internal network services
- Access localhost services

These endpoints use the `expensivePublicProcedure` middleware (rate-limited but no auth required).

## Suggested fix

Add URL validation that blocks private/reserved IP ranges before making the fetch request. Either:

1. Resolve the hostname and check against private IP ranges (10.x, 172.16-31.x, 192.168.x, 127.x, 169.254.x, ::1, fc00::/7, etc.)
2. Use a library like `ssrf-req-filter` or implement DNS rebinding protection

The check should happen after DNS resolution but before connecting, to prevent DNS rebinding attacks.

**Severity:** Medium — endpoints are public and rate-limited but allow probing.

Found during code review in #823.

— Claude

brendanlong:
## Summary

Fixes #826. The `feeds.preview`/`feeds.discover` endpoints (and feed fetching, full-content fetching, and WebSub hub callbacks) fetched user-supplied URLs server-side without blocking private/internal IP ranges, allowing probes of localhost, internal services, and cloud metadata (`169.254.169.254`).

- New shared helper `withSsrfProtection(url, init)` in `src/server/http/ssrf.ts`:
  1. **Synchronous** rejection of literal private/reserved IP hosts (e.g. `http://169.254.169.254/`, `http://127.0.0.1/`, decimal-encoded IPs) — undici skips the custom DNS lookup for IP literals, so they're checked here.
  2. **Connection-time** validation via a custom undici dispatcher whose DNS `lookup` blocks if *any* resolved address is private and connects only to the vetted IP — closing the DNS-rebinding TOCTOU gap and covering redirects (fetch reuses the dispatcher).
- Wired into `fetchUrl`/`fetchHtmlPage`, `fetchFeed`, the discovery `tryFetchAsFeed`, and the WebSub hub POST, so preview/discover, background fetches, full-content, and WebSub are all protected.
- Blocks loopback, RFC 1918, CGNAT, link-local (incl. cloud metadata), documentation/test, multicast, and reserved ranges for IPv4 and IPv6 (and IPv4-mapped IPv6).
- `ALLOW_PRIVATE_NETWORK_FETCH=true` disables the block for dev/test (default in `.env.test`).
- Added `undici` as an explicit dependency (already present transitively) for the typed `Agent` import.

## Test plan

- [x] `pnpm typecheck`, `pnpm lint`, `pnpm knip` (no new violations)
- [x] `pnpm test:unit` — added `tests/unit/ssrf.test.ts` covering IPv4/IPv6 private vs public classification, IPv4-mapped IPv6, and fail-closed on non-IP input
- [x] Runtime verification: literal private IPs, IPv6 `[::1]`, decimal-encoded IPs, and `localhost` (via DNS) are all blocked; `example.com` is allowed; flag-on path attaches no dispatcher
- [ ] Confirm no production feeds break (public feeds unaffected)


brendanlong:
Thanks for the review. Addressed in 482b5385:

**Gemini (critical) — IPv4-mapped/compatible IPv6 bypass:** Real bug, fixed. The literal-IP check only matched dotted-decimal mapped form (`::ffff:127.0.0.1`), so hex-mapped (`::ffff:7f00:1`) and IPv4-compatible (`::7f00:1`) encodings of private addresses slipped through. Rather than hand-roll an IPv6 expander, I added IPv4-mapped subnets (`::ffff:<range>/(96+prefix)`) for every private IPv4 range plus `::/96` (covers IPv4-compatible + `::`/`::1`) to the `BlockList`. `BlockList` matches numerically, so it's agnostic to textual encoding (hex, compressed, dotted) — verified against all the suggested vectors. Dropped the fragile regex.

**Gemini (medium) — tests:** Added regression cases for hex-mapped and IPv4-compatible forms (both blocked-private and an allowed public mapped `::ffff:808:808`).

**Gemini (high) — `ssrfLookup` options:** No change. `LookupOptions` is typed as an object and undici always passes one, so `options.all` can't throw. On the family concern: I spread `{ ...options, all: true }` into `dnsLookup`, which *preserves* `family`/`hints` (the suggested rewrite manually re-extracts only those two and is no safer), and `dnsLookup` then only returns addresses of the requested family, so the single-address return path stays correct.

**CodeQL — SSRF in `websub.ts`:** This is a static-analysis false positive. The `fetch(feed.hubUrl, …)` call is wrapped in `withSsrfProtection(feed.hubUrl, …)`, which CodeQL doesn't model as a sanitizer. The hub URL is validated the same way as every other outbound fetch.

— Claude

PR #905, opened 2026-06-12
Add feed fetch health monitoring via healthchecks.io dead-man's-switch

brendanlong:
## Problem

The "Unknown feed format" outage (#904) went unnoticed for ~2 days because nothing watches whether feed fetching works *globally* — per-feed failure tracking exists (Feed Status page), but you only see it when you look.

## Design

A new `monitor_feed_health` singleton job (same mechanism as `renew_websub`) runs in the worker every 15 minutes and enforces the invariant:

> **At least one feed must fetch successfully every N minutes** (default 120, `FEED_HEALTH_MAX_SUCCESS_AGE_MINUTES`)

Since feeds are polled at least hourly in steady state, going 2 hours with zero successes *anywhere* means fetching is broken globally (worker stuck, fetch/parse regression like #904, network egress failure) regardless of the specific cause. A "successful fetch" is a pollable feed (web type, ≥1 active subscriber) with a recent `last_fetched_at` and `last_error IS NULL` — any failure mode that sets `last_error` freezes this signal, so the June outage would have alerted within ~2 hours instead of 2 days.

## Alerting: healthchecks.io, three independent checks

All alerting goes through [healthchecks.io](https://healthchecks.io) (or any compatible dead-man's-switch) via shared `pingHealthcheck` / `startHeartbeat` helpers, rather than a bespoke channel. **Each ping URL is a separate check**, so the long-running processes get three independent ones — concurrent failures stay individually visible, and a dead worker is distinguishable from a fetch regression:

| Check | Env var | Pinged by | Signals |
| ----- | ------- | --------- | ------- |
| **Feed fetch health** | `FEED_HEALTH_HEARTBEAT_URL` | `monitor_feed_health` (worker, every 15 min) | `/fail` when no feed has fetched successfully within the threshold; **POSTs a body** (status, reason, last success, failing/pollable counts, most recent feed error) that healthchecks.io drops into the alert email |
| **Worker liveness** | `WORKER_HEARTBEAT_URL` | worker process (every 1 min) | success while the job loop is active, `/fail` if it wedges past the staleness threshold; missing pings = worker dead/gone |
| **Discord bot liveness** | `DISCORD_BOT_HEARTBEAT_URL` | discord-bot process (every 5 min) | missing pings = bot dead/crash-looping |

All three are optional (a process skips pinging when its URL is unset). A fully dead worker trips both the worker and feed-health checks (the monitor job stops pinging too); the worker check is the more specific signal, while a feed-health `/fail` with a green worker check points at the fetch pipeline rather than the process. healthchecks.io owns delivery, dedup, and recovery emails, so there's no in-process alert state.

This **removes the unused Discord webhook path** (`DISCORD_ALERT_WEBHOOK_URL` was never set in production) and the per-startup "worker started" Discord notice it powered.

New Prometheus gauges `feed_last_successful_fetch_age_seconds` and `feeds_failing` remain for dashboards.

## Testing

- `tests/unit/feed-health.test.ts` — the evaluation rule (boundary/never/no-feeds/partial-outage), the email body builder, and `buildPingUrl` (suffix appending, trailing slash, query-string preservation).
- `tests/integration/feed-health.test.ts` — the snapshot SQL against real Postgres (failing feeds don't count as successes, unsubscribed feeds excluded) and the status the handler reports.
- Full `pnpm test:integration` (350) and `pnpm test:unit` (1857) pass; `pnpm typecheck` and `eslint` clean.

## Setup (you)

Create up to three healthchecks.io checks, add your email as the notification method on each, and set the matching Fly secret:

1. **Feed fetch health** — period ~15m, grace ~1h → `FEED_HEALTH_HEARTBEAT_URL` (the important one)
2. **Worker liveness** — period ~1m, grace ~5m → `WORKER_HEARTBEAT_URL`
3. **Discord bot liveness** (optional) — period ~5m, grace ~15m → `DISCORD_BOT_HEARTBEAT_URL`

No `DISCORD_ALERT_WEBHOOK_URL` needed — that path is removed. `.env.example` documents all three with suggested periods/grace.


brendanlong:
Thanks for the review — applied all three suggestions in `7ea078e5`:

1. **Single-scan health query** — collapsed the four CTE subqueries into one aggregate scan with `count(*) FILTER (...)` / `max(...) FILTER (...)`, keeping `sample_error` as the lone subquery.
2. **`maxSuccessAgeMinutes` NaN guard** — added the trailing `|| 120`. This one matters here specifically: a non-numeric value makes the threshold `NaN`, and `ageMs > NaN` is always false, so the monitor would silently report healthy forever — exactly the failure this feature exists to catch.
3. **`/fail` URL building** — now uses the `URL` API (extracted to a `buildFailUrl` helper) with a string-concat fallback, so query strings / fragments survive.

Separately, I also reverted an unintended change to `migrations/schema.sql` that the bot read as "index cleanup": it wasn't intentional. `pnpm test:integration` runs `db:schema:test`, which regenerates `schema.sql` from the test database — and that DB is missing `idx_entries_feed_published_coalesce` / `idx_entries_published_coalesce` because migrations 0060/0061 create them with `CREATE INDEX CONCURRENTLY`, which the migration runner can't apply inside its transaction. Those indexes are real in production (created by migration, never dropped), so the regenerated schema was wrong and is now reverted to match master.

— Claude

brendanlong:
Reworked per discussion: dropped the Discord path entirely (`DISCORD_ALERT_WEBHOOK_URL` was never set in production and only powered an unused worker-start notice) and moved **all** monitoring to healthchecks.io.

- `monitor_feed_health` now pings `FEED_HEALTH_HEARTBEAT_URL` (success / `/fail`) and **POSTs a plain-text body** so the email tells you *why* it failed: status, reason, last successful fetch, failing/pollable counts, and the most recent feed error. No more in-process alert-state — healthchecks.io owns dedup/cadence/recovery, which also let me delete the payload throttle logic.
- Worker liveness is now the same check's missing-ping signal (the job pings every 15 min; a dead worker stops pinging).
- The Discord bot, which has no recurring job, gets its own 5-min liveness heartbeat to `DISCORD_BOT_HEARTBEAT_URL`.

`.env.example` documents the new vars; `pingHealthcheck`/`buildPingUrl`/the email-body builder are unit-tested. Squashed into the single PR commit and force-pushed.

— Claude

brendanlong:
Good call — split into three independent checks so concurrent failures are individually visible and a dead worker is distinguishable from a fetch regression:

| Check | Env var | Pinged by |
| ----- | ------- | --------- |
| Feed fetch health | `FEED_HEALTH_HEARTBEAT_URL` | `monitor_feed_health` (every 15m) |
| Worker liveness | `WORKER_HEARTBEAT_URL` | worker process (every 1m) |
| Discord bot liveness | `DISCORD_BOT_HEARTBEAT_URL` | discord-bot (every 5m) |

Previously I'd folded worker liveness into the feed-health check, so "worker dead" and "feeds broken" looked identical. Now the worker process pings its own check every minute — success while its job loop is active, `/fail` if the loop wedges past the staleness threshold (reusing the existing Fly health-check threshold), and missing pings if the process is gone. So a feed-health `/fail` with a green worker check points at the fetch pipeline, while a red worker check points at the process.

Generalized `startHeartbeat` to take a per-tick resolver so the worker can report `/fail` on a wedged loop instead of going silent. `.env.example` and DESIGN.md document all three with suggested period/grace values. Force-pushed.

— Claude

brendanlong:
Ran another review pass (subagent). No high-severity issues; the three earlier Gemini items are already in. Addressed the actionable findings in the latest push:

- **`startHeartbeat` resolver could throw uncaught** — the per-tick `resolve()` (used by the worker to report a wedged loop) is now wrapped in try/catch and falls back to a plain success ping, honoring the "pings never break the monitored work" contract. (Latent only — the one resolver just reads `getStats()`.)
- **`feed_last_successful_fetch_age_seconds` gauge in the never-succeeded case** — clarified in a comment that the gauge is intentionally left unset when no feed has *ever* succeeded (setting 0 would look healthy); Prometheus alerts for that state should key on `feeds_failing`, which is always set, and the healthchecks.io `/fail` ping is the primary signal regardless.
- **`sampleError` doc wording** — tweaked to match the query (`last_error IS NOT NULL`) rather than implying it's tied to `consecutive_failures`.

Two findings I did **not** change here:
- **Worker heartbeat stop fn discarded** (vs. the bot capturing it) — harmless since the interval is `unref`'d; not worth adding shutdown plumbing to the worker script.
- **Singleton self-creation race** — there's no unique constraint on `jobs.type`, so `claimSingletonJob`'s insert-and-catch can let two workers create the same singleton row during a deploy. Pre-existing (already affects `renew_websub`) and benign for health (healthchecks.io de-dupes), so I filed it separately as #914 rather than expanding this PR's scope.

— Claude

PR #915, opened 2026-06-19
Route all unread counts through visible_entries, fix merged-feed undercount

brendanlong:
## Context

We're about to drop delta-based unread-count updates on the frontend and recompute counts via query on every change, so these queries will run much more often. While investigating, I benchmarked the count queries and found (a) a pre-existing correctness bug and (b) some queries scaling with *total* entries instead of *unread* entries. This PR consolidates all unread counts onto the `visible_entries` view, which fixes both.

## What was wrong

1. **Merged-feed undercount (correctness).** `subscriptions.list` (and its `unreadOnly` filter) counted unread by matching `entries.feed_id` to the subscription's *current* `feed_id`, bypassing `subscription_feeds`. A subscription that absorbed a redirected/merged feed owns entries under the *old* feed_id too — those were silently dropped. Demonstrated by a regression test (returns 3, not 2).

2. **O(total) scans (performance).** Several count queries put `read = false` only inside a `FILTER` aggregate, so the scan read every visible row rather than letting the partial `idx_user_entries_unread` index drive. Pushing `read = false` into `WHERE` makes them scale with unread count instead.

## Approach

Route every unread count through `visible_entries` with `read = false` in `WHERE`. The view already encapsulates visibility + the `subscription_feeds` mapping, so the counts are correct by construction, and the planner pushes the predicate onto the partial unread index.

| Query (500K entries, 1% unread) | Before | After |
| --- | --- | --- |
| Global all/starred/saved (`counts.ts`) | ~171ms | ~40ms |
| Per-subscription list (`subscriptions.ts`) | ~67ms | ~42ms |
| Per-tag (`tags.ts`) | ~75ms | ~44ms |
| Uncategorized (`tags.ts`) | 61–306ms | ~45ms |

All rewrites verified to return **identical** counts to the originals on non-merge data; the per-feed rewrite additionally fixes the merge undercount.

### Subtleties handled
- **`getEntryRelatedCounts`**: the entry-info lookup was co-located in the count scan via `MAX(CASE …)`. Pushing `read = false` would drop the target entry's row (it's usually already read when counts are recomputed), so the lookup is split into its own tiny indexed query.
- **Active-subscription scoping**: `visible_entries` surfaces *starred* entries from *unsubscribed* feeds (non-null `subscription_id`). Those belong to Starred, not to a tag/uncategorized badge. Tag + uncategorized counts now join `user_feeds` (active-only). This also makes `counts.ts` consistent with `listTags`, which was already active-only while the SSE/mutation tag counts were not.

## Tests
New integration tests, each confirmed to **fail on the pre-fix query**:
- per-subscription unread through a merged feed; user-scoping on a shared feed
- uncategorized: basic, merged-feed, and starred-orphan exclusion
- per-tag: starred-orphan exclusion

Full `counts` / `tags` / `subscriptions` / `entries` / `saved-articles` integration suites pass; typecheck + lint clean. Stale "500ms+ at 50K" projection in the profiling doc updated with measured numbers.

### DIFF
--- a/docs/profiling-results-2026-02-25.md
+++ b/docs/profiling-results-2026-02-25.md
@@ -105,6 +105,39 @@ during SSR skip HTTP entirely, so the actual SSR prefetch cost is lower.
 Module loading is a one-time cost (cold start). After warmup, the CPU profile
 shows the server spending most time in I/O wait (waiting for Postgres responses).
 
+## Update 2026-06-19: unread-count scaling measured + fixed
+
+The "500ms+ at 50K" projection below was **not borne out**. Benchmarking on a
+synthetic single heavy user (50 subs, 30 tagged / 20 uncategorized, 5 tags, 10%
+then 1% unread) at 50K / 200K / 500K entries showed the uncategorized count at
+~18 / ~110 / ~306ms — never 500ms at 50K. The `subscription_feeds` junction
+(#1873) and the partial `idx_user_entries_unread` index had already turned the
+old GIN-array `NOT EXISTS` plan into a healthy `Hash Anti Join`.
+
+The remaining growth was **O(total entries)** rather than O(unread), because
+several count queries applied `read = false` only inside a `FILTER` aggregate
+(so the scan read every visible row) or hand-rolled joins by `entries.feed_id`
+(which also bypassed `subscription_feeds` — a latent correctness bug for merged
+feeds). Both were addressed by routing all unread counts through
+`visible_entries` with `read = false` in the `WHERE` (lets the partial unread
+index drive) and scoping tag/uncategorized counts to active subscriptions via
+`user_feeds`:
+
+| Query (500K entries, 1% unread)            | Before   | After |
+| ------------------------------------------ | -------- | ----- |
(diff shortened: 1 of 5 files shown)


brendanlong:
Re: the `userFeeds` join comments — addressed in 60f6bdb3 by adding `eq(userFeeds.userId, userId)` to all 6 joins.

To be precise about the reasoning: these were **not** a correctness or performance bug. The join is `userFeeds.id = visibleEntries.subscriptionId`, where `subscriptions.id` is a globally-unique PK (`subscriptions_pkey`) and `visible_entries` is already filtered to `user_id = userId` — so the join already matched exactly one row (this user's subscription), with no cross-user scan.

I benchmarked both forms at 500K entries / 1% unread to confirm there was no optimizer impact. Once cache-ordering effects are controlled for (each query warms the ~18K-buffer working set for the next), the two are identical — the plans are structurally the same and the `subscriptions` access is a 50-row seq scan either way. So the change is a no-op for correctness and performance-neutral; I added it purely to make the user-scoping explicit and consistent with the rest of the codebase.

— Claude

PR #917, opened 2026-06-19
Renew job lease via heartbeat so slow jobs aren't double-run (#871)

### LINKED ISSUE #871: Stale-job recovery can double-run non-idempotent jobs concurrently
## Problem

The background job queue's stale-job recovery can cause a single job to run in two workers at once, and the affected handlers are not idempotent mid-flight.

Flagged during architecture review; please verify against current `master`.

### Mechanism

- `src/server/jobs/queue.ts` (~L166) reclaims any job whose `running_since` is older than 5 minutes.
- `src/server/jobs/worker.ts` (~L43) has a 5-minute `jobTimeoutMs` that abandons the in-flight promise **without aborting the underlying work** (no `AbortController` propagated to HTTP fetches / DB writes).
- So a slow `handleFetchFeed` (which does full-content fetches and WebSub subscribe side effects) or `handleProcessOpmlImport` can still be running when a second worker reclaims and re-runs it.

## Why it matters

Concurrent execution duplicates outbound HTTP load on feeds/origin servers and races DB writes for the same entries. The "respectful fetching" guarantees in DESIGN.md assume one fetch at a time per feed.

## Suggested fix

Options to consider:
- Propagate an `AbortController` so the timeout actually cancels the in-flight work before the job becomes reclaimable.
- Make the reclaim window safely larger than the hard timeout, and ensure the original worker stops touching the job once it times out.
- Make the side-effecting handlers idempotent (and/or take an advisory lock keyed by feed id during fetch).

Related job-queue note: `claimSingletonJob` create-race (`queue.ts` ~L407–440) relies on insert+catch with no visible unique constraint — two workers could both insert a `renew_websub` row; worth confirming a unique constraint exists.

Found during a multi-agent architecture review.

— Claude

brendanlong:
Fixes #871.

## Problem

Stale-job recovery (`claimJob`/`claimFeedJob`/`claimSingletonJob`) reclaims any job whose `running_since` is older than the 5-minute stale threshold — which was **equal** to the worker's 5-minute `jobTimeoutMs`. The timeout abandons the in-flight promise *without aborting the underlying work*, so a slow `handleFetchFeed` (full-content fetches + WebSub side effects) or `handleProcessOpmlImport` could still be running when a second worker reclaimed and re-ran the same non-idempotent job. That duplicates outbound HTTP load and races DB writes, breaking the "one fetch at a time per feed" assumption in DESIGN.md.

## Fix

Add a **lease heartbeat**. While a handler runs, the worker renews `running_since` every minute via the new `renewJobLease()`. Because the lease keeps moving forward, stale-job recovery only fires once the worker process stops heartbeating (i.e. it actually died) — never while the handler is still executing, even if it has blown past `jobTimeoutMs`.

Key design points:
- The heartbeat is tied to the handler's **real lifetime** (a `finally` in `processJob`), *not* the worker-core timeout wrapper. So abandoned-but-still-running work stays leased until it truly settles, closing the timeout-abandonment race.
- Heartbeat interval (1 min) is comfortably smaller than the stale threshold (5 min), so ~5 consecutive missed renewals are tolerated before reclaim.
- `renewJobLease` is guarded on `running_since IS NOT NULL`, so a heartbeat racing with `finishJob` is a harmless no-op and can't resurrect a finished job's lease.
- Renewal failures are logged, never thrown — a transient DB hiccup won't take down the handler.

A truly-hung job (alive process, never-returning await) now holds its lease until the process restarts rather than being reclaimed — the correct trade-off vs. concurrent double-execution.

## Already-fixed sub-note

The issue also flagged the `claimSingletonJob` create-race. That's already handled on `master`: the `jobs_singleton_type_unique` partial index (migration `0069`) plus the unique-violation-only catch make concurrent self-creates safe.

## Tests

Added integration tests in `tests/integration/job-queue.test.ts`:
- a renewed lease keeps a 10-min-old running job from being reclaimed as stale
- a heartbeat racing with `finishJob` does not resurrect the lease
- `renewJobLease` returns `false` for a non-existent job

`pnpm test:unit` (worker), full `pnpm test:integration`, typecheck, and eslint all pass.

### DIFF
--- a/src/server/jobs/queue.ts
+++ b/src/server/jobs/queue.ts
@@ -15,7 +15,7 @@
  * See docs/job-queue-design.md for detailed documentation.
  */
 
-import { and, eq, isNotNull, sql } from "drizzle-orm";
+import { and, eq, sql } from "drizzle-orm";
 import { db } from "../db";
 import { jobs, type Job } from "../db/schema";
 import { generateUuidv7 } from "../../lib/uuidv7";
@@ -137,6 +137,14 @@ export interface FinishJobOptions {
   success: boolean;
   nextRunAt: Date;
   error?: string;
+  /**
+   * The `running_since` value this worker last held (its lease token). When
+   * provided, the finish is fenced on it: if another worker has since reclaimed
+   * the job (different `running_since`), the update applies to no rows and
+   * {@link finishJob} returns null instead of clobbering the new owner's state.
+   * Omit only for callers that aren't holding a lease (e.g. tests).
+   */
+  expectedRunningSince?: Date;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


brendanlong:
Addressed the review feedback in f40f009f.

**Lease stealing / fencing (the HIGH finding):** Confirmed and fixed. Both Gemini and an independent reviewer flagged the same gap — the heartbeat only guarded on `running_since IS NOT NULL`, so a worker that stalls past the 5-min stale threshold (alive but unresponsive) could have its job reclaimed by a second worker and then *steal the lease back* on a delayed heartbeat (split-brain).

The lease is now a **fencing token** — the `running_since` value the worker last wrote:

- `renewJobLease(jobId, expectedRunningSince)` updates only when `running_since` still matches and returns the new `running_since` (or `null` if the lease was lost). The heartbeat tracks this token, advances it each renewal, and on a `null` result gives up and stops the interval — it can't renew a job another worker reclaimed.
- `finishJob` also takes an optional `expectedRunningSince` and fences its write on it (returning `Job | null`), so a worker that lost its lease can't clobber the new owner's scheduling/state or clear `running_since` out from under a job running elsewhere. Unfenced callers keep the old throw-on-missing behavior.
- The worker **stops the heartbeat before finishing** (awaiting any in-flight renewal) so the token is final — this avoids a self-race between the heartbeat and the finish where the worker could falsely conclude it lost its own lease.

**Tests:** Added integration coverage for the actual two-worker race — a stalled worker cannot steal a reclaimed job via `renewJobLease`, and `finishJob` no-ops (returns `null`) without clobbering when the token no longer matches — plus updates to the existing renew/reclaim tests for the new signature.

One note on the residual: a worker frozen *longer* than the stale threshold will still get reclaimed (a frozen process can't heartbeat — that's the definition of "presumed dead"), so the heartbeat narrows but can't eliminate that window. What the fencing guarantees is that the reclaim is now *clean*: the thawed worker detects it lost the lease and neither renews nor finishes, so there's no split-brain. Fully closing the concurrent-execution window for a frozen-but-alive worker would require cooperative cancellation (an AbortController threaded through the fetch/DB stack), which I left out of scope here.

— Claude

PR #918, opened 2026-06-19
Replace delta count tracking with absolute server counts

brendanlong:
## Summary

Started as an SSE review (comparing against the fix in brendanlong/clawed-abode#361) and turned into eliminating **all** delta-based unread-count tracking in the realtime/cache layer, in favor of server-computed absolute counts.

The original SSE architecture was already sound (single `EventSource`, cursor-in-ref so the connection never churns, explicit catch-up sync on every open, full polling-fallback state machine). The one real defect was that `new_entry` counts were applied as a non-idempotent `+1` delta, so an entry delivered by **both** the live SSE stream and a reconnect catch-up sync (overlapping windows) could double-count. From there we removed every remaining delta.

## Commits

1. **`new_entry` → absolute counts (the bug fix).** The SSE route computes the connected user's counts per fan-out via `getNewEntryRelatedCounts`; `sync.events` attaches one bulk snapshot per batch. Client sets absolute counts, so re-delivery is idempotent. Adds a double-apply regression test.
2. **Delete dead delta code.** `handleEntriesMarkedRead` / `handleEntryStarred` / `handleEntryUnstarred` (entry mutations already used absolute counts) and the now-unused delta helpers (`adjustSubscriptionUnreadCounts`, `adjustTagUnreadCounts`, `calculateTagDeltasFromSubscriptions`, …).
3. **`subscription_created`/`deleted` → absolute counts; drop sidebar `feedCount`.** Both events/mutations now carry server-absolute counts (delete captures the sub's tag IDs before removing the associations, then computes post-delete counts). The sync.events catch-up path can't recompute a deleted sub's former tags, so it omits counts and the client invalidates `tags.list` + `entries.count` — a single refetch, only on reconnect catch-up, never on the live path. `adjustEntriesCount` is deleted, so **no delta count code remains anywhere**. `feedCount` no longer drives sidebar visibility (it lives only in settings now, which refetches); visibility gates on unread / show-read mode.

## Design notes / decisions

- **Live path stays pure push** (immediate, only-affected-lists, no refetch). Only the rare cross-device-during-disconnect catch-up for a deleted subscription refetches the two count caches.
- **`feedCount` removal**: its only load-bearing sidebar use was the "No subscriptions yet" prompt, which is unreachable by default (new users are auto-subscribed to the announcement feed at signup). Empty tags/sections now show in show-read mode.
- Unsubscribe stays optimistic (structural removal in `onMutate`, counts from the server response in `onSuccess`).

## Testing

- `pnpm typecheck`, `pnpm lint` clean (run via the direct binaries — the sandbox's `tsc`/`prettier` OOM under the pnpm/husky wrappers).
- `test:unit` 1804 passing (incl. new idempotency + absolute-count cases).
- `test:integration` 347 passing (incl. new create/delete absolute-counts coverage, including the tagged-delete post-removal count).
- e2e not run here (Playwright won't install on this OS); the realtime spec's comments were updated to match. The `new_entry`/`entry_state_changed` e2e assertions still hold; no subscription-event e2e exists.

— Claude


brendanlong:
Rebased onto latest master (clean, no conflicts) and re-reviewed the full diff. The re-review caught one real bug, fixed in the new fourth commit:

**Web-feed path published `new_entry` before the `user_entries` fanout.** `processEntry`/`processEntryWithCache` published per-entry inside the batch loop, while `createUserEntriesForFeed` runs after the loop. Since this PR makes the SSE endpoint compute absolute counts from `visible_entries` at event arrival, events published before the fanout produced counts that *excluded* the new entries — badges would have stayed stale until the next count-bearing event. (Harmless under the old +1 delta scheme, which is why the ordering existed; email and saved paths already published after their inserts.) `new_entry` is now published in `processEntries` after the fanout, with an integration regression test asserting the `user_entries` row exists at delivery time — the existing tests seeded `user_entries` manually before publishing, so they never exercised the worker's real ordering.

Also restored strict per-connection SSE delivery order via a send promise-chain: the async count query could otherwise let a `new_entry` with an older counts snapshot land after a newer count-bearing event.

All suites green after rebase: tsc, eslint, 1884 unit, 377 integration.

— Claude

PR #940, opened 2026-06-27
Render MathJax CHTML math as MathML in entry content

brendanlong:
## Problem

Math was missing from LessWrong articles (e.g. [Neuralese is actually probably good for alignment](https://www.lesswrong.com/posts/8iYx7wcBM4vawuin9/neuralese-is-actually-probably-good-for-alignment)) — sentences like *"Let ___ be the context… Let ___ be the answer…"* rendered with every variable gone.

## Root cause

LessWrong delivers math **pre-rendered as MathJax v3 CommonHTML (CHTML)** — both via the GraphQL/saved path *and* the RSS `feed.xml` path. CHTML is a tree of custom `<mjx-*>` elements plus a `<style>` block, where each glyph is an **empty** `<mjx-c class="mjx-c1D465 …">` element whose visible character comes from CSS (the codepoint is encoded in the class name, `mjx-c1D465` → U+1D465 `𝑥`).

Our entry sanitizer intentionally drops `<style>` and any non-allow-listed tag, and `<mjx-*>` tags aren't allowed. Because the glyphs live in CSS rather than in text nodes, stripping the CSS **and** the `<mjx-*>` elements made the math vanish entirely.

## Fix

Add `convertMathJaxChtmlToMathml`, a pre-sanitization transform (in `sanitizeEntryHtml`) that rewrites the CHTML tree to presentation **MathML** — already on the sanitizer allow-list and rendered natively by modern browsers (incl. mobile Firefox), no JS/MathJax needed.

- Handles tokens (`mi`/`mo`/`mn`/`ms`/`mtext`), `mrow`/texatom, `msup`/`msub`, `mover`/`munder`/`munderover` (with correct base/script reordering), `mfrac`, `msqrt`. Unknown `<mjx-*>` wrappers are unwrapped so inner tokens still survive.
- Cheap no-op (string check) for the common case with no embedded math.
- Bumps `SANITIZER_VERSION` to 2, so **already-fetched entries self-heal** — the read path re-sanitizes from retained raw content on next read, no re-fetch required.

## Verification

- New unit tests (`tests/unit/mathjax-chtml.test.ts`) using real CHTML samples from the post; full unit suite passes (1905), `pnpm typecheck` clean.
- Rendered the converted output in Chromium — the exact paragraph the bug report cited now shows all variables (`x`, `ŷ`, `y`, `z`, `θ`) and display equations like `y_{|x} ∼ p_θ(y∣x,z) p_θ(z∣x)` render correctly.


brendanlong:
Thanks for the review — both points were valid and are now fixed in 2093aa0c.

**Nested-fraction denominator hijacking (high):** Confirmed. `el.querySelector("mjx-den")` descends the whole subtree, so an outer fraction picked up a *nested* fraction's `<mjx-den>` (which precedes the outer one in document order because it lives inside the numerator), silently corrupting the math. Replaced the two `querySelector` calls with `findFractionParts`, a scoped traversal that descends through MathJax's layout wrappers but stops at this fraction's own `<mjx-num>`/`<mjx-den>` and at any nested `<mjx-mfrac>`. Added the suggested nested-fraction test plus a flat-fraction test.

I also hardened the converter based on a parallel review: the recursion runs on untrusted feed bodies at ingest time, so I wrapped the conversion in a try/catch in `sanitizeEntryHtml` — a pathological input (e.g. extreme nesting overflowing the stack) now degrades to "math stripped" instead of crashing the write path (verified: a 60k-deep nest logs a warning and falls back cleanly). And tightened the fast-path guard to `<mjx-container` so content merely mentioning the string in text doesn't trigger a full reparse.

— Claude

brendanlong:
Follow-up commit a39f5fba rewrites the converter against **real mathjax-full 3.2.2 output** generated with ForumMagnum's exact configuration (fixtures checked into `tests/unit/fixtures/mathjax-chtml-v3.json`), fixing the must-fix findings from the deep review:

- **Assistive MathML preferred verbatim** when present (`<mjx-assistive-mml>`), instead of discarding it — this makes the upstream ForumMagnum fix (assistive-MML PoC) flow through untouched.
- **Truncation fix**: parse as-is + `document.toString()` instead of synthetic `<body>` wrapping — content after a stray `</body>` is no longer dropped.
- **Display operator limits** (`\sum_{i=1}^n` etc.): handles the `mjx-box`/`mjx-row` composition layout and the inline `limits="false"` script layout.
- **Stretchy delimiters**: reads the codepoint from the `mjx-stretchy-v/h` element itself (assembly pieces are classless) — `\left(...\right)` no longer vanishes.
- **New coverage**: `msubsup`, `mtable`/`mtr`/`mtd` (matrices, cases), `mroot`, `mspace` (width preserved), `mjx-utext` (CJK etc.), surrogate-codepoint guard, zero-match early-out, and unknown-wrapper logging as a canary for MathJax layout drift.

`SANITIZER_VERSION` bumped to 3. Full unit suite (1925) + typecheck pass; end-to-end check against the original post now renders all 41 equations.

— Claude

PR #964, opened 2026-07-02
Services-layer cleanup: sanitization chokepoint, MCP validation, saved.save dedup (#956)

### LINKED ISSUE #956: Maintainability: services-layer erosion (saved.save duplication, sanitization/validation at wrong layer)
From a multi-agent design review (2026-07-02, verified at commit 0976051). The services-layer pattern ("routers are thin wrappers; invariant helpers are chokepoints") is eroding in several places. These are refactors, groupable into a couple of focused PRs.

## Duplication

- [ ] **`saved.save` in the router is ~540 lines duplicating the ~200-line `saveArticle` service** (`routers/saved.ts:227-788` vs `services/saved.ts:273-481`), with `extractMetadata` copied verbatim. They've **already diverged**: the service enforces `maxSavedArticleSizeBytes` and does Markdown-frontmatter extraction; the router does neither. The service cleans content synchronously on the request thread (`services/saved.ts:427`) while the router uses `cleanContentInWorker` — so MCP saves of large pages block the event loop. `saved.delete` (`routers/saved.ts:799-842`) duplicates `deleteSavedArticle` line-for-line. **Biggest single refactor.** Fold the router-only capabilities (refetch, provided-HTML, Google Docs auth) into `saveArticle` as options; have both callers and `saved.delete` use the service.

## Invariants enforced at the wrong layer (chokepoint gaps)

- [ ] **Read-path sanitization lives in the entries *router*, not the service** (`routers/entries.ts:268-450`), so **MCP `get_entry`, Google Reader, and Wallabag serve raw unsanitized HTML** (`services/entries.ts:500-501` → `mcp/tools.ts:98-100`, `google-reader/format.ts:59`, `wallabag/format.ts:96,156`). GReader clients render this. The sanitized columns are already persisted, so moving `resolveSanitizedContent`/`toFullEntry` into `services/entries.ts` is cheap and closes the gap for all four consumers. Update CLAUDE.md/DESIGN.md, which describe the read-path sanitizer as *the* chokepoint (only true for tRPC today).
- [ ] **MCP tool handlers take `any` args with no runtime validation** (`mcp/tools.ts:28,77-84,120,145,169`, eslint-disabled — violates the no-`any`/use-Zod rules). The advertised `inputSchema` is never enforced server-side (`api/mcp/route.ts:147` spreads raw args), so a client can pass `maxLimit: 1000000` (an internal Google-Reader override) straight into `listEntries` and blow past the 100-row cap; malformed `cursor`/`publishedAfter` types flow into SQL construction. Define per-tool Zod schemas and parse before calling services.
- [ ] **`entries.fetchFullContent` is ~140 lines of business logic in the router** (`routers/entries.ts:849-987`) — fetch/hash/sanitize/persist/narration-invalidation inline; only the raw HTTP fetch is in a service. It also mutates the *shared* `entries` row's `fullContentError` (`:903-910`), so one subscriber's failed fetch overwrites state visible to all subscribers. Extract to `services/full-content.ts` and decide deliberately whether error state should be shared.
- [ ] **`markAllEntriesRead` tag filter isn't user-scoped in the service** (`services/entries.ts:736-751`) — the `tagId` subquery joins `subscription_tags → subscription_feeds` with no `tags.userId` check, unlike the shared `entry-filters.ts:97-107`. The router pre-validates ownership today, but the guard belongs in the service (the documented reuse layer). Reuse `buildTaggedFeedIdsSubquery`.

## Smaller consistency items

- [ ] **`bookmarklet html` input is unbounded** (`routers/saved.ts:239` `z.string().optional()`, no max) and the router path never checks `maxSavedArticleSizeBytes` — a `saved:write` client can persist arbitrarily large rows (also sanitized ×4). Enforce the size limit.
- [ ] **Fabricated zeros in `getEntryRelatedCounts`** (`services/counts.ts:166-172`) — returns `{all: 0, starred: 0}` when the entry isn't visible even though the real global counts were just computed two lines above; a cache-patching caller would zero the user's badges. Return the computed counts.
- [ ] **Inconsistent invalid-cursor handling** — `entries` throws `errors.validation` (`services/entries.ts:217-228`); `subscriptions` silently restarts from page one (`services/subscriptions.ts:242-244`). Pick one.
- [ ] **`FeedIdsCondition` typed as `any`** (`services/entry-filters.ts:43-44`, eslint-disabled) — violates the no-`any` rule.
- [ ] **Doc drift** — DESIGN.md's "Error Responses" shape and its services table are out of date vs `trpc.ts:40-58` and the actual service exports (missing `users.ts`, `markAllEntriesRead`, `updateEntryStarred`, `getEntries`).

— Claude (filed from a design-review session)

brendanlong:
Fixes #956.

Restores the services-layer pattern ("routers are thin wrappers; invariant helpers are chokepoints") across the areas flagged in the issue, one commit per area:

## Read-path sanitization moved into the services layer
`resolveSanitizedContent`/`selectFullEntry`/`toFullEntry` moved from the entries router into `services/entries.ts`, and `getEntry`/`getEntries` now select the persisted `*_sanitized` columns (with the same version-gated self-heal). **This closes the gap where MCP `get_entry`, Google Reader, and Wallabag served raw unsanitized feed HTML.** CLAUDE.md/DESIGN.md updated to describe the services layer as the chokepoint. New integration tests cover the service-level path.

## MCP tool arguments validated with Zod
Each tool handler now parses its args against a Zod schema (failures → `InvalidParams`), and the advertised `inputSchema` is generated from that same schema via `z.toJSONSchema` so they can't drift. Unknown keys are stripped (no more `maxLimit` injection past the 100-row cap) and the authenticated `userId` is passed separately from client args instead of being spread into them. Integration tests cover rejection, stripping, and userId spoof-prevention.

## `entries.fetchFullContent` extracted to `services/full-content.ts`
The ~140-line fetch/hash/sanitize/persist/narration-invalidation flow is now `fetchAndStoreFullContent`. The write to the shared `entries` row (including `fullContentError`) is kept and documented as deliberate: full content and its fetchability are properties of the source URL, not the requesting user.

## `saved.save` folded into the `saveArticle` service (biggest diff)
The ~540-line router mutation duplicating (and diverging from) the service is gone. `saveArticle` now owns the whole flow with the router-only capabilities as options:
- `html` (bookmarklet-provided DOM), `refetch`/`force` with the content-quality guard, and `googleDocsAuth` for the interactive private-Google-Docs OAuth flow (public docs already came via the plugin)
- returns an `outcome` (`created`/`updated`/`existing`) so the router picks the right counts query
- Readability runs in the worker pool for **all** callers (MCP saves no longer block the event loop)
- `maxSavedArticleSizeBytes` enforced on every acquisition path, including provided HTML (plus a Zod max on the router input — the previously unbounded bookmarklet body)
- MCP callers now also get Markdown-frontmatter extraction and size limits consistently; `saved.delete` uses `deleteSavedArticle`

## Smaller consistency fixes
- `markAllEntriesRead` tag filter is user-scoped in the service via the shared `buildTaggedFeedIdsSubquery` (was router-pre-validation only) — with a regression test
- `getEntryRelatedCounts` returns the real global counts for non-visible entries instead of fabricated zeros
- `listSubscriptions` throws a validation error on invalid cursors (matching entries) instead of silently restarting from page one
- `FeedIdsCondition` is `string[] | SQLWrapper` instead of an eslint-disabled `any`

## Docs
DESIGN.md's Error Responses section now matches the actual tRPC `errorFormatter` output, and the services table lists the real exports (`users.ts`, `retention.ts`, `markAllEntriesRead`, `updateEntryStarred`, `getEntries`, …).

## Testing
- `pnpm typecheck` / `pnpm lint` clean
- `pnpm test:unit`: 1892 passed
- `pnpm test:integration`: 406 passed (incl. new tests for service-level sanitization, MCP arg validation, saved-size limit, tag scoping, non-visible-entry counts)
- `pnpm test:e2e`: 4 passed


brendanlong:
I ran a multi-agent self-review over the branch diff (8 finder angles) before finalizing, and pushed four follow-up commits fixing what it found:

**Regressions the review caught in my own refactor (now fixed):**
- Public Google Docs saves stored the plugin's `wrapHtmlFragment`-wrapped document as the article body; the sanitizer keeps `<title>` *text*, so the doc title leaked as stray text above the content and into the excerpt. Fixed by having `skipReadability` plugins (Google Docs, LessWrong, GitHub — the latter two were pre-existing) return bare fragments; documented on `SavedArticleContent`.
- When Readability failed, the unified `saveArticle` stored the raw full page as `contentCleaned` (the old router stored NULL) — inflating the refetch quality guard and serving site chrome as the article. Restored the NULL fallback.
- A `CONTENT_TOO_LARGE` error from a plugin was swallowed by the plugin-failure fallback; it's now rethrown.
- The stdio MCP transport demanded a `userId` tool argument that the Zod-generated schemas (`additionalProperties: false`) forbid, making it unusable for schema-strict clients. The acting user is now configured via `LION_READER_USER_ID` at startup.
- `listSubscriptions`' new cursor validation still let non-UUID ids through to a Postgres uuid-cast 500; now rejected as a validation error.

**Hardening/cleanup from the same review:**
- `SavedArticle.contentCleaned` now carries the *sanitized* body on all outcomes — MCP `save_article` and the Wallabag POST return it verbatim, so this closes the last raw-HTML escape from the services layer.
- MCP `create_tag`/`update_tag` enforce the same hex-color schema as tRPC (colors are rendered into inline styles); `uuidSchema`/`tagColorSchema` are now shared from `trpc/validation.ts`; `readOnly`/`unstarredOnly` (which worked through the old unvalidated spread) are documented schema params; service `TRPCError`s map to structured MCP `InvalidParams`; `registerTools()` is memoized.
- Sanitized-content resolution is split per family, so `getEntry`/`getEntries` (MCP/GReader/Wallabag) no longer fetch or re-sanitize full-content HTML they never return; the two full-entry select-field lists are now base + spread; the full-content persist logic is shared between the user-initiated fetch and the background worker (`persistFullContentResult`).

**Known trade-offs left as-is** (deliberate, flagged by the review): MCP now *rejects* out-of-range `limit` values instead of silently clamping; invalid subscription cursors error instead of silently restarting from page one; bookmarklet captures over `maxSavedArticleSizeBytes` (5MB default) are now rejected where they previously saved unbounded rows; the per-row sanitization self-heal in `getEntries` is O(stale rows) after a `SANITIZER_VERSION` bump (rare, self-limiting — a batch heal wasn't worth the complexity).

Full suite is green: typecheck, lint, 1892 unit, 406 integration, 4 e2e.

— Claude

PR #1036, opened 2026-07-07
Style <details>/<summary> blocks with a card border (#1031)

### LINKED ISSUE #1031: Details style
We should make the style for `<details>` nicer, so the expander is next to the text and we display some sort of card border around the contents. We should also make sure the spacing looks right both collapsed and expanded.

Example post with details: https://www.lesswrong.com/posts/kS5efcTP86EDTDbGh/brendan-long-s-shortform?commentId=sLa5mynmqifExHyck

brendanlong:
Fixes #1031.

## Problem

Collapsible `<details>` blocks (common on LessWrong posts) rendered with bare browser defaults — no visual framing. And because LessWrong wraps the summary title in a `<p>` (a block element), the disclosure marker was pushed onto its own line *above* the title text.

Example: https://www.lesswrong.com/posts/kS5efcTP86EDTDbGh/brendan-long-s-shortform?commentId=sLa5mynmqifExHyck

## Change

CSS-only, scoped to `.reader-prose` in `globals.css` so it applies to all feeds (not just LessWrong) without touching doc pages that use plain `.prose`:

- Card **border + rounded corners + subtle background** around the whole block (light + dark).
- Disclosure **marker stays beside the title** — force `summary > *` inline so the wrapping `<p>` no longer breaks the line.
- Summary bleeds to the card edges so the whole header is the click/hover target, with a hover fill.
- When open, a **divider** (the card border) separates the header from the body.
- The card's vertical padding is `0`, so a **collapsed** block is exactly the summary height; the body's spacing comes from the prose margins on its own children — spacing looks right in both states.

## Verification

Rendered the LessWrong details structure (`<summary><p><span>…</span></p></summary><div><blockquote>…</blockquote></div>`) with the new CSS and screenshotted collapsed + expanded states in a browser. `pnpm typecheck` passes.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -242,6 +242,68 @@ body {
 }
 
 /* =================================
+   Reader Prose <details>/<summary>
+
+   Feeds (notably LessWrong) ship collapsible <details> blocks. Browser
+   defaults + Tailwind Typography leave them bare, and because LessWrong wraps
+   the summary title in a <p>, the disclosure marker sits on its own line above
+   the text. Give the block a card border, keep the marker beside the title,
+   and make the spacing look right both collapsed and expanded.
+
+   The card's vertical padding is 0 so a collapsed <details> is exactly the
+   height of its <summary>; the body's spacing comes from the prose margins on
+   its own children (which don't collapse out through the card's border).
+   ================================= */
+
+.reader-prose details {
+  margin: 1.25em 0;
+  padding: 0 1em;
+  border: 1px solid #e4e4e7; /* zinc-200 */
+  border-radius: 0.5rem;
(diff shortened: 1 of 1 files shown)


brendanlong:
Here's how the styled `<details>` blocks look — collapsed and expanded, in light and dark, using the actual CSS from this PR against the LessWrong `<summary><p>…</p></summary>` structure:

details styling — light and dark, collapsed and expanded

— Claude

brendanlong:
Updated: moved the disclosure marker **inside the card**, left of the summary text (the conventional placement) — previously it sat in the margin outside the card border. Collapsed + expanded, light + dark:

details styling — marker inside the card

— Claude

PR #1038, opened 2026-07-07
Add route-level e2e tests for Wallabag and Google Reader compat APIs (#1021)

### LINKED ISSUE #1021: Add route-level tests for Wallabag and Google Reader compat APIs (auth 401/happy-path)
## Background

While auditing the Wallabag API (after the scope changes in #870) I found that both the Wallabag and Google Reader compatibility APIs returned **HTTP 500 instead of 401** for unauthenticated/expired-token requests — `requireAuth` was `throw`ing a `Response`, which Next.js App Router doesn't convert into the HTTP response. Fixed in #1018.

The bug went unnoticed because **neither compat API has any route-level tests**. The services layer they call is unit/integration-tested, but the HTTP surfaces (`src/app/api/wallabag/**`, `src/app/api/greader.php/**`) — auth handling, response formatting, status codes — are not exercised by any test. `pnpm typecheck`/`pnpm lint`/unit tests all passed with the 500 bug in place.

## Ask

Add e2e/integration coverage for these two HTTP surfaces. Minimum valuable cases:

**Auth (regression guard for #1018):**
- Unauthenticated request to each protected endpoint returns **401, not 500** (Wallabag: `/api/wallabag/api/{user,entries,tags,search,entries/exists}`; GReader: `/api/greader.php/reader/api/0/{user-info,unread-count,subscription/list,tag/list,token,...}`).
- Invalid/garbage token returns 401.

**Happy paths:**
- Wallabag: password grant → `access_token`; save article (`POST /entries`); list; tags; refresh_token grant; delete.
- GReader: `ClientLogin` → `Auth=` token; `user-info`; `subscription/list`; `stream/items/ids`.

**Auth negatives:**
- Wrong password → 401 on both token endpoints (already works; lock it in).

## Notes

- These are HTTP-server tests, so they fit the `tests/e2e/` harness (real server + DB) rather than `tests/integration/` (services/DB only). The e2e helpers (`tests/e2e/helpers.ts`) already seed users/sessions directly; a password user can be seeded with `argon2.hash` + `generateUuidv7` (see #1018 verification).
- Consider also asserting the security behavior tracked separately (scope/audience enforcement) once that's decided.

Reported by Claude.

brendanlong:
## Summary

Closes #1021.

Neither the Wallabag (`src/app/api/wallabag/**`) nor the Google Reader (`src/app/api/greader.php/**`) compat HTTP surface had any route-level tests. That's why the "500 instead of 401 for unauthenticated requests" bug (fixed in #1018) slipped past `typecheck`/`lint`/unit tests — the services they call are tested, but the HTTP routes' auth handling and status codes weren't exercised by anything.

This adds two e2e specs that hit the real server over HTTP (`tests/e2e/`, real DB), covering the cases the issue asked for.

## What's tested

**Auth regression guard for #1018**
- Every protected endpoint returns **401, not 500**, for missing tokens and for garbage tokens (Wallabag: `user`, `entries`, `tags`, `search`, `entries/exists`; GReader: `user-info`, `unread-count`, `subscription/list`, `tag/list`, `token`, `stream/items/ids`).

**Auth negatives**
- Wrong password → 401 on both token endpoints; unknown user → 401; invalid refresh token → 401.

**Happy paths**
- Wallabag: password grant → bearer token → `user`/`entries`/`tags`; `refresh_token` grant; save an article via `POST /entries` (against a throwaway local HTTP server, so the real fetch → clean → persist pipeline runs), confirm it in the list + `entries/exists`, then delete.
- Google Reader: `ClientLogin` → `Auth=` token → `user-info`; token echo; `subscription/list`; `stream/items/ids`.

## Notes

- Adds `createPasswordUser` to `tests/e2e/helpers.ts` (confirmed user with an argon2 password hash, no session).
- Logins are memoized per user: the shared per-IP `expensive` rate-limit bucket (capacity 10) is exercised by both compat login endpoints, so logging in on every case would trip a 429. The refresh-token test uses its own grant because rotation revokes the access token.
- The save happy-path relies on `ALLOW_PRIVATE_NETWORK_FETCH=true` (set in `.env.test`) so the app can fetch the test's localhost article server.

All 17 new tests pass locally (`pnpm test:e2e:local wallabag-api greader-api`).

### DIFF
--- a/tests/e2e/greader-api.spec.ts
+++ b/tests/e2e/greader-api.spec.ts
@@ -171,3 +171,38 @@ test.describe("Google Reader API happy path", () => {
     }
   });
 });
+
+/**
+ * A Google Reader token is a *scoped* session (`reader:full-access`), minted so
+ * it can't be replayed as a full-access browser session. It must be rejected
+ * (as if invalid) by the main tRPC/REST surface — whether presented as a bearer
+ * token or as the `session` cookie a browser would send — and by the Wallabag
+ * API (which authenticates OAuth access tokens, not sessions). These lock in
+ * that a leaked Google Reader token can't reach account management.
+ */
+test.describe("Google Reader API credential isolation", () => {
+  test("token cannot reach account-management endpoints as a bearer token", async ({ request }) => {
+    const token = await getToken(request);
+    // Session-only account surface (listing the user's active sessions).
+    const res = await request.get("/api/v1/users/me/sessions", {
+      headers: { Authorization: `Bearer ${token}` },
+    });
+    expect(res.status()).toBe(401);
+  });
+
+  test("token cannot be replayed as a browser session cookie", async ({ request }) => {
+    const token = await getToken(request);
+    const res = await request.get("/api/v1/users/me/sessions", {
+      headers: { Cookie: `session=${token}` },
+    });
+    expect(res.status()).toBe(401);
+  });
+
+  test("token cannot be replayed against the Wallabag API", async ({ request }) => {
+    const token = await getToken(request);
+    const res = await request.get("/api/wallabag/api/user", {
+      headers: { Authorization: `Bearer ${token}` },
+    });
+    expect(res.status()).toBe(401);
+  });
+});
--- a/tests/e2e/greader-api.spec.ts
+++ b/tests/e2e/greader-api.spec.ts
@@ -1,0 +1,173 @@
+/**
+ * Route-level tests for the Google Reader compatibility API
+ * (`src/app/api/greader.php/**`).
+ *
+ * The Google Reader HTTP surface — ClientLogin, GoogleLogin bearer auth,
+ * response formatting, status codes — has no unit/integration coverage. This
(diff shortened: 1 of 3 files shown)


brendanlong:
Added credential-isolation tests (the scope/audience enforcement the issue flagged as "tracked separately") — proving a token minted for one compat surface can't be replayed to escalate:

**Wallabag token** (OAuth access token, `reader:full-access`):
- Rejected at `/api/mcp` (lacks the `mcp` scope)
- Rejected on the session-only account-management REST surface (`/api/v1/users/me/sessions`)
- Rejected on the Google Reader API

**Google Reader token** (scoped session, `reader:full-access`):
- Rejected by the main tRPC/REST surface both as a `Bearer` token and as a `session` cookie (`validateSession` is fail-closed for scoped sessions)
- Rejected on the Wallabag API

Server logs during the run confirm the intended branches fire (`OAuth token missing MCP scope`; `You must be logged in`). 23 tests total, all green.

— Claude

Issue #1054, opened 2026-07-07
Hyper-optimize MathJax CHTML→MathML conversion (linkedom DOM dominates math-heavy sanitize cost)


brendanlong:
## Background

While benchmarking the entry-content sanitizer (`sanitizeEntryHtml`), the MathJax CHTML→MathML pre-sanitization transform (`convertMathJaxChtmlToMathml` in `src/server/html/mathjax-chtml.ts`) stood out as the dominant cost on math-heavy content.

Measured on a synthesized math-dense article (112 KB, ~80 MathJax blocks), per full `sanitizeEntryHtml` call:

| Component | Time |
|---|---|
| `convertMathJaxChtmlToMathml` (linkedom parse + tree walk + serialize) | **~16.4 ms** |
| `sanitize-html` itself | ~4.3 ms |
| Total pipeline | ~17 ms |

So on math content the MathJax conversion is ~4× the cost of the sanitization it's meant to support. This is somewhat ironic given the sanitizer was deliberately built DOM-free (`sanitize-html` over `htmlparser2`) to avoid `jsdom`; `convertMathJaxChtmlToMathml` reintroduces a DOM via `linkedom` (`parseHTML` + `querySelectorAll` + `document.toString()`).

## Why it's low urgency (but worth doing later)

- It only runs when the body contains `<mjx-container` (cheap `String.includes` early-return otherwise), so most entries pay nothing.
- Sanitized output — including this transform — is **persisted** (`*_sanitized` columns, `SANITIZER_VERSION`) and served from the DB, so this cost is normally paid once at write time, not per read.
- The recently added worker-thread offload (>10 KB bodies sanitize in a piscina worker) keeps even the cold/self-heal path off the app-server event loop.

## Possible directions

- Rewrite the transform to a streaming/SAX approach over `htmlparser2` (+ `domhandler` for the localized subtree restructuring) instead of a full `linkedom` document, matching the "prefer SAX-style parsing" guidance in CLAUDE.md. The structural work (fractions, scripts, tables) is localized to `<mjx-math>` subtrees, so a full document DOM may be avoidable.
- Or scope the linkedom parse to just the `<mjx-container>` substrings rather than the whole body, then splice results back — avoids parsing/serializing the entire (possibly large) article for a few equations.
[...]

PR #1055, opened 2026-07-07
Offload entry-content sanitization to worker threads on request paths

brendanlong:
## Summary

Benchmarking the entry-content sanitizer (per the analysis in this session) confirmed two things:

1. **The current `sanitize-html` approach is the right one** — ~9× faster than the old `jsdom + DOMPurify`, and it keeps `jsdom` out of the server bundle. No reason to revert.
2. **Sanitization is synchronous and CPU-bound** (tens of ms on large bodies), so a large sanitize blocks the app-server event loop that serves UI requests.

This PR moves sanitization **off the event loop for UI-facing request paths** using the piscina pool we already run, while leaving background jobs on the fast inline path.

## Changes

- **`pool.ts`**: `sanitizeEntryHtmlInWorker` offloads to the worker pool for bodies over **~10 KB** (inline otherwise), backed by a new `sanitizeEntryHtml` worker task. `cleanContentInWorker({ sanitizeCleaned: true })` fuses the cleaned-HTML sanitize into the same worker task so saved-article writes don't ship the string across the thread boundary twice.
- **`sanitize-entry.ts`**: `withSanitizedEntryContentAsync`, the async twin of the write-path chokepoint. Offloads per-field and accepts a `presanitized` hint so the fused worker result is reused without re-sanitizing — still the single place that assigns the `*_sanitized` columns.
- **Applied to app-server paths**: saved article save/upload, on-demand full-content fetch (`persistFullContentResult` gains `offloadSanitize`, default on), and the read-path self-heal (which fans out right after a `SANITIZER_VERSION` bump on deploy).
- **Background paths keep the synchronous helper**: feed worker and email ingest already run off the request path; the feed worker passes `offloadSanitize: false`.
- **`createUploadedArticle`** sanitizes once (original === cleaned) and reuses it for both columns.

## Why 10 KB

Small bodies sanitize in well under a millisecond, so the thread hop + string copy isn't worth it. The gate is deliberately low so even short UI-triggered writes don't block the loop.

## Tests

- `handleTask` worker tasks (sanitize + fused clean) proven **byte-identical** to inline `sanitizeEntryHtml`.
- `withSanitizedEntryContentAsync` proven equivalent to the sync chokepoint and to honor the `presanitized` hint.
- Verified a real >10 KB body round-trips through a spawned worker thread (using the compiled `dist/worker-thread.js`) and equals inline output.
- Full unit suite (2054) + saved/full-content/sanitized integration tests (474) green.

## Follow-up

- MathJax CHTML→MathML conversion (`linkedom`) dominates sanitize cost on math-heavy content — filed as #1054.
- `scripts/bench-sanitize.mts` added as a reproducible benchmark.


brendanlong:
Ran an independent (Fable) code review of this PR. It confirmed **no live stored-XSS** and verified the critical invariants clean: worker/inline equivalence, the 10 KB gate, no import cycles / no piscina recursion into the worker bundle, and a complete background-vs-app-server split. The `presanitized` hint was proven to always equal `sanitizeEntryHtml(the exact raw value stored)` on every current path.

It flagged latent hazards, which I've fixed in `cacd4f7`:

1. **Hint trust not enforced in code** — the `presanitized` reuse now requires the value to be provided (not `undefined`) *and* the raw field present in `values`, so a stale/misapplied hint can't persist a `*_sanitized` column that disagrees with its raw column.
2. **Missing fused result → spurious NULL** — dropped `?? null` on the fused-clean hint so an absent worker result falls through to a normal sanitize instead of storing NULL at the current `SANITIZER_VERSION` (blank body, self-heal never fires).
3. **Stale `dist/worker-thread.js` in dev** — the bundle now embeds the sanitizer allow-list, so a stale one could sanitize with an out-of-date `SANITIZE_OPTIONS` while the main process stamps the current version. Dev now prefers the TS source; production still uses the bundle. Verified both paths round-trip a >10 KB body byte-identically to inline.
4. **Read-path 500 on worker failure** — `sanitizeEntryHtmlInWorker` now falls back to inline on a task-level rejection (worker crash/OOM), matching the pool-init fallback, so the self-heal path can't turn `entries.get` into a 500.

Added regression tests for the hint enforcement. Remaining review items were nits (left as-is). Two follow-ups filed separately: #1054 (MathJax conversion cost) and #1056 (full-content fetch runs Readability inline on the event loop).

— Claude

PR #1072, opened 2026-07-08
greader: consolidate saved-feed synthetic subscription into a shared layer (#1069)

### LINKED ISSUE #1069: Google Reader saved-feed synthetic subscription is a scattered route-layer concern
## Context

PR #1068 exposes saved (read-it-later) articles through the Google Reader API by presenting the per-user saved feed as a synthetic, uncategorized "Saved Articles" subscription (`feed/{int64(savedFeedId)}`). It works, but the concept is implemented as a **route-layer bolt-on**, re-derived in each handler rather than living in a shared layer:

- `subscription/list` appends `formatSavedSubscription(savedFeedId)`
- `unread-count` adds a saved line + `countEntries({ type: 'saved' })`
- `stream/contents` and `stream/items/ids` resolve it via `resolveFeedStream` → `type: 'saved'`
- `mark-all-as-read` had to be patched separately to use `resolveFeedStream` (it was silently no-op'ing — the bug found in review that motivated this issue)

## Problem

Because "the saved feed is a subscription" is materialized only at the HTTP layer, every endpoint that lists subscriptions or resolves a feed stream must re-orchestrate it by hand. That's how `mark-all-as-read` got missed initially, and any future Google Reader endpoint (or internal consumer) that enumerates subscriptions will have to remember to include it too. The `type === 'saved'` special-case is likewise scattered across route handlers.

## Idea

Consider hoisting the concept to a lower layer so endpoints inherit it for free, e.g. one of:

- A `listSubscriptions`-level option (or a thin `listSubscriptionsIncludingSaved`) that returns the saved feed as a first-class synthetic subscription, so `subscription/list` / `unread-count` don't special-case it.
- Or surface the saved feed in the `user_feeds` view (weigh carefully — it would leak the synthetic subscription into the core data model and the main app UI, which is exactly what PR #1068 deliberately avoided).

There's a real tension here: keeping it compat-API-only is intentional (the saved feed is not a real subscription in the app), so the fix should probably be a **shared helper within the `src/server/google-reader/` layer** (e.g. a single "resolve/enumerate greader subscriptions including saved" utility) rather than pushing it into the core services/views. Not urgent — the current code is correct after #1068 — but worth consolidating before more greader endpoints accrete copies of the special-case.

---
Filed by Claude at the maintainer's request following a code review of #1068.

brendanlong:
Fixes #1069.

## Problem

PR #1068 exposed saved (read-it-later) articles through the Google Reader API as a synthetic, uncategorized "Saved Articles" subscription (`feed/{int64(savedFeedId)}`), but the concept was a **route-layer bolt-on** re-derived in each handler:

- `subscription/list` appended `formatSavedSubscription(savedFeedId)`
- `unread-count` added a saved line + a separate `countEntries({ type: 'saved' })`
- `stream/contents` / `stream/items/ids` each ran the same `resolved.kind === 'saved'` switch
- `mark-all-as-read` had to be patched separately (it was the bug that motivated the issue)

Because "the saved feed is a subscription" was materialized only at the HTTP layer, every endpoint had to re-orchestrate it by hand, and a new one could silently omit it.

## Change

Hoist the concept into one shared helper module, `src/server/google-reader/subscriptions.ts` (kept compat-API-only, as #1068 intended — nothing leaks into core services/views/UI):

- **`listGreaderSubscriptions`** — drains the subscription cursor and appends the saved feed as a first-class synthetic `Subscription`. `subscription/list` and `unread-count` now enumerate through this and inherit saved for free.
- **`resolveFeedStreamFilter`** — maps a `feed/{int64}` stream to the entries-service filter (`{ type: "saved" }` or `{ subscriptionId }`), used identically by `stream/contents`, `stream/items/ids`, and `mark-all-as-read`. This is the single translation point that once diverged.
- **`getSavedSubscription`** — the saved feed as a synthetic `Subscription` (with unread count); the one place its metadata is materialized.

Because the saved feed is now a first-class synthetic `Subscription`:
- `formatSavedSubscription` is deleted (a synthetic sub through `formatSubscription` produces byte-identical output).
- `formatUnreadCounts`' `savedFeed` param is gone — saved is counted and folded into the reading-list total like any real feed.
- The saved-vs-subscription origin stream id (duplicated between item formatting and item refs) is unified into an exported `entryFeedStreamId` helper.

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm test:unit` (2062 passed)
- `pnpm test:e2e:local greader-api` — all 18 pass, including the saved-articles synthetic subscription and mark-all-as-read-on-saved regression tests.

### DIFF
--- a/src/app/api/greader.php/reader/api/0/mark-all-as-read/route.ts
+++ b/src/app/api/greader.php/reader/api/0/mark-all-as-read/route.ts
@@ -13,7 +13,7 @@
 import { requireAuth } from "@/server/google-reader/auth";
 import { parseFormData, textResponse, errorResponse } from "@/server/google-reader/parse";
 import { parseStreamId } from "@/server/google-reader/streams";
-import { resolveFeedStream } from "@/server/google-reader/id";
+import { resolveFeedStreamFilter } from "@/server/google-reader/subscriptions";
 import { resolveTagByName } from "@/server/google-reader/tags";
 import { db } from "@/server/db";
 import { markAllEntriesRead } from "@/server/services/entries";
@@ -46,20 +46,14 @@ export async function POST(request: Request): Promise<Response> {
   // Translate GReader stream into shared service params
   switch (parsedStream.type) {
     case "feed": {
-      const resolved = await resolveFeedStream(db, userId, parsedStream.subscriptionInt64);
-      if (!resolved) {
+      const filter = await resolveFeedStreamFilter(db, userId, parsedStream.subscriptionInt64);
+      if (!filter) {
         return textResponse("OK");
       }
 
-      // The saved feed has no subscription; mark all saved entries read via the
-      // type filter (issue #730), mirroring stream/contents' resolution.
-      await markAllEntriesRead(db, {
-        userId,
-        ...(resolved.kind === "saved"
-          ? { type: "saved" as const }
-          : { subscriptionId: resolved.subscriptionId }),
-        before: beforeDate,
-      });
+      // `filter` is the saved-feed type filter or a real subscription id (issue
+      // #730), resolved identically to the stream/contents endpoints.
+      await markAllEntriesRead(db, { userId, ...filter, before: beforeDate });
       break;
     }
     case "state": {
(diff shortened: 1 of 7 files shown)


brendanlong:
**Reviewer note (behavior delta for the record):** the saved feed's `newestItemTimestampUsec` in `unread-count` now derives from the synthetic sub's `subscribedAt` (epoch → `"0"`) instead of the previous `Date.now()`. This field is only a recency-sort hint for feed ordering — it's not used for read-state sync or continuation — and real subscriptions already report `subscribedAt` here, so this makes the saved feed *consistent* with every other feed rather than the lone one reporting "now". No known Google Reader client depends on it and no test asserts it.

Also for the record: `subscription/list` now pays one extra (discarded) `countEntries({type:"saved"})` query via the shared `getSavedSubscription`, gated behind the saved feed existing. It's a connect/refresh path, not hot, so this is an acceptable cost for the single-materialization-point win.

— Claude

Issue #1090, opened 2026-07-08
Misc hardening: rate-limit fail-open, webhook replay, worker sanitizer guard, retention gaps


brendanlong:
Found in the repo-wide audit (verified in code). Lower-severity hardening / maintainability items across auth, jobs, sanitization, and retention.

### Rate limiting fails fully open on Redis outage (security)
`src/server/rate-limit/index.ts:119-127,151-160`: `checkRateLimit` returns `{allowed:true}` when Redis is down, so both per-IP and per-account password brute-force protection evaporate exactly when the system is degraded. Consider fail-closed (or a small in-memory cap) specifically for password-accepting paths.

### Mailgun webhook has no timestamp-freshness check (security)
`src/app/api/webhooks/email/mailgun/route.ts:36-53,164-180`: valid `(timestamp, token, signature)` is replayable forever (dedup by Message-ID bounds it, but a different previously-captured email re-ingests). Reject stale timestamps (~15min) and cache the token as a nonce.

### `parseScopes` treats scope keys as valid values (correctness, latent)
`src/server/oauth/utils.ts:222-226`: `s in OAUTH_SCOPES` matches keys (`"MCP"`), so `scope=MCP` parses as valid. No escalation today (`validateScopes` re-filters against values), but drop the `s in OAUTH_SCOPES` clause.

### Worker pool never verifies the bundle's `SANITIZER_VERSION` (security, latent)
`src/server/worker-thread/pool.ts:63-76,207-213`: a stale `dist/worker-thread.js` would sanitize offloaded bodies with old options while the main process stamps the current version, silently persisting under-sanitized HTML the version gate never revisits. `getWorkerSanitizerVersion()` exists but is only called by a script. Call it at pool init and disable the pool (inline fallback) with a loud error on mismatch.

### Stored AI summaries are never re-sanitized after a version bump (security)
`src/server/services/summarization.ts:249-251`: `entry_summaries` has no sanitizer-version column and cached summaries render via `dangerouslySetInnerHTML` (`SummaryCard.tsx:133`). If a bump ever fixes a sanitizer hole, old summaries keep the vulnerable output. Add a version stamp and re-sanitize/regenerate on mismatch.

### MathJax transform loses content on implied `<mjx-math>` close (correctness, low)
[...]

PR #1107, opened 2026-07-08
Misc hardening: rate-limit fallback, webhook replay, worker guards, retention (#1090)

brendanlong:
Addresses the lower-severity hardening items from the repo-wide audit in #1090.

## Changes

**Security**
- **Rate-limit fail-open on Redis outage** — `checkRateLimit` now takes a `fallback` option; `checkAccountRateLimit` (the shared per-account password guard) uses `"memory"`, degrading to a per-process, size-bounded (LRU) in-memory token bucket when Redis is down instead of allowing unlimited guesses. Ordinary paths still fail open (availability > strictness).
- **Mailgun webhook replay** — reject timestamps outside a ~15-min freshness window and record the webhook token as a single-use Redis nonce, so a captured `(timestamp, token, signature)` triple can't be replayed with an arbitrary body (the signature covers only timestamp+token, not the body).
- **Worker pool sanitizer guard** — probe the worker bundle's `SANITIZER_VERSION` once at first use (`getVerifiedPool`); on mismatch, tear the pool down and sanitize inline with current rules (loud error), so a stale `dist/worker-thread.js` can't silently persist under-sanitized HTML.
- **`.well-known/*` + `/api/mcp` rate limiting** — now use the generous `oauth` bucket, matching DESIGN.md (previously unrated).

**Correctness**
- **`parseScopes`** — dropped the `s in OAUTH_SCOPES` clause so enum KEY names (e.g. `"MCP"`) no longer parse as valid scopes.
- **MathJax transform** — only advance `mathEnd` on an *explicit* `</mjx-math>` close; an implied close no longer swallows absorbed article content (regression test added).
- **Job singleton starvation** — round-robin feed vs singleton claim order so a `fetch_feed` backlog can't starve `renew_websub`/`cleanup`/`monitor_feed_health`/`resanitize_entries`.
- **Wedged-handler lease** — hard cap the lease heartbeat at 6× the job timeout; a permanently-wedged handler stops heartbeating and alerts (Sentry) so the job becomes reclaimable, instead of pinning its lease until restart.

**Maintainability**
- **Retention gaps** — sweep expired/revoked `api_tokens` (mirrors sessions) and terminal `opml_imports` past a 30-day window; documented why the content-hash-keyed caches stay append-only.

**Stored AI summaries** — investigated and found the concern already mitigated: the summarization read path unconditionally re-sanitizes cached summaries with current rules on every read, so a `SANITIZER_VERSION` bump auto-applies without a version column. Documented the invariant and added an integration test that locks it in (no migration needed).

## Tests
- Unit: `parseScopes` (keys rejected), MathJax implied-close content recovery (verified failing without the fix).
- Integration: `api_tokens`/`opml_imports` retention, cached-summary re-sanitize-on-read.
- Full unit suite (2099), lint, and typecheck pass.


brendanlong:
Ran an independent Opus-level review over the diff. It confirmed the rate-limit fallback, retention sweeps, `parseScopes`, MathJax guard, job-worker round-robin/lease-cap, and MCP/OAuth rate-limit wiring are correct, and surfaced two issues that are now fixed in 6fec597a:

1. **Mailgun nonce could drop legitimate mail (medium-high).** Mailgun reuses the identical `(timestamp, token, signature)` on every retry, so consuming the token *before* processing meant a crash/kill between the nonce write and the `200` would get the retry nonce-rejected and the email lost. Reworked to **check-before / mark-after-durable-completion**: the token is marked seen only once `processInboundEmail` resolves, never on an unexpected throw, so a retry after an incomplete run reprocesses (concurrent duplicates stay covered by the durable Message-ID dedup).
2. **Worker pool: transient probe failure permanently disabled offload (low).** A one-off worker crash/timeout during the version probe would tear down the pool for the whole process. The probe now distinguishes `mismatch` (permanent disable — correct) from `error` (skip offload for this call only, retry next time).

typecheck, lint, and the full unit suite (2099) pass; retention + summarization integration tests pass against a local DB.

— Claude

brendanlong:
Ran a second, Fable-powered review focused specifically on silent-failure modes in the Mailgun and worker changes (per maintainer request). It caught one **critical** bug in this PR's own Mailgun change, now fixed in 80ff7e64:

**M1 (critical): 15-min freshness window vs Mailgun's ~8h retry schedule.** Mailgun retries reuse the *original* `(timestamp, token, signature)` triple, so after any outage longer than the window, every later retry arrived "stale" → 406 → Mailgun treats 406 as don't-retry → **permanent silent mail loss for everything sent during a >15min outage**. Widened the window to 9h (full retry horizon); actual replays within it are still blocked by the Redis nonce, and Message-ID dedup bounds the Redis-down case.

**M2: webhook `catch` returned 200, so transient processing failures (e.g. a DB blip) silently dropped the email** — and contradicted the nonce design's whole premise that a retry reprocesses. Now returns 500 so Mailgun retries; safe because an incomplete run never marks the nonce and reprocessing is idempotent (`uq_entries_feed_guid` + Message-ID dedup). This was pre-existing behavior, but this PR's nonce redesign is what made fixing it safe.

**W1: lease hard-cap floor.** A `jobTimeoutMs` small enough that 6× is under one heartbeat would trip the cap on the first tick, silently disabling the issue-#871 anti-double-execution lease. Floored at 3 heartbeats. (Latent — prod defaults were safe.)

Also documented the `SINGLETON_JOB_TYPES` ordering invariant (fast-rescheduling singletons must stay last).

**New test coverage for both previously-untested silent-failure areas:**
- `createWorkerClaimJob` extracted as an injectable factory + 8 unit tests pinning the round-robin (singletons claimed under sustained feed backlog, fall-through both directions, `jobTypes` filtering unchanged).
- Lease heartbeat interval made injectable + 2 integration tests against the real DB (renewals stop at cap; post-cap finish commits when unreclaimed, CAS no-ops when reclaimed).
- 6 Mailgun route integration tests with real signed requests against Postgres+Redis: end-to-end ingest, stale 406, **4h-old retry accepted** (the M1 regression test), replay of a completed triple with a different body acked-but-not-ingested, bad signature/missing fields 406.

The review verified the rest as correct: hard-cap CAS fencing (walked through reclaim races), round-robin efficacy against worker-core's claim loop, `jobTypes` filtering equivalence, idle-poll query cost, concurrent-duplicate webhook race (safe via `uq_entries_feed_guid`), and timestamp unit handling.

Full suites green: 2110 unit, 549 integration, lint, typecheck.

— Claude

Issue #986, opened 2026-07-04
claude.ai web MCP connector can't connect (claude.ai-side bug); consider mcp.lionreader.com subdomain workaround


brendanlong:
> **STATUS (updated 2026-07-12) — subdomain shipped; confirmed claude.ai-side bug, no server-side fix remains.**
>
> The `mcp.lionreader.com` subdomain workaround proposed below has been **built, deployed, and tested**: MCP endpoint at `https://mcp.lionreader.com/mcp`, OAuth at the origin root, byte-comparable to Notion/Linear/Sentry (PRs #1037, #1148, #1150). It did **not** fix the claude.ai web connector, but it got far enough to pin down exactly where the connector breaks.
>
> **Both connector paths fail, both provably client-side (full server logs in the comments):**
> - **Auto-DCR:** discovery ✅ → `POST /register` returns **201** with a same-origin `registration_client_uri` ✅ → claude.ai then **aborts before ever calling `/authorize`** and re-POSTs `/mcp` with no token.
> - **Manual Client ID:** the **entire OAuth flow completes** — `authorize` → consent → `POST /token` → **`OAuth token issued`** (valid Bearer, `scope=mcp`, `resource=https://mcp.lionreader.com/mcp`) — and then the very next `POST /mcp` (`User-Agent: Claude-User`) arrives ~2.8s later with **no `Authorization` header** → 401. Textbook anthropics/claude-ai-mcp#136 (token issued, then dropped), now reproduced on a pristine dedicated subdomain, which rules out every server-side cause.
>
> **Server-side work is exhausted.** Everything claude.ai receives from us is verified equivalent to the servers that work with the connector. Remaining action is the upstream report to `anthropics/claude-ai-mcp` (needs the `ofid_…` + error text from a fresh attempt paired with the server-log timestamps in the comments). Interim: connect via **Claude Code / Claude Desktop / `mcp-remote`**, which all work.
>
> Debug tooling `MCP_HOST` + `LOG_MCP_REQUESTS` shipped in the above PRs (see docs/DESIGN.md). `LOG_MCP_REQUESTS` should be turned off once the upstream report is filed.
>
> ---
>
> _Original issue below is retained as the historical record (it predates the subdomain work; some details — e.g. the `/api/mcp` resource identifier and the "possible workaround" framing — are superseded by the status above)._

## Summary
[...]

Issue #367, opened 2026-01-21
Frontend Testing: Add cache operations tests


brendanlong:
## Summary
Expand tests for cache operations in `src/lib/cache/operations.ts`. These functions handle all cache interactions when entries are marked read, starred, etc.

## Priority: High
Cache operations are critical for correctness - bugs here cause UI inconsistencies.

## Current Coverage
Basic tests exist in `tests/unit/frontend/cache/operations.test.ts` covering:
- `handleEntriesMarkedRead`
- `handleEntryStarred`
- `handleEntryUnstarred`
- `handleNewEntry`

## Expansion Needed
1. Test `handleSubscriptionCreated` and `handleSubscriptionDeleted`
2. Add more granular tests for cache update logic
3. Test edge cases (null subscriptions, empty arrays, etc.)
4. Consider integration tests that verify actual cache state

## Notes
The current mock `TRPCClientUtils` tracks operations but doesn't simulate the full cache behavior. For deeper testing, consider:
- Creating a more sophisticated mock that tracks cache state
- Using integration tests with a real QueryClient

## Labels
enhancement

Issue #324, opened 2026-01-17
Discord save article bot


brendanlong:
It would be nice if we could save articles from Discord by emojiing the message linking to them, and having a bot that listens for that and saves the link.

PR #1121, opened 2026-07-10
Align MCP OAuth surface with working remote MCP servers (#986)

brendanlong:
## Why

Investigating #986 (claude.ai web connector can't connect), I did a field-by-field live comparison of our MCP/OAuth surface against the three remote MCP servers that demonstrably work with claude.ai's web connector — `mcp.linear.app`, `mcp.sentry.dev`, `mcp.notion.com` — replaying claude.ai-style requests at every step. Several concrete differences turned up that the subdomain workaround (#1037) doesn't address. This PR closes all of them, so the only remaining observable difference from the working servers is the host/URL shape.

## The differences found and fixed

**1. Token-endpoint auth methods (most suspicious — matches both failure signatures in #986):**

| | Us (before) | Linear / Sentry / Notion |
|---|---|---|
| `token_endpoint_auth_methods_supported` | `["none"]` | `["client_secret_basic","client_secret_post","none"]` |
| DCR with `client_secret_basic` | **400 reject** | 201 |
| DCR with method omitted | registers `none` | registers `client_secret_basic` (RFC 7591 §2 default) |
| `/token` with Basic credentials | ignored ("Missing client_id parameter") | client resolved from Basic header |

The old metadata also contradicted the register endpoint (which accepted `client_secret_post` while the metadata said only `none` exists) — the exact failure class of anthropics/claude-ai-mcp#285, and it declared manually-registered confidential clients (client ID + secret pasted into claude.ai's Advanced settings — failure #1 in #986) unsupported. Now: one shared list (`SUPPORTED_TOKEN_ENDPOINT_AUTH_METHODS`) drives both metadata and register validation; `extractClientCredentials` (new `oauth/client-auth.ts`) reads credentials from the form body or an HTTP Basic header for both the token and revocation endpoints.

**2. Trailing-slash 308:** `POST /api/mcp/` used to 308-redirect (Next's default); the working servers answer in place. claude.ai's connector uses python-httpx (doesn't follow redirects on POST) and has been observed appending trailing slashes (anthropics/claude-ai-mcp#324). **(Reworked after review)**: the custom HTTP server (`scripts/server.ts`, used in dev and prod) strips trailing slashes from exactly the OAuth/MCP paths (`src/server/http/trailing-slash.ts`) before Next routing. Next's built-in 308 stays in force for every other URL, and `next.config.ts`/`src/proxy.ts` are untouched relative to master. The first iteration used `skipTrailingSlashRedirect` + a site-wide middleware matcher, which made all slash handling depend on middleware coverage (an unmatched slashed path under that flag is served a broken empty 200) — this version has no such dependency. `/register` is deliberately not normalized: claude.ai POSTs its DCR to slashless `/register`, and `GET /register/` keeps redirecting to the signup page.

**3. Caching:** OAuth well-known documents were served with `Cache-Control: public, max-age=3600`; the working servers allow no caching (Notion sends `no-store`). Now `no-store` — clients re-run discovery on every connect and must never act on a stale document.

**4. Path-inserted AS metadata:** `/.well-known/oauth-authorization-server/api/mcp` 404'd; Linear and Sentry serve it, and claude.ai's second-stage discovery has constructed that URL from the resource path (upstream #367/#490). Now served (re-export of the root handler).

**5. Byte-level response parity:** RFC 7009 revocation endpoint at `/oauth/revoke`, advertised as `revocation_endpoint` (all three working servers advertise one); `registration_client_uri` in the DCR response (all three include it; decorative, like Linear's, which 404s — RFC 7592 management isn't implemented and can't be used without a `registration_access_token` anyway); RFC 6750-style 401 body on `/api/mcp` (`{"error":"invalid_token",...}` instead of `{"error":"Unauthorized"}`).

## Compatibility

- Existing registered clients are untouched; the RFC 7591 default-method change only affects new registrations that omit `token_endpoint_auth_method` (MCP SDK clients send `"none"` explicitly, and clients that omit it demonstrably handle a secret — Linear/Sentry/Notion already default to `client_secret_basic`).
- `GET /register/` still 308s to the signup page; `POST /register` and `POST /register/` both reach the DCR handler.
- Site-wide trailing-slash behavior is unchanged (308 to the slashless URL), still emitted by Next's built-in redirect.

## Testing

- `pnpm typecheck`, lint, `pnpm test:unit` (2127 pass, incl. new `oauth-client-auth` unit tests + proxy trailing-slash cases), `pnpm test:integration:local` (548 pass, incl. new register auth-method and RFC 7009 revocation tests).
- Verified end-to-end against a local dev server: slashed OAuth/MCP URLs answer in place, other slashed paths 308, DCR with basic/post/none matches the working servers' responses, Basic-auth token exchange resolves the client, `/oauth/revoke` returns 200, metadata/cache headers match.

Related: #986, #1037 (complementary — this PR removes the spec-level differences; #1037 removes the host-shape difference).


brendanlong:
Post-open verification:

- **Independent Opus review** (client auth, secret comparison, proxy rewrite targets, header-parsing edge cases, matcher blast radius): no blockers. Two intentional behaviors flagged for sign-off, both already described above: (1) DCR requests that *omit* `token_endpoint_auth_method` now register as confidential (`client_secret_basic`, RFC 7591 §2 default, matching Linear/Sentry/Notion) instead of public — MCP SDK clients (Claude Code, mcp-remote) send `"none"` explicitly and are unaffected; (2) `registration_client_uri` intentionally 404s on GET, same as Linear's.
- The review also confirmed: query strings and encoded slashes (`%2F`) survive the trailing-slash rewrite; the Wallabag rotation path and Google Reader API are untouched.
- **Full e2e suite passes locally** (39/39) against the changed middleware — including the SSE realtime tests, which were the reviewer's one requested confidence check for the widened matcher.

— Claude

brendanlong:
Rebased onto master (resolved a DESIGN.md conflict with #1107's rate-limiting rework — `/oauth/revoke` joins the `oauth`-bucket write endpoints; the well-known/`/api/mcp` no-rate-limit rule is preserved) and **reworked the trailing-slash handling to be much smaller in scope**:

- No more `skipTrailingSlashRedirect`, no middleware changes — `next.config.ts` and `src/proxy.ts` are now identical to master.
- Instead, the custom HTTP server (`scripts/server.ts`, the entry point in both dev and production) strips trailing slashes from *only* the OAuth/MCP paths (`/api/mcp`, `/oauth/*`, `/token`, `/authorize`, `/.well-known/*`) before Next.js routing. Next's built-in 308 redirect keeps handling every other URL, so there's no site-wide dependency on middleware matcher coverage and no possibility of the broken-empty-200 failure mode anywhere.
- `/register` is excluded from normalization: `GET /register` (and the 308 from `/register/`) still serves the signup page exactly as before; claude.ai POSTs its DCR to slashless `/register`, which the existing method-split proxy handles. (The `registration_client_uri` in DCR responses points at `/oauth/register/<id>` — a different path from the signup page, so its intentional GET 404 can't collide with the UI.)

Re-verified after the rework: 2150 unit / 561 integration / 39 e2e tests pass, and live checks confirm `POST /api/mcp/`, `/oauth/token/`, `/oauth/register/` answer in place while `/all/`, `/register/`, and unknown slashed paths still 308 via Next's own redirect.

— Claude

PR #1149, opened 2026-07-12
Add e-paper theme mode for e-ink screens (#1017)

### LINKED ISSUE #1017: E-paper theme mode
We should add an e-paper theme alongside light/dark mode which is similar to light mode but uses pure white as the background and which we ensure looks good in grayscale (it doesn't need to be grayscale, but every color needs to look right if forced to be grey). If possible we should apply this in "auto" mode if we can detect that the screen is an e-reader.

brendanlong:
Closes #1017.

Adds an **E-paper** theme alongside Light/Dark, and resolves **Auto** to it when the display looks like an e-reader.

## What it does

- **New `.epaper` palette** in `globals.css`: pure white background, near-black text, and colors chosen to survive a forced grayscale conversion — accents stay colored (blue-800, grays to a clearly dark tone; links remain underlined in prose), borders sit several zinc steps darker than light mode (`--edge` zinc-400 vs zinc-200) because e-ink panels dither faint grays into invisibility, and the primary button / "on" states are pure black-on-white. The block documents these rules for future palette edits.
- **Grayscale-safe overrides** for the few hardcoded colors: Tailwind Typography prose variables, the `reader-prose` `<details>` cards, and the narration highlight (a 30% yellow tint is near-invisible in grayscale, so e-paper uses a solid gray fill).
- **Auto-detection on e-readers**: when the theme is Auto, `EInkSystemThemeOverride` (in `ThemeProvider`) applies the e-paper theme if the `(monochrome)` media query matches or the user agent is a known e-reader/e-ink browser (Kindle, Kobo, EinkBro, Tolino, PocketBook, …). Detection is best-effort — false negatives just mean picking E-paper manually. It re-asserts the class through a MutationObserver because next-themes applies its own theme class from the provider's effect (which runs after child effects) and re-applies it imperatively on system-preference changes.
- **Settings UI**: new E-paper option in Appearance → Color Mode, with "Following system preference (currently e-paper)" shown when Auto resolves to it.
- The system-dark `prefers-color-scheme` CSS fallbacks now exclude `.epaper`; the narration-highlight fallback also now correctly excludes an explicit `.light` class (pre-existing latent bug).
- `.epaper` declares `color-scheme: light` itself since next-themes only sets it for light/dark.
- Demo appearance article and `src/components/CLAUDE.md` updated.

## Verification

- `pnpm typecheck`, `pnpm lint`, `pnpm test:unit` all pass (new unit test covers the e-reader UA detection).
- Verified in headless Chromium against a dev server:
  - explicit `epaper` theme → `<html class="epaper">`, white background, `color-scheme: light`, new palette variables active;
  - Auto + EinkBro user agent → resolves to e-paper;
  - Auto + normal Chrome UA → still resolves to light/dark (no `(monochrome)` false positive).

### DIFF
--- a/src/app/demo/articles/appearance.ts
+++ b/src/app/demo/articles/appearance.ts
@@ -8,13 +8,13 @@ const article: DemoArticle = {
   title: "Appearance & Themes",
   author: null,
   summary:
-    "Customize fonts, text size, alignment, and switch between light and a sleep-friendly, low-blue-light dark theme.",
+    "Customize fonts, text size, alignment, and switch between light, a sleep-friendly low-blue-light dark theme, and an e-paper theme for e-ink screens.",
   publishedAt: new Date("2025-12-28T12:00:00Z"),
   starred: false,
   heroImage: "/demo/appearance.png",
   heroImageAlt:
     "The Lion Reader lion holding a card split into a light half with a sun and a dark half with a moon, with a paintbrush and color swatches.",
-  summaryHtml: `<p>Lion Reader provides comprehensive appearance customization including a <strong>sleep-friendly dark mode</strong> that swaps blue accents for a warm red accent and neutral zinc tones to reduce blue light, multiple font families (serif and sans-serif), adjustable text sizes, and alignment choices. All settings save locally and apply instantly, and the app installs as a Progressive Web App.</p>`,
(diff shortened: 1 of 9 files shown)


brendanlong:
Pushed 994c0179 addressing two gaps an Opus reviewer subagent found in the Auto→e-paper override path:

1. **Stale inline `color-scheme`**: next-themes writes the resolved system theme's `color-scheme` as an inline style, so on an e-ink device with a dark OS preference the e-paper theme rendered with dark form controls/scrollbars. The override now re-asserts `color-scheme: light` alongside the class.
2. **`NarrationHighlightStyles` missed**: it emits its own dark-mode highlight CSS separate from globals.css. Its `prefers-color-scheme: dark` fallback now excludes explicit `.light`/`.epaper` themes, and e-paper gets solid gray highlight arms (the translucent yellow tints are near-invisible in grayscale). Also fixed a pre-existing bug where the `.dark`/media prefixes only applied to the first selector when multiple paragraphs are highlighted.

Re-verified in headless Chromium: dark system pref + e-ink UA on Auto → `class=epaper`, computed `color-scheme: light`, white background; switching to explicit Dark from that state fully restores dark mode.

— Claude

PR #1151, opened 2026-07-12
E-paper: make the page canvas pure white (bg-canvas token)

brendanlong:
Follow-up to #1149 (issue #1017): the e-paper background wasn't actually pure white.

The app shell paints its page background with literal `bg-zinc-50 dark:bg-zinc-950` utilities (`LayoutShell`, the app/demo main scroll containers, and the auth/save/complete-signup/oauth-consent/extension full-page layouts). Those bypass the `.epaper` CSS variables, so the e-paper theme showed a zinc-50 (`#fafafa`) canvas — the slight grey Brendan spotted.

Per the semantic-color convention, this tokenizes the page-canvas color as `--canvas` / `bg-canvas`:

| Theme | `--canvas` |
| --- | --- |
| Light | `#fafafa` (zinc-50, unchanged) |
| Dark | `#09090b` (zinc-950, unchanged) |
| E-paper | `#ffffff` (pure white) |

All twelve canvas call sites now use `bg-canvas`. The extension save/callback pages previously used `dark:bg-zinc-900` where every other page used zinc-950; they now share the standard dark canvas (a subtle darkening, flagging in case it was deliberate). The unread entry-row tint (`bg-zinc-50`) is deliberately untouched — it's a state indicator, not the canvas, and quantizes to white on e-ink anyway.

Verified computed canvas colors in headless Chromium: e-paper `rgb(255,255,255)`, light `rgb(250,250,250)`, dark `rgb(9,9,11)`. Typecheck, lint, and unit tests pass; `src/components/CLAUDE.md` semantic-color table updated.

### DIFF
--- a/src/app/(app)/AppLayoutContent.tsx
+++ b/src/app/(app)/AppLayoutContent.tsx
@@ -181,7 +181,7 @@ export function AppLayoutContent({ initialCursors }: AppLayoutContentProps) {
               <OfflineBanner />
 
               {/* Main content */}
-              <MainScrollContainer className="flex-1 overflow-y-auto bg-zinc-50 dark:bg-zinc-950">
+              <MainScrollContainer className="bg-canvas flex-1 overflow-y-auto">
                 <AppRouter />
               </MainScrollContainer>
             </LayoutShell>
--- a/src/app/(auth)/AuthLayoutContent.tsx
+++ b/src/app/(auth)/AuthLayoutContent.tsx
@@ -15,7 +15,7 @@ interface AuthLayoutContentProps {
 
 export function AuthLayoutContent({ children }: AuthLayoutContentProps) {
   return (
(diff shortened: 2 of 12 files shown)


brendanlong:
An Opus reviewer subagent checked the diff: CSS token wiring, all className edits, the extension dark-canvas normalization (their inner divs have no background fill, so nothing newly stands out), and test assertions all came back clean. It caught one scope gap, fixed in 7394d7bc: `LegalProse` (public terms/privacy pages) used the same literal `bg-zinc-50 dark:bg-zinc-950` pattern and would have stayed slightly grey in e-paper.

`admin/layout.tsx` was deliberately left alone — its light canvas is `bg-white`, which already renders pure white under e-paper, and converting it would change its light-mode color for no e-paper benefit.

— Claude

PR #1147, opened 2026-07-12
Add trigger-maintained unread counters + reconciliation job (#1117 step 5a)

brendanlong:
Step 5a of #1117 — the **write side** of the unread-counter design. Nothing reads the counters yet; step 5b flips the badge queries to arithmetic after the daily reconciliation confirms the triggers are exact in production.

## What ships

- **Migration 0092**: `subscriptions.unread_count` / `starred_unread_count` and `users.saved_unread_count` / `starred_unread_count` (**spam permanently excluded** — `user_entries.is_spam` is immutable, so a row's contribution never changes out from under a counter), three statement-level triggers with transition tables, and an in-migration backfill that runs under the trigger-creation write lock, so counters start exact and stay exact.
- **Trigger design**: bulk writes (fanout `INSERT…SELECT`, mark-all-read, subscribe populate, deletion cascades) cost one grouped counter update per statement. `ON CONFLICT DO NOTHING`-suppressed rows never appear in transition tables; `changedAt`-stale replays update zero rows and contribute nothing. The UPDATE trigger uses signed old/new contributions, so the merge job's `subscription_id` re-stamp moves counts between subscriptions automatically — and unsubscribe/resubscribe need **zero** counter writes (rows keep their stamp; dead-sub counters stay maintained and will feed the "all"-badge starred-orphans term in 5b).
- **`reconcile_counters` singleton job** (daily): recomputes all four counters from ground truth in two single-statement fixes, repairs any drift, and logs fixes at **error** level — the triggers should keep counters exact, so a nonzero fix is a bug signal, not routine self-healing.
- **Rollout/rollback safe by construction**: the triggers live in the database, so writes from the previous release maintain the counters during the canary and indefinitely after a rollback.

## Verification

New integration suite (`tests/integration/unread-counters.test.ts`) exercises every mutation path — subscribe populate, fanout (incl. spam exclusion), read/unread flips + stale-replay no-ops, star/read interaction, mark-all-read, merge re-stamp, saved-article upload + hard-delete cascade, and dead-subscription accuracy — asserting both the expected values and that `reconcileCounters()` finds **zero** fixes afterward (the exact invariant the production job checks daily).

All gates pass locally: typecheck, lint, unit (2190), integration (583), e2e (39).

After this deploys: watch one `reconcile_counters` run report zero fixes, then step 5b flips the reads (and folds in the `user_feeds` cleanup — link checks move to `subscriptions`, the counter joins the view).

Refs #1117

[...]

Issue #1117, opened 2026-07-09
Clean up DESIGN.md and refactor for simplicity and performance


brendanlong:
This will likely need a whole multi-part plan, but this issue documents where we want to go so we can simplify our code, reduce necessary explanations in DESIGN.md and CLAUDE.md, and improve database performance.

This will cover problems and the solution we're considering, although all of these are open to revision if they would cause bigger problems or not solve our intended goals.

## Shared feed + entries

The feed + entries + subscriptions + user_entries setup is very complicated. Maintaining entry linking correctly between users, linking new users to existing feeds, and migrating feeds are all extremely complicated. It's also hard to query this code efficiently.

I think the way to simplify this is to make "feeds" a worker-only concept and "subscriptions" a user-only concept, so entries have a feed and user_entries have a subscription and are *not* linked to a particular feed (the subscription might have a link to the current feed_id, but it's not necessary to find the entries).

The way the "what entries are in this subscription" query is just subscription join user_entries join entries (no feed join).

When a user subscribes to a feed that's already active, we use this algorithm:

1. Check if the feed fetch is expired (will be common if we're using WebSub!); if just link the user's subscription to the feed and re-fetch. Re-fetch should be written to always add a user_entries row if one doesn't exist (insert ... on conflict do nothing).
2. If the feed is not expired, link every entry that was visible in the last fetch to the new user (same insert ... on conflict do nothing).

## Unread counts

Our unread count logic is both complicated and one of the most performance-impactful parts of the site. We have several problems here:

1. The unread query is extremely complicated. See "Shared feed + entries" above. If every entry has exactly one subscription, this query becomes simpler.
2. Calculating unread counts on-demand is inherently inefficient and (to the best of my knowledge can't be fully prevented with indexes. We also don't want to calculate them separately because keeping a count cache in-sync would be incredibly annoying. One option is a fully dedicated service for this, but the simplest option might be to denormalize unread counts onto subscriptions and then maintain them with triggers, as long as this can be done fully-automated with guaranteed correctness (i.e. update within the same statement with appropriate where clauses).
[...]

Issue #1074, opened 2026-07-08
greader subscription/list computes unread counts it discards (skip via includeUnreadCounts)


brendanlong:
## Context

The Google Reader `subscription/list` endpoint (`src/app/api/greader.php/reader/api/0/subscription/list/route.ts`) enumerates subscriptions via `listGreaderSubscriptions` → `subscriptionsService.listSubscriptions`, then maps each through `formatSubscription`. The resulting `GoogleReaderSubscription` shape has **no unread-count field** — `formatSubscription` never reads `Subscription.unreadCount`.

But `listSubscriptions` always computes unread counts: `buildSubscriptionBaseQuery` (`src/server/services/subscriptions.ts`) left-joins an `unreadCountsSubquery` — a `GROUP BY subscription_id` aggregate over `visible_entries WHERE read = false` — into the same query that fetches the subscription list.

So `subscription/list` pays for that aggregate and throws the entire result away.

## Why it's worth fixing

The two costs scale with different things:

- **Subscription listing**: `O(#subscriptions)` — a bounded scan of `user_feeds` (+ tags join). Small.
- **Unread-count aggregate**: `O(#unread entries)` — a grouped aggregate over the `visible_entries` view (which itself applies the `user_entries`→entries→subscription visibility join per row), scaling with the user's **unread backlog** (potentially thousands+).

For a user with a real backlog, the discarded aggregate is plausibly the **dominant cost** of the `subscription/list` query — the subscription scan is the cheap part bolted onto it. (This is reasoning from the query shape, not a measurement; an `EXPLAIN ANALYZE` against a seeded backlog would confirm and give a regression guard.)

Note: this is separate from the cheap `countEntries({ type: "saved" })` that `getSavedSubscription` runs — that's a small count over saved articles and isn't the concern here. The concern is the **real-feed** unread aggregate.

## Proposed fix

Add an `includeUnreadCounts?: boolean` option (default `true`) to `ListSubscriptionsParams` / `buildSubscriptionBaseQuery`. When `false`, omit the `unreadCountsSubquery` left-join, its `groupBy` column, and select `unreadCount` as a literal `0` (use Drizzle's `.$dynamic()` for the conditional join so the column list stays declared once).

Then `listGreaderSubscriptions` (the `subscription/list` path) calls it with `includeUnreadCounts: false` — zero unread-aggregate work — while the `unread-count` endpoint keeps the default `true`.

[...]

PR #1152, opened 2026-07-12
Read unread badges from the denormalized counters (#1117 step 5b)

brendanlong:
Step 5b of #1117: with the trigger-maintained counters (#1147) deployed and the daily reconciliation reporting zero fixes in production, every badge read flips from O(unread-entries) scans to O(subscriptions) arithmetic.

## The algebra (benchmarked at ~0.5ms vs ~90ms scans for a 50k-unread user)

- subscription = `subscriptions.unread_count` · tag/uncategorized = `SUM(unread_count)` over active member subscriptions · starred/saved = `users.starred_unread_count` / `saved_unread_count`
- **all** = `SUM(unread_count)` over active subs + saved + `SUM(starred_unread_count)` over **inactive** subs — the last term is the starred-orphans correction, derived at read time so subscription-state changes need zero counter writes. A new integration test asserts the full algebra end-to-end (active sub + unsubscribed starred orphan + saved article).

## Also in this PR

- **Migration 0093**: `user_feeds` view gains `unread_count` (appended column, `CREATE OR REPLACE VIEW`, old code ignores it — expand/contract-safe). The subscription list reads it directly; `unreadOnly` becomes `unread_count > 0`; the #1074 `includeUnreadCounts` opt-out is deleted (a counter read is free).
- **`user_feeds` is now display-only** (rule documented in DESIGN.md): ownership/scoping checks in `entry-filters.ts` and `markAllEntriesRead` query `subscriptions` directly instead of paying the view's feeds join.
- **Semantics change**: badges now exclude spam everywhere (counters never count spam; previously scans included it). Payload shapes are unchanged, so no frontend changes and `FRONTEND_STATE.md` stands.
- `countEntries`/`countTotalEntries` (generic filtered counts for MCP/Wallabag) deliberately stay as scans.

Net **−73 lines**. All gates pass locally: typecheck, lint, unit (2209), integration (583), e2e (39); the counters-perf equality assertion (scan == arithmetic) and the trigger-exactness suite stay green.

Refs #1117


brendanlong:
Added `27eb8b92` per review follow-up: the top-level sidebar badges now come from the counters too, and **unread counts exclude spam everywhere, unconditionally**.

- `countEntries` loses its `showSpam` parameter — the list can still include spam when the user opts in, but the badge never does. The scan path hard-codes the exclusion.
- The three global badge shapes (`{}`, `{starredOnly}`, `{type:'saved'}`) fast-path to `getGlobalUnreadCounts` (same counter arithmetic as everything else), so the initial page load gets the ~0.5ms path instead of three scans; scoped filters keep the scan.
- Same tRPC input/output, so zero frontend or cache-contract changes. New integration test covers spam exclusion on both paths + fast-path/scan/counter agreement.

All gates re-run green: typecheck, lint, unit (2209), integration (584), e2e (39).

— Claude

PR #1161, opened 2026-07-13
Return a clean 4xx when saving an unfetchable URL (Wallabag + tRPC)

brendanlong:
## Problem

[LION-READER-WEB-G](https://brendan-long.sentry.io/issues/7607379484/): a user saved a URL to Wallabag that 404'd — the URL was `https://laneless.substack.com/p/youre-the-only-person-who-can-do)` with a **trailing `)`** (a markdown-link artifact; the real page exists without the paren). Substack returned 404 for the paren'd path.

Two things went wrong:

1. **It was reported to Sentry as an unexpected error.** `SAVED_ARTICLE_FETCH_ERROR` was classified as `INTERNAL_SERVER_ERROR`. A failed fetch of a *user-provided* URL is a client/input error, not a server bug — but the tRPC timing middleware only exempts `UNAUTHORIZED/NOT_FOUND/BAD_REQUEST/FORBIDDEN` from Sentry, so this 500 was captured.
2. **The Wallabag client got no useful response.** The `saveArticle` `TRPCError` bubbled unhandled out of the `POST /api/wallabag/api/entries` route into Next.js, which returned a 500 (and auto-captured it in Sentry). The Wallabag Android app had nothing to display.

**Is the page broken?** No — it's a valid page; the saved URL just had a stray `)` on the end.

## Fix

- Reclassify `SAVED_ARTICLE_FETCH_ERROR` → `BAD_REQUEST` so it's treated as a client error everywhere (no Sentry noise on the tRPC frontend save path; proper 4xx to callers).
- Catch the thrown `TRPCError` in the Wallabag entries `POST` route and return a clean Wallabag error envelope for **client (4xx)** errors via a new `clientErrorResponse` helper. Genuine 5xx/unknown errors still propagate, so real bugs keep reaching Sentry.

## Test

Adds an e2e test: saving a reachable-but-404 URL returns a 4xx with a Wallabag `{error, error_description}` envelope instead of a 500.

### DIFF
--- a/src/app/api/wallabag/api/entries/route.ts
+++ b/src/app/api/wallabag/api/entries/route.ts
@@ -28,6 +28,7 @@ import { requireAuth } from "@/server/wallabag/auth";
 import {
   jsonResponse,
   errorResponse,
+  clientErrorResponse,
   parseEntryListParams,
   parseBody,
 } from "@/server/wallabag/parse";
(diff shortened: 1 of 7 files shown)


brendanlong:
Addressed the reviewer's Medium finding in 81a63a71.

**Finding:** `SITE_BLOCKED` (the closest sibling to the reported 404 — a target site refusing our fetch bot) maps to HTTP 502, so the original status-range gate re-threw it → still an unhandled 500 to Sentry on the Wallabag path, and it was already reported on the tRPC frontend save path too.

**Fix:** Instead of keying the "should this go to Sentry?" decision on HTTP status range, I introduced an explicit shared concept in `errors.ts` (`isExpectedClientError` / `EXPECTED_CLIENT_ERROR_CODES = {SITE_BLOCKED}`) that decouples *what status to return the client* from *is this a server bug worth reporting* — the approach the reviewer suggested. `SITE_BLOCKED` keeps its honest 502 for clients but is now exempt from Sentry in both the tRPC timing middleware and the Wallabag `clientErrorResponse` helper. Also DRYed the `errorFormatter` onto the new `getAppErrorCode` and added unit tests for the classification.

The Low finding (Google-Docs auth errors via Wallabag) is benign — that branch isn't reachable from the Wallabag route (it passes only `url`/`title`, no `googleDocsAuth`), and even if it were, the result is a clean 4xx, not a 500.

— Claude

Issue #1105, opened 2026-07-08
Sync delta queries are unindexable (cross-table GREATEST scan)


brendanlong:
Follow-up to #1084 (item 2). Item 1 shipped in #1103 (stop bumping `entries.updated_at` on `last_seen_at` refresh), which shrinks the "changed" set these queries scan but does not make them indexable.

## Problem

`sync.events` (`src/server/trpc/routers/sync.ts` — the `GREATEST(entries.updated_at, user_entries.updated_at)` keyset scan, ~lines 256-322) and the Wallabag `since` filter (`src/server/services/entries.ts:598`, via `visible_entries.updated_at = GREATEST(e.updated_at, ue.updated_at)`) both filter and sort on a cross-table `GREATEST(e.updated_at, ue.updated_at)`.

No single index can serve a `GREATEST` over columns from two different tables, so every call is a full per-user scan + sort regardless of `LIMIT`. `sync.events` is the SSE-down polling fallback, so a Redis outage turns every open tab into a repeating full-timeline scan.

## Proposed fix

Split the delta into two index-driven scans and UNION them in SQL:

- **State changes**: `user_entries.updated_at > cursor` served by the existing `idx_user_entries_updated_at` (`user_id, updated_at`).
- **Entry-side (metadata/new) changes**: bound per subscribed feed via a feed-level `feeds.last_entries_updated_at > cursor` pre-filter, then `entries.updated_at > cursor`.

Union the two, then apply the existing keyset pagination `(GREATEST(...), entry_id)` ordering on the merged set.

## Risks / notes

This is a non-trivial rewrite against `sync.events`'s existing invariants — be careful to preserve:
- The `(GREATEST(...), entry_id)` keyset tiebreaker so tied-timestamp groups (mark-all-read, subscribe-time inserts) page correctly (#1080).
- µs-precision SQL classification of `metadataChanged` / `stateChanged` / `isNew` against the same keyset (not JS `Date` math) so selection and categorization never disagree (#1080).
- Fail-closed visibility matching `visible_entries` (active subscription OR starred OR saved).

Wants matching integration coverage (large changed-set pagination + EXPLAIN assertions that the new indexes are used), and the Wallabag `since` path updated to match.

Reported by Claude.

PR #1164, opened 2026-07-13
Google Reader item IDs → stored serial column (#1117 step 6a)

brendanlong:
Step 6a of #1117: replace Google Reader **item IDs** — previously derived at runtime by projecting each entry's UUIDv7 to an int64 (`uuidToInt64`) and reverse-looked-up with an expensive timestamp-window index scan (`batchInt64ToUuid` + `uuidFloor`/`uuidCeil` substring hacks) — with a plain global serial column, `entries.greader_item_id`. An item ID is now just a stored `bigint`; the reverse lookup is a single `greader_item_id = ANY(...)` seek on a unique index.

This deletes the gnarliest part of the compat-ID machinery and eliminates the last-writer-wins collision risk on the old 63-bit projection.

## Scope
- **This PR:** GReader item IDs only.
- **Not touched (later steps):** feed stream IDs, tag sortids, and user IDs still use `uuidToInt64` (6b); the Wallabag 31-bit hash (6c).

## Migration (0094) — rewrite-free
Add nullable `bigint`, sequence + default, in-migration backfill, `SET NOT NULL`, unique index, then recreate `visible_entries` with the column appended. The backfill is **~1.5s** (measured on a 250k-row scale model): the 4.2 GB total relation size is almost all TOAST (content columns), so a `bigint`-column UPDATE rewrites only the small main-heap tuples, not the out-of-line content. Expand-only: the previous release ignores the new column.

## Client impact
GReader clients re-sync from scratch (item IDs change) — accepted; few users, and the app instructs a reset. Feed stream IDs are unchanged, so subscription lists stay stable.

## Threading
`greaderItemId` flows through `visible_entries` → `EntryListItem`/`EntryFull` (mirroring how `wallabag_id` lives in the view). The tRPC/REST output Zod schemas strip unknown keys, so the field never reaches the main app or REST responses (and no bigint-in-JSON serialization concern on the non-superjson REST path). MCP serializes service results directly with `JSON.stringify` (which throws on bigint), so the `list_entries`/`get_entry` handlers strip the field explicitly — only the Google Reader layer, which calls the services directly, sees it.

## Deleted vs kept in `id.ts`
Deleted `batchInt64ToUuid`, `uuidFloor`, `uuidCeil`, `MAX_TIMESTAMP_MS` (the latter three were used only inside `batchInt64ToUuid`; grep-confirmed zero other users). Kept `uuidToInt64`, `parseItemId`, the feed-stream helpers, and `extractTimestamp`/`extractRandomBits`/`timestampHex`/`uuidTimestampIn` (still used by `feedStreamIdToSubscriptionUuid`).

All gates green locally: `db:migrate:local`, typecheck, lint, unit (2209), integration (585), e2e (39). The GReader ID integration test was rewritten for the serial model (old tests asserted UUID→int64 projection math that no longer exists); the e2e `stream/items/ids → stream/items/contents` round-trip exercises real stored serials end-to-end unchanged.

Refs #1117


brendanlong:
Rebased onto master and reviewed. Two real bugs found; both fixed and pushed.

## Rebase (`286dce04` → `6a3a2960`)
Master had taken migration slots 0094 (`idx_entries_feed_updated_at`, #1105) and 0095 (narration paragraph map), so the migration is renumbered to **0096** (file + `_journal.json`, both new `entries` indexes kept in `schema.ts`). The view definition in the migration still matches master's current `visible_entries` — neither intervening migration touched it. The #1105 `sync.events` rewrite merged cleanly with the `entries.ts` changes.

## Review findings

**1. Out-of-range item ids poisoned the reverse-lookup batch** (fixed in `6a3a2960`)
`parseItemId` accepts unbounded hex/decimal input — e.g. the well-formed long id `tag:google.com,2005:reader/item/ffffffffffffffff` parses to 2⁶⁴−1, beyond Postgres's `bigint` range. Passing that into `greader_item_id = ANY(...)` makes Postgres reject the **whole query** (reproduced locally), 500ing `stream/items/contents`/`edit-tag` for the entire batch instead of skipping the bad id. The old `batchInt64ToUuid` guarded exactly this (its comment even says "poisoning the batch"), but the guard and its regression test were deleted in the rewrite. Restored: ids are filtered to the signed-64-bit range before querying, plus the two tests.

**2. MCP `list_entries`/`get_entry` hard-broken by the bigint** (fixed in `fe381dd4`)
The PR body claimed MCP output Zod schemas strip the field — MCP has no output schemas. Both transports (`/api/mcp` and stdio) serialize tool results with plain `JSON.stringify`, which **throws on bigint**, so threading `greaderItemId` through `EntryListItem`/`EntryFull` made every `list_entries`/`get_entry` call fail with an internal error. Fixed by stripping the field in those two handlers (the only MCP tools that return entries — `mark_entries_read`/`star_entries` return their own shapes), with a regression test asserting MCP entry results are JSON-serializable and don't leak the id. PR description updated accordingly.

## Verified fine
- **Migration 0096**: expand-only; old release inserts omit the column (sequence default fills it) and reads use explicit projections, so the appended view column is invisible to it. Applied cleanly on a fresh DB.
- **No other bigint leaks**: tRPC/REST strip via output schemas (superjson would handle it anyway on the browser path); Wallabag/Google Reader formatters build explicit objects; `sync.events` selects explicit fields; no full-row `entries` select→re-insert path exists.
- **No time-ordering assumptions on item ids survive**: `stream/items/ids` ordering/continuation derive from `published_or_fetched_at` cursors and `timestampUsec` from `fetchedAt`, never from the id — the non-time-ordered backfilled serials are fine.
- `uuidToInt64` correctly retained for feed streams/tags/users; `batchInt64ToUuid` has zero remaining callers.

## Gates (local throwaway DB)
typecheck ✓ · lint ✓ · unit 2280 ✓ · integration 626 ✓ (incl. migration 0096 + new regression tests) · greader e2e spec 18 ✓

— Claude 🤖

PR #1166, opened 2026-07-13
Support saving private Google Docs via Wallabag/MCP compat APIs (#1165)

### LINKED ISSUE #1165: Support saving private Google Docs via compat APIs (Wallabag/MCP) when Google auth is already granted
## Summary

Saving a **private** Google Doc only works from the web UI today. The non-interactive save surfaces — the Wallabag API (`POST /api/wallabag/api/entries`) and MCP `save_article` — can't save private docs even when the user has **already linked their Google account with the Docs/Drive scopes**, so no interactive consent is actually needed. We should let those paths use the user's stored credentials, and surface a clean, correct error when auth genuinely isn't set up.

## Background

`saveArticle` (`src/server/services/saved.ts`) handles Google Docs in two tiers:

- **Public / link-shared docs** — fetched by the Google Docs plugin's `savedArticle.fetchContent` (`saved.ts:541`) with **no auth**. These already work from every surface, including Wallabag/MCP. ✅
- **Private docs** — handled by `fetchPrivateGoogleDocWithAuth` (`saved.ts:372`), gated behind `params.googleDocsAuth` (`saved.ts:593`). Only the tRPC `saved.save` router passes `googleDocsAuth: true` (`src/server/trpc/routers/saved.ts:151`). Wallabag (`src/app/api/wallabag/api/entries/route.ts`) and MCP pass only `url`/`title`, so the private path is never attempted and the save falls through to a plain HTML fetch, which fails.

The flag exists because, for users who **haven't** linked Google or granted scopes, `fetchPrivateGoogleDocWithAuth` throws `NEEDS_GOOGLE_SIGNIN` / `NEEDS_DOCS_PERMISSION` / `NEEDS_GOOGLE_REAUTH`, which the **web UI** turns into interactive consent prompts. An API client (Wallabag Android, an MCP client) can't perform that consent dance.

**But** the function already uses *stored* OAuth tokens — `getOAuthAccount(userId, "google")` + `getValidGoogleToken(userId)` (`saved.ts:382`, `430`). So when the user has previously linked Google **and** granted `documents.readonly` + `drive.readonly`, fetching a private doc needs **no interactivity** — it just works. The gate is the only thing preventing it. (`SaveArticleParams.googleDocsAuth`'s own doc comment at `saved.ts:67-75` already notes it will "fetch with the user's OAuth token when already granted.")

## Proposal

1. **Attempt the private-docs path non-interactively on the compat surfaces.** Let the Wallabag (and MCP) save path try `fetchPrivateGoogleDocWithAuth` for Google Docs URLs when the user already has a linked Google account with the required scopes — i.e. decouple "use stored Google credentials" from "can drive an interactive consent prompt". Likely a second flag (e.g. `googleDocsAuth: "non-interactive"` vs the existing interactive mode) rather than overloading the single boolean.

2. **Surface the errors correctly when auth isn't set up.** When the user has *not* linked Google / granted scopes, the compat API should return a clean, truthful error rather than a generic fetch failure or a 500:
   - Wallabag: a proper 4xx envelope (the `NEEDS_*` errors map to `UNAUTHORIZED`/`FORBIDDEN`; `clientErrorResponse` in `src/server/wallabag/parse.ts` already turns those into a clean 401/403 — verify the message is understandable to an API client).
   - MCP: a clear tool error explaining the doc is private and Google Docs access must be granted in the web app first.
   - The message should point the user at the web UI to complete the one-time Google authorization, since the API client can't do it inline.

## Acceptance criteria

- [ ] A user who has linked Google with Docs/Drive scopes can save a **private** Google Doc via the Wallabag API and MCP `save_article` (no interactive step).
- [ ] A user **without** that auth gets a clear, correctly-classified 4xx error (not a 500, not a generic "failed to fetch page") telling them to grant Google Docs access in the web app.
- [ ] Public Google Docs continue to work unchanged on all surfaces.
- [ ] Tests cover both the authorized-success and unauthorized-error paths for at least the Wallabag surface.

## Context

Came up while fixing #1161 (a Wallabag save of an unfetchable URL was a 500 → Sentry). During review it was noted that private-Google-Docs auth errors aren't reachable from Wallabag because the route doesn't opt into the Google path — this issue tracks closing that gap intentionally.

_Filed by Claude._

brendanlong:
Fixes #1165.

## Problem

Saving a **private** Google Doc only worked from the web UI. The non-interactive save surfaces — the Wallabag API (`POST /api/wallabag/api/entries`) and MCP `save_article` — never attempted the private-docs path even when the user had **already linked their Google account with the Docs/Drive scopes**, so those saves fell through to a plain HTML fetch that failed (and, before #1161, could 500).

`saveArticle`'s private-docs path (`fetchPrivateGoogleDocWithAuth`) already uses the user's **stored** OAuth token, so it needs no interactivity once the account is linked and the scopes are granted. The single `googleDocsAuth` boolean gate was the only thing blocking it.

## Change

Turn `SaveArticleParams.googleDocsAuth` from a boolean into a mode:

- **`"interactive"`** (web UI, via `saved.save`): throws the machine-readable `NEEDS_GOOGLE_SIGNIN` / `NEEDS_DOCS_PERMISSION` / `NEEDS_GOOGLE_REAUTH` codes the UI matches to drive in-page consent prompts — **unchanged behavior**.
- **`"non-interactive"`** (Wallabag + MCP): attempts the same stored-credential fetch, but when auth isn't set up throws the **same** `UNAUTHORIZED`/`FORBIDDEN` (4xx) classification with a **human-readable message** pointing the user at the web app to complete the one-time Google authorization.

Both modes attempt exactly the same token fetch; they differ only in the message thrown when auth isn't configured. Wallabag's existing `clientErrorResponse` already turns those 4xx TRPCErrors into clean 401/403 envelopes, so a user without auth now gets a correctly-classified 4xx instead of a generic fetch failure or a 500.

## Acceptance criteria

- [x] A user with Docs/Drive scopes can save a private Google Doc via Wallabag/MCP (no interactive step) — same stored-credential fetch the web UI uses, just decoupled from the interactive gate.
- [x] A user without that auth gets a clear, correctly-classified 4xx (401 not-linked / 403 missing-scopes) telling them to grant Google Docs access in the web app.
- [x] Public Google Docs continue to work unchanged on all surfaces (plugin path is untouched).
- [x] Tests cover the unauthorized-error paths for the save service (not-linked → 401, missing-scopes → 403) plus the interactive-mode web-UI contract.

## Tests

`tests/integration/saved-google-docs-auth.test.ts` — exercises the auth gate deterministically without network (the public-docs plugin fetch no-ops when no service account is configured, so `saveArticle` falls straight through to the private-docs path). The authorized-success path (a live private-doc fetch) needs a real Google token + network, so it isn't covered by an automated test.

### DIFF
--- a/src/app/api/wallabag/api/entries/route.ts
+++ b/src/app/api/wallabag/api/entries/route.ts
@@ -118,6 +118,10 @@ export async function POST(request: Request): Promise<Response> {
     const article = await savedService.saveArticle(db, auth.userId, {
       url: articleUrl,
       title: body.title || undefined,
+      // Use the user's stored Google credentials for private Google Docs when
+      // already linked/granted (no interactive consent possible from an API
+      // client). Falls back to a clean 4xx pointing at the web app otherwise.
+      googleDocsAuth: "non-interactive",
     });
 
     // Handle archive/starred flags if provided
--- a/src/server/discord/bot.ts
+++ b/src/server/discord/bot.ts
@@ -484,7 +484,13 @@ async function handleSaveReaction(
 
   for (const url of urls) {
     try {
-      const saved = await saveArticle(db, resolved.userId, { url });
+      // Discord is a non-interactive surface like Wallabag/MCP: use the user's
+      // stored Google credentials for private Google Docs when already
+      // linked/granted (the failure just surfaces as the error reaction below).
+      const saved = await saveArticle(db, resolved.userId, {
+        url,
+        googleDocsAuth: "non-interactive",
+      });
       logger.info("Saved article via Discord", {
         userId: resolved.userId,
         discordUser: user.tag,
--- a/src/server/mcp/tools.ts
+++ b/src/server/mcp/tools.ts
@@ -310,13 +310,18 @@ function buildTools(): Tool[] {
     {
       name: "save_article",
       description:
-        "Save a URL for later reading. Fetches the page, extracts clean content using Readability, and stores it. Returns the saved article if already saved.",
+        "Save a URL for later reading. Fetches the page, extracts clean content using Readability, and stores it. Returns the saved article if already saved. Private Google Docs are supported when the user has linked their Google account and granted Google Docs access in the web app; otherwise a clear error explains how to authorize.",
       inputSchema: toInputSchema(saveArticleArgs),
       handler: async (db, userId, args) => {
         const params = parseArgs(saveArticleArgs, args);
         return savedService.saveArticle(db, userId, {
           url: params.url,
           title: params.title,
+          // Use the user's stored Google credentials for private Google Docs
+          // when already linked/granted; otherwise throw a clear error telling
+          // them to authorize Google Docs access in the web app (an MCP client
+          // can't run the interactive consent flow).
+          googleDocsAuth: "non-interactive",
         });
       },
     },
--- a/src/server/services/saved.ts
+++ b/src/server/services/saved.ts
@@ -65,14 +65,24 @@ export interface SaveArticleParams {
   /** With refetch, update even if the new content appears lower quality. */
   force?: boolean;
   /**
-   * Enable the interactive private-Google-Docs auth flow: when a Google Docs
-   * URL can't be fetched publicly, throw NEEDS_GOOGLE_SIGNIN /
-   * NEEDS_DOCS_PERMISSION / NEEDS_GOOGLE_REAUTH errors that the web UI turns
-   * into consent prompts (or fetch with the user's OAuth token when already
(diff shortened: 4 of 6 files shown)


brendanlong:
Follow-up from review: also extended the non-interactive Google Docs auth to the **Discord bot** save-via-reaction path (`src/server/discord/bot.ts`), since it's the same kind of non-interactive client as Wallabag/MCP. A missing-auth failure just surfaces as the existing error reaction. (a15f58eb)

— Claude

PR #1181, opened 2026-07-13
docs: main agent owns local services; subagents shouldn't start their own

brendanlong:
Adds an **Ownership** note to CLAUDE.md's "Local Services (no Docker)" section.

**Why:** a subagent's background tasks are torn down when the subagent finishes, so a subagent that runs `pnpm services` itself loses the throwaway Postgres/Redis out from under any later work (and can orphan the DB processes). The lifecycle belongs to the persistent main agent.

**Rule documented:** the main agent starts and keeps `pnpm services` running (restarting if it dies); subagents assume services are already up and use the `*:local` scripts, which read the `.env.local-services*` files the main agent wrote. If services drop mid-run, the subagent reports back rather than starting its own, and the main agent restarts.

Also cross-references clawed-abode#424 (the harness periodically SIGTERMs background tasks even for the main agent), so a mid-run DB-connection error should be treated as "restart and retry", not a code failure.

Docs only — no code change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -36,6 +36,22 @@ it stops both servers and deletes its temp dir + env files:
 pnpm services            # run in the BACKGROUND; leave it running
 ```
 
+**Ownership when using subagents:** the persistent **main agent** owns `pnpm
+services` — it starts it and keeps it running (restarting if it dies).
+**Subagents must not run `pnpm services` themselves**: a subagent's background
+tasks are torn down when the subagent finishes, which kills the DB out from under
+later work (and can orphan the Postgres/Redis processes). A subagent should assume
+the services are already up and just use the `*:local` scripts below — they read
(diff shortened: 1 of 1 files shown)


brendanlong:
Closing — a follow-up test disproved the premise this note was based on.

I had a persistent agent (main) start \`pnpm services\`, stopped them, then had a **subagent** start them and finish. The services the subagent started **survived the subagent finishing** — the cluster stayed alive and DB-reachable for 14+ minutes (42/42 liveness checks) after the subagent exited, and is still up. So "a subagent's background tasks are torn down when the subagent finishes" is false.

The earlier "services keep dying every few minutes" behavior (which hit the main agent's \`pnpm services\` too, not just subagents') was a host-level issue that a reboot cleared — not a subagent-lifecycle thing. Since the technical rationale doesn't hold and the reaping is gone, there's no correctness reason to forbid subagents from starting services, so this doc note would be misleading. Dropping it.

— Claude

Issue #1178, opened 2026-07-13
Codemod remaining raw zinc (~276) and stray blue/purple/indigo utilities onto semantic tokens


brendanlong:
## Problem

Follow-up from #1169 / PR #1175, which tokenized the **status** color layer (danger/success/warning/star) and left the neutral and one-off brand colors for a separate sweep to keep that PR focused. Two categories of raw utilities remain in `src/components` + `src/app`:

### 1. Raw `zinc-*` — ~276 occurrences across ~52 files

The neutral semantic tokens (`text-strong`/`text-muted`/`bg-surface`/`border-edge`/…) already exist and are documented in `src/components/CLAUDE.md` → "Semantic Colors"; these call sites just never got migrated. Most are exact pairs with a clean 1:1 mapping, e.g.:

- `text-zinc-900 dark:text-zinc-50` → `text-strong`
- `text-zinc-600 dark:text-zinc-400` → `text-muted`
- `text-zinc-700 dark:text-zinc-300` → `text-body`
- `text-zinc-400 dark:text-zinc-500` → `text-faint`
- `bg-white dark:bg-zinc-900` → `bg-surface`
- `bg-zinc-100 dark:bg-zinc-800` → `bg-surface-muted`
- `bg-zinc-50 dark:bg-zinc-950` → `bg-canvas`
- `border-zinc-200 dark:border-zinc-800` → `border-edge`
- `border-zinc-300 dark:border-zinc-700` → `border-edge-input`

**Deliberate exceptions to keep (allow-list, don't convert):**

- The `active:` press step is intentionally raw: `active:bg-zinc-100 dark:active:bg-zinc-700` (documented in CLAUDE.md as one step darker than the `surface-muted` hover).
- The `btn-primary` utility's `hover:bg-zinc-600 … dark:hover:bg-zinc-300 …` deltas in `globals.css`.
- Placeholder-only shades and any pair that has no exact token equivalent — flag rather than force-fit; add a token if a real gap emerges (don't invent drift).

### 2. Stray `blue`/`purple`/`indigo` — ~50 occurrences across 8 files

Not part of the status palette, so #1169 left them untouched. These split into two sub-cases:

[...]

Issue #1169, opened 2026-07-13
Tokenize the status-color layer (success/warning/danger/star) and codemod raw color utilities


brendanlong:
## Problem

The neutral/accent/`--info` layer of our color system is cleanly tokenized (`--text-*`, `--surface-*`, `--edge-*`, mapped through `@theme inline` in `src/app/globals.css`), but the **status-color layer was never tokenized**. There are **no `--success`, `--warning`, or `--danger` tokens** — only `--info`. Every success/warning/error/star is a hand-rolled raw Tailwind pair, copy-pasted and drifted.

Measured across `src/components/` + `src/app/`:

- **9 distinct red text shades** for error/danger (`text-red-100`…`text-red-900`)
- **16 distinct green combos** for success — including `text-green-600`, which **fails WCAG AA on white** (~3.2:1). Success text is under-contrast in places.
- **Warning is two hues at once** — both `amber-*` and `yellow-*` for the same role (~30 shade variants).
- **~403 raw non-neutral color utilities** total, plus **~274 raw `zinc-*`** still in components that bypass the existing neutral tokens.
- The **star** is un-tokenized amber drift (`text-amber-500 dark:text-amber-400` in `EntryListItem`, plus amber/yellow elsewhere).

This is the main source of the "our styles feel random" problem. It's the same refactor already done for neutrals, just never applied to semantics.

## Proposal

Add token trios per status role, mirroring the existing `--info` pattern, themed once each in `:root` / `.dark` / `.epaper` and exposed via `@theme inline`:

```css
--danger  / --danger-subtle  / --danger-subtle-foreground  / --danger-border
--success / --success-subtle / --success-subtle-foreground / --success-border
--warning / --warning-subtle / --warning-subtle-foreground / --warning-border
--info    (exists)
--star    (formalize the amber star)
```

Dark mode keeps status colors their conventional hues (red/green/amber/blue-400) — extreme-low-blue-light is about dominant surfaces and the accent, not about forbidding a red error message. E-paper pairs each with contrast per existing rules; keep greys cool (see accent issue).

Then codemod the raw utilities onto tokens: `StatusCard`, `Alert`, `Input` error state, and every `bg-red-50 dark:bg-red-950 text-red-800 …` collapses to `bg-danger-subtle text-danger-subtle-foreground border-danger-border`, exactly like the neutral tokens. Fix the `text-green-600` AA contrast bug (use `green-700`) as part of the sweep.

[...]

PR #1184, opened 2026-07-13
Design system: canonical color tokens + component set + migration plan

brendanlong:
Adds [`docs/design/design-system.md`](docs/design/design-system.md) — the canonical, minimal role-based token set, the component inventory, and the plan to finish moving the app onto a consistent set of colors. Follow-on from the accent exploration (#1172).

## What it establishes

- **The minimal token set**, grouped by role, folding in the warm-amber accent. Flags the handful of redundant tokens (`--info-muted`, `--info-foreground`, unused `-solid-hover`) as trim candidates rather than churning them.
- **Components are ~done.** Only two changes needed: rework the `btn-primary` utility so it derives hover/active from `--primary-solid` (today it hardcodes raw zinc, which would break a branded amber primary), and add a `Button` `danger` variant (≈6 files hand-roll destructive red buttons today). Everything else (`StatusCard`, `Alert`, `Input`, `Card`, `NavLink`, `TextLink`, star) is already token-driven after #1169/#1178.
- **The overlap decision:** warning and accent are both amber; accepted because they never share a treatment (accent = links/dots/focus, warning = icon + bordered box). Yellow is the one-file lever if in-situ review shows ambiguity.

## The plan (each phase ships independently)

1. **Accent re-hue + branded primary** — one PR, essentially `globals.css` only; the whole app re-hues from it because it's fully tokenized.
2. **`Button` `danger` variant** + migrate the hand-rolled red buttons.
3. **Finish the neutral mop-up** — ~172 stray `zinc-*` → tokens (same codemod as #1178).
4. **CI guardrail** — block new raw color utilities so drift can't return.
5. *(optional)* trim verified-unused tokens.

Filing issues for phases 1–4. The big finding: the migration is already ~80% done, so "moving the app to the consistent set" is now mostly Phase 1 (one file) + mechanical cleanup.

### DIFF
--- a/docs/design/design-system.md
+++ b/docs/design/design-system.md
@@ -1,0 +1,177 @@
+# Lion Reader design system — colors & components
+
+The canonical, minimal set of theme colors and components, and the plan to finish
+moving the app onto it. Companion to the accent rationale in
+[`accent-exploration.md`](accent-exploration.md).
+
+Principle: **one token per role, themed once.** A component never writes a raw
+Tailwind color (`bg-red-50 dark:bg-red-950`); it writes a role (`bg-danger-subtle`),
+and the three themes (`:root` / `.dark` / `.epaper` in `src/app/globals.css`) supply
+the value. Retheming is editing `globals.css`, never call sites.
+
+Most of this already exists. The status layer (`--danger/--success/--warning/--star`),
+neutrals, surfaces, and edges are tokenized and consumed by `StatusCard`, `Alert`,
+`Input`, etc. This doc pins down the **target** set (including the amber accent that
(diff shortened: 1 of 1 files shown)


brendanlong:
Phase issues filed:

- Phase 1 — Adopt warm-amber accent + branded primary: #1185
- Phase 2 — Button `danger` variant + migrate destructive buttons: #1186
- Phase 3 — Finish neutral zinc → token mop-up (~172): #1187
- Phase 4 — CI guardrail against raw color utilities: #1188

Phase 5 (optional token trim) is left as a checklist note in the doc rather than a tracked issue.

— Claude

Issue #1080, opened 2026-07-08
sync.events delta endpoint: tie data-loss, fail-open visibility, duplicate events


brendanlong:
Found in the repo-wide audit (verified in code; two agents independently flagged the visibility half).

The offline/polling delta endpoint `sync.events` (`src/server/trpc/routers/sync.ts:243-320`) hand-rolls a query that diverges from `visible_entries`/`listEntries` in three ways:

### 1. Data loss on timestamp ties (correctness, high)
Pages by `GREATEST(entries.updated_at, user_entries.updated_at)` with **no id tiebreaker** and a strict `>` cursor (`sync.ts:259-267`). `markAllEntriesRead` stamps one identical timestamp on every affected row, and the subscribe-time insert does the same. After a "mark all read" over >500 entries, catch-up returns 500 rows all sharing that timestamp with `hasMore: true`; the client advances its cursor to that timestamp and the strict `>` then **excludes every remaining tied row permanently**. Ordering among ties is also nondeterministic, so the 500 aren't even a stable prefix.
- Fix (maintainer agrees): keyset-paginate on `(GREATEST(...), entry_id)` — add `entry_id` to ORDER BY, carry `(ts, lastId)` in the cursor, use row comparison `(ts, id) > (cursor_ts, cursor_id)`. Same pattern `listEntries` already uses.
- Sub-bug at `sync.ts:319-320`: ms-truncated JS `Date` compared against µs-precision SQL selection means a change within the same millisecond as the cursor is selected but emits **no** event and the cursor never advances past it.

### 2. Fail-open visibility predicate (correctness, medium)
`sync.ts:263` uses `(subscriptions.unsubscribed_at IS NULL OR starred)` over LEFT JOINs. For an orphaned `user_entries` row (no `subscription_feeds` match), `NULL IS NULL = TRUE`, so it passes — the exact case migration `0073` changed the view to fail-closed to hide. Sync-fed clients see entries the web app never shows.

### 3. Duplicate events via overlapping `subscription_feeds` (correctness, low)
The LEFT JOIN can match multiple `subscription_feeds` rows per (user, feed), emitting duplicate events per entry.

## Suggested fix
Query `visible_entries` directly (or replicate its fail-closed predicate + `DISTINCT`), which fixes #2 and #3 in one move, and add the `(sortkey, id)` keyset for #1.

PR #1191, opened 2026-07-13
Make sync.events delta index-driven (UNION), fixing the cross-table GREATEST scan (#1105)

### LINKED ISSUE #1105: Sync delta queries are unindexable (cross-table GREATEST scan)

brendanlong:
Fixes #1105.

## Problem

`sync.events` filtered and sorted on `GREATEST(entries.updated_at, user_entries.updated_at)` — a value spanning two tables. No index can serve a `GREATEST` over columns from two tables, so **every call scanned + sorted the user's entire history**, and `LIMIT` gave no help (the sort must materialize all rows before it can order + limit). `sync.events` is the SSE-down polling fallback, so a Redis outage turned every open tab into a repeating full-timeline scan.

This was **not** a global full-table scan — the `user_id` predicate bounds it to one user — but within a user it was a full scan of their whole `user_entries` set + nested-loop joins + a full sort, cost ∝ lifetime entry count rather than the changed-since-cursor delta.

## Fix

Since `GREATEST(e, ue) > cursor  ⟺  e.updated_at > cursor OR ue.updated_at > cursor`, the delta is built from index-driven arms and `UNION`ed, then decorated + keyset on the (now bounded) candidate set:

- **Arm A** — `user_entries.updated_at` (`idx_user_entries_updated_at`). Covers every state change (read/star flips, mark-all-read) **and new entries**: the feed fanout inserts `user_entries` rows with `updated_at = now()`, so new entries ride this arm too.
- **Arm B1** — `entries.updated_at` within the user's **subscribed** feeds. Catches content refetches that bump `entries.updated_at` **without** touching the `user_entries` row (`updateEntryContent`) — the case Arm A misses. Pre-filtered to feeds whose `last_entries_updated_at` moved (it's stamped `>=` every `entry.updated_at` in the feed, so it never drops a changed entry), then seeks the new `idx_entries_feed_updated_at`.
- **Arm B2** — same, for the user's saved-articles feed (no subscription row, never polled → `last_entries_updated_at` stays NULL), keyed by feed id.

Arms compare with `>=` so a tied-timestamp boundary row stays a candidate; the outer query re-applies the exact `(GREATEST, id)` keyset, preserving the #1080 invariants (keyset tiebreaker, µs-precision SQL categorization, fail-closed visibility).

## New index

`migrations/0094_entries_feed_updated_at_index.sql` — `entries(feed_id, updated_at)`, serving both entry-side arms. (Header notes the CONCURRENTLY-out-of-band convention for large prod tables.)

## Benchmark

1 user, 300 feeds, 60k entries, realistic 10-row delta (`EXPLAIN ANALYZE`, local Postgres):

| | Execution | Shared buffers | Plan |
|---|---|---|---|
| **Old** | 28.8 ms | 2702 | parallel seq scan of 60k `user_entries` + 60k `entries` → hash joins → **sort** |
| **New** | 0.13 ms | 177 | `idx_user_entries_updated_at` + `idx_entries_feed_updated_at` seeks → dedup → PK decorates |

**~220×** faster; work now scales with the delta, not the history. (The benchmark's `feeds` table is tiny/single-user so the planner seq-scans it in Arm B1; on a large global `feeds` table the planner drives Arm B1 from the user's subscriptions via `uq_subscriptions_user_feed (user_id, feed_id)`, bounding to the user first.)

## Wallabag `since`

**Left as-is — it's already correct.** Its filter is the same `GREATEST(entry.updated_at, user_entries.updated_at)`, so it *does* detect entry-side content updates (covered by an existing test in `entries.test.ts`). It also doesn't hit the acute problem: it's scoped to saved articles and sorts by `publishedOrFetchedAt` with `LIMIT` pushdown, so `GREATEST` is only a **residual filter** there, not the sort key. Updated the code comment to explain why it deliberately keeps the simple form.

## Tests

- New `entry-side changes (#1105)` block: content refetch delivered via Arm B1 and Arm B2, starred-orphan (unsubscribed) visibility, and the non-starred-unsubscribed negative case.
- `EXPLAIN` assertions that each candidate arm seeks its index (no seq scan), mirroring the router arms.
- Full `sync-events` + `entries` integration suites pass (65 tests).

### DIFF
--- a/migrations/0094_entries_feed_updated_at_index.sql
+++ b/migrations/0094_entries_feed_updated_at_index.sql
@@ -1,0 +1,22 @@
+-- Index entries(feed_id, updated_at) for the sync.events delta rewrite (#1105).
+--
+-- sync.events used to filter and sort on GREATEST(entries.updated_at,
+-- user_entries.updated_at) — a value spanning two tables, so no index could
+-- serve it and every call scanned + sorted the user's entire history. The
+-- rewrite splits the delta into index-driven arms and UNIONs them; the
+-- entry-side arm (content refetches that bump entries.updated_at WITHOUT
+-- touching the user_entries row) needs to seek changed entries within a feed:
+--
+--   entries.feed_id = <subscribed feed> AND entries.updated_at >= cursor
+--
+-- This index serves that seek. It also serves the saved-articles arm, which
+-- keys on the user's saved feed id directly (saved feeds are never polled, so
+-- their feeds.last_entries_updated_at stays NULL and can't pre-filter them).
+--
+-- NOTE: on a large `entries` table, build this CONCURRENTLY out-of-band first
+-- (CREATE INDEX CONCURRENTLY can't run inside the migration runner's
+-- transaction); this IF NOT EXISTS statement then no-ops there and only builds
+-- the index on databases that don't have it yet.
+
+CREATE INDEX IF NOT EXISTS idx_entries_feed_updated_at
+  ON entries (feed_id, updated_at);
--- a/migrations/meta/_journal.json
+++ b/migrations/meta/_journal.json
@@ -652,6 +652,13 @@
       "when": 1782952600000,
       "tag": "0093_user_feeds_unread_count",
       "breakpoints": true
+    },
+    {
+      "idx": 93,
+      "version": "7",
+      "when": 1782952700000,
+      "tag": "0094_entries_feed_updated_at_index",
+      "breakpoints": true
     }
   ]
 }
--- a/migrations/schema.sql
+++ b/migrations/schema.sql
@@ -719,6 +719,8 @@ CREATE INDEX idx_entries_feed_published_coalesce ON public.entries USING btree (
 
 CREATE INDEX idx_entries_feed_type ON public.entries USING btree (feed_id, type);
 
+CREATE INDEX idx_entries_feed_updated_at ON public.entries USING btree (feed_id, updated_at);
+
 CREATE INDEX idx_entries_fetched ON public.entries USING btree (feed_id, fetched_at);
 
 CREATE INDEX idx_entries_last_seen ON public.entries USING btree (feed_id, last_seen_at) WHERE (type = 'web'::public.feed_type);
--- a/src/server/db/schema.ts
+++ b/src/server/db/schema.ts
@@ -513,6 +513,10 @@ export const entries = pgTable(
     index("idx_entries_spam").on(table.feedId, table.isSpam),
     // For filtering by entry type
     index("idx_entries_type").on(table.type),
+    // sync.events / Wallabag delta: seek changed entries within a feed
+    // (feed_id = <subscribed/saved feed> AND updated_at >= cursor). Backs the
+    // entry-side arm of the sync.events UNION rewrite (#1105). Migration 0094.
+    index("idx_entries_feed_updated_at").on(table.feedId, table.updatedAt),
     // Expression indexes for entry list sorting (Drizzle can't express these, created via raw SQL):
     // - idx_entries_published_coalesce: (COALESCE(published_at, fetched_at) DESC, id DESC)
     //   Enables limit pushdown for "all entries" queries. Migration 0060.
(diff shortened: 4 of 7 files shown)


brendanlong:
Follow-up: an independent correctness review caught a real bug in the first version of Arm B1, now fixed in 959367e1.

**The bug.** Arm B1 pre-filtered subscribed feeds on `feeds.last_entries_updated_at >= cursor`, on the premise that the column is stamped `>=` every `entry.updated_at` in the feed. That premise is false: `feeds.last_entries_updated_at` is written from the poll's **start-time `now`** (`handlers.ts` captures it before the HTTP fetch), while each changed entry's `updated_at` is a **later** wall-clock write in `createEntry`/`updateEntryContent` (after fetch+parse). So `entry.updated_at > feeds.last_entries_updated_at` by the fetch duration.

**Impact.** Once a client's cursor advanced past `last_entries_updated_at` but not past the entries — e.g. paging across the `MAX_ENTRIES` (500) boundary of a >500-entry same-poll content burst in one feed — the pre-filter pruned the feed and the tail of that burst was dropped from the delta **permanently** (until an unrelated state change on each entry). Arm A can't recover it (a content refetch leaves `user_entries.updated_at` stale) and it isn't saved, so Arm B2 misses it too. Latent (needs a large same-poll burst) but silent permanent loss — worth fixing.

**Fix.** Removed the pre-filter and the now-purposeless `feeds` join. Arm B1 drives from the user's subscriptions straight into `idx_entries_feed_updated_at`, seeking `(feed_id, updated_at >= cursor)` per feed — which already bounds the work to the delta without the unsafe timestamp prune. Note Arm B2 (saved) never used the pre-filter, keying on the feed id directly — the asymmetry is what flagged it.

Replaced the misleading "pre-filter is a superset" test with a regression test where `last_entries_updated_at` **lags** the entry's `updated_at` (as production always produces); confirmed it fails with the old pre-filter and passes without it.

**Updated benchmark** (1 user, 300 feeds, 60k entries, 10-row delta): 25.2 ms → 0.19 ms (**~133×**), 2702 → 238 shared buffers, 0 seq scans (all index seeks). Slightly more buffers than the pre-filtered version (177) since Arm B1 now seeks per subscribed feed, but still bounded by subscription count, not entry count.

— Claude

PR #1214, opened 2026-07-14
Fix broad Wallabag token invalidation + add Wallabag API logging

brendanlong:
## Why

Investigating reports that the Wallabag API was flaky on **setup and sync**, I found two problems.

**1. There were essentially no logs.** The Wallabag route handlers emit nothing on the request path — success, `401`/`403`, `429`, and handled `4xx` all just return a `Response`. Only an unhandled `500` reaches Sentry (and nothing was there). The proxy's `LOG_MCP_REQUESTS` matcher doesn't cover `/api/wallabag/*` either. So the failures were invisible.

**2. Refresh-token reuse detection had a far too broad blast radius.** On reuse of a rotated refresh token, `handlePossibleRefreshTokenReuse` revoked **every** active token for the token's `user + client_id`. Every Wallabag credential is minted under the single shared `client_id "wallabag"`, so one stale-token replay — a lost refresh response the client retries, or a second device — signed the user out of **every** Wallabag session at once. That matches "flaky setup/sync that needs re-login" exactly.

### Scope reuse revocation to the leaked chain (`src/server/oauth/service.ts`)
Revoke only the leaked token's rotation **family**, per OAuth 2.1 guidance (RFC 9700 §4.14.2), instead of all `user+client` tokens. The chain is strictly linear (each rotation atomically claims one predecessor and stamps `replaced_by_id`), so a recursive walk of `replaced_by_id` forward from the presented token reaches whichever branch is live; ancestors are already revoked. Sibling grants for the same user (other devices) are left untouched. No schema change. The bare `console.warn` becomes a structured `logger.warn` with the revoked counts.

New integration test asserts a reuse event on one grant spares an independent sibling grant for the same `user+client`; the existing chain-revocation test still passes.

### Add Wallabag observability logging (`src/server/wallabag/auth.ts`, token route)
Structured logs (`component: "wallabag"`) at the choke points:
- **token endpoint** — per-IP and per-account rate-limit rejections (tests the "429 on setup" hypothesis)
- **passwordGrant / refreshTokenGrant** — grant outcomes with a `reason` on failure (an `invalid_or_revoked` refresh is the key flaky-sync signal)
- **requireAuth** — auth failures, distinguishing `missing_bearer` from `invalid_token` (expired-token 401s are expected churn, logged at `info`, not `warn`)

No tokens, passwords, or emails are logged.

## Testing
- `pnpm typecheck`, `pnpm lint` clean
- `pnpm test:integration` — 625 passed (incl. new sibling-grant isolation test)

Once deployed, filter production logs with `"component":"wallabag"` to see grant outcomes and rate-limit hits, and `"OAuth refresh token reuse detected"` to catch any remaining reuse events.

### DIFF
--- a/src/app/api/wallabag/oauth/v2/token/route.ts
+++ b/src/app/api/wallabag/oauth/v2/token/route.ts
@@ -25,13 +25,23 @@ import { passwordGrant, refreshTokenGrant } from "@/server/wallabag/auth";
 import { parseBody } from "@/server/wallabag/parse";
 import { jsonResponse, errorResponse } from "@/server/wallabag/parse";
 import { checkRouteRateLimit, checkAccountRouteRateLimit } from "@/server/rate-limit";
+import { logger } from "@/lib/logger";
 
 export const dynamic = "force-dynamic";
 
 export async function POST(request: Request): Promise<Response> {
   // Per-IP limit first (cheap, rejects floods before parsing).
   const rateLimitResponse = await checkRouteRateLimit(request, "expensive", { json: true });
-  if (rateLimitResponse) return rateLimitResponse;
+  if (rateLimitResponse) {
+    // Setup/sync surfaces the token endpoint through the strict per-IP
+    // `expensive` bucket; a 429 here (e.g. several devices behind one NAT, or a
+    // client retrying) reads to the user as flaky setup. Log it so we can tell.
+    logger.warn("Wallabag token request rate-limited", {
+      component: "wallabag",
+      bucket: "expensive_per_ip",
+    });
+    return rateLimitResponse;
+  }
 
   const body = await parseBody(request);
 
@@ -49,7 +59,14 @@ export async function POST(request: Request): Promise<Response> {
     // Per-account limit: throttles distributed, IP-rotating brute-force against
     // a single account, shared with the tRPC login and Google Reader paths.
     const accountRateLimitResponse = await checkAccountRouteRateLimit(username, { json: true });
-    if (accountRateLimitResponse) return accountRateLimitResponse;
+    if (accountRateLimitResponse) {
+      logger.warn("Wallabag token request rate-limited", {
+        component: "wallabag",
+        bucket: "expensive_per_account",
+        grantType: "password",
+      });
+      return accountRateLimitResponse;
+    }
 
     const result = await passwordGrant(username, password, clientId);
     if (!result) {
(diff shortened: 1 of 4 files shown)


brendanlong:
## Follow-up: fixed a HIGH from review + the actual dominant flakiness cause

An independent review flagged that narrowing reuse revocation to the rotation chain introduced a gap: `rotateRefreshToken` wasn't transactional, so a crash between minting the successor and stamping `replaced_by_id` could leave a **live-but-unlinked** successor the chain-walk can never reach — defeating reuse detection. While fixing that I probed the concurrency path and found something bigger.

**Empirically confirmed the real flakiness cause.** With a throwaway probe against a local DB, 5 concurrent rotations of the same refresh token (exactly what Wallabag clients do — they batch parallel sync requests, each triggering a refresh) left the **winning** request's brand-new tokens dead:

```
Before fix:  ACCESS_VALID=false  REFRESH_ROTATES=false   <- client logged out after a successful refresh
After fix:   ACCESS_VALID=true   REFRESH_ROTATES=true
```

The losing requests present the just-revoked token a few ms later; reuse detection was treating that as a leak and revoking the winner's fresh chain. This existed on master too (the old broad revoke), and chain-scoping alone didn't fix it (the new chain is linked forward from the replayed token).

### Changes in the new commit (`66a3978`)
1. **Transactional rotation** — claim + mint + link now run in one `db.transaction`; either the fully-linked successor exists or nothing does (closes the review's crash-window gap). `createTokens` gained an optional executor to enroll in the caller's tx.
2. **Grace window for reuse detection** — a rotated token replayed within 10s of its revocation is treated as benign concurrency and does **not** revoke the live successor. A genuine stale-token replay lands outside the window and is still caught + its chain revoked.

### Trade-off worth a look
The grace window is a deliberate, small security relaxation: a stolen token that was *already successfully rotated by the attacker* won't trigger proactive chain revocation if the legit client happens to refresh within 10s of the attacker's rotation (it's still caught on any later replay). For a read-it-later API I think availability wins decisively here — the status quo logs clients out on nearly every concurrent refresh — but flagging it in case you'd want the window shorter (or longer).

### Tests
- winner survives 5-way concurrent rotation
- immediate replay is benign (no revocation)
- replay backdated past the grace window still revokes the leaked chain, sparing sibling grants
- full oauth/wallabag/greader integration set green (`tsc` + `lint` clean)

— Claude

PR #1218, opened 2026-07-14
Support narration OS media controls in PWA (#410)

### LINKED ISSUE #410: Support narration media controls in PWA
I'm unsure if we need to do anything special here, but our narration should support standard native media controls when installed as a PWA, including the standard controls notification and play/pause from bluetooth device buttons.

brendanlong:
Closes #410.

## Problem

Narration was supposed to expose native OS media controls when installed as a PWA (lock-screen/notification widget, and play/pause + track buttons from Bluetooth/headphone devices), but it didn't work:

- Narration plays through the **Web Speech API** (browser voices) or the **Web Audio API** (Piper enhanced voices, via `AudioBufferSourceNode`). **Neither registers as media playback** with the browser, so setting `navigator.mediaSession` metadata + action handlers had **no visible effect** — the OS controls never appeared and hardware buttons weren't routed to us.
- Media Session was also only wired for **browser voices**, explicitly skipped for Piper.

## Fix

- **`src/lib/narration/silent-audio.ts`** (new): plays a **silent, looping `<audio>` element** while narration is active. Browsers only surface OS media controls while an `HTMLMediaElement` is playing, so this silent element is what makes the notification/lock-screen controls appear and routes Bluetooth/media-key events into our handlers. The clip is a runtime-generated silent WAV `data:` URI — no binary asset shipped.
- **`src/lib/narration/media-session.ts`** (rewritten): now **provider-agnostic** — takes plain `play`/`pause`/`stop`/`previousTrack`/`nextTrack` callbacks instead of a concrete `ArticleNarrator`, and owns the silent-audio lifecycle via `updateMediaSessionPlaybackState` / `clearMediaSession`.
- **`src/components/narration/useMediaSession.ts`** (new hook): keeps metadata, action handlers, and playback state in sync; OS button handlers dispatch through a ref so they always call current callbacks without re-registering.
- **`useNarration`**: calls `useMediaSession` for **both** providers, replacing the old browser-voices-only effect.

Artwork falls back to the PWA manifest icons when no per-article image is supplied. (Wiring per-article `imageUrl` through would require exposing it on the entry read model — left as a follow-up to keep this focused.)

## Testing

- `tests/unit/narration-silent-audio.test.ts` — pure WAV-byte structure/silence generation.
- `tests/unit/frontend/narration-media-session.test.ts` — jsdom test stubbing `navigator.mediaSession` and `HTMLMediaElement.play/pause`; verifies metadata, handler→controls dispatch, playback-state mirroring, and silent-audio start/stop.
- `pnpm typecheck`, `pnpm lint`, and full `pnpm test:unit` (2299 tests) pass.

Note: the silent-audio activation depends on autoplay being permitted, which installed PWAs and post-interaction pages allow; if blocked, it degrades gracefully (no controls, narration unaffected).

### DIFF
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -575,6 +575,9 @@ app/
 - `components/feeds/` - Feed list, add feed dialog
 - `components/narration/` - Audio playback controls
 - `components/saved/` - Saved article views
+
+**Narration OS media controls.** Narration plays through the Web Speech API (browser voices) or the Web Audio API (Piper enhanced voices), neither of which the browser treats as media playback — so setting `navigator.mediaSession` metadata alone surfaces no OS controls (lock screen / notification widget, Bluetooth & media-key buttons), especially in an installed PWA. To make them appear, a **silent looping `<audio>` element** (`src/lib/narration/silent-audio.ts`, a runtime-generated silent WAV data URI) is played while narration is active; that element is what the browser recognizes as media, which surfaces the controls and routes hardware buttons into our action handlers. `src/lib/narration/media-session.ts` is provider-agnostic (plain `play`/`pause`/`stop`/`prev`/`next` callbacks) and the `useMediaSession` hook (`src/components/narration/`) keeps metadata, action handlers, and playback state in sync for **both** providers. Artwork falls back to the PWA manifest icons when no per-article image is supplied.
+
 - `components/settings/` - Settings page components
 - `components/subscribe/` - Subscription flow components
 - `components/summarization/` - AI summary display
--- a/src/components/narration/useMediaSession.ts
+++ b/src/components/narration/useMediaSession.ts
@@ -1,0 +1,97 @@
+/**
+ * useMediaSession Hook
+ *
+ * Wires narration into the OS Media Session so an installed PWA (and any browser)
+ * shows native media controls — the lock-screen/notification widget and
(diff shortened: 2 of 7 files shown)


brendanlong:
Ran an independent code review over the diff and traced the silent-audio lifecycle (error / unmount / article-switch / natural-end / SSR paths — all confirmed leak-free). Two actionable items were found and fixed in the follow-up commit:

1. **Bug:** OS controls disappeared mid-narration when the article's `title`/`feedTitle`/`artwork` changed (e.g. an initial `"Untitled"` resolving to the real title, or an SSE `entry_updated`). The setup effect's cleanup stopped the silent audio and re-setup didn't restart it, while the status effect didn't re-run (status unchanged). `useMediaSession` now re-asserts the current playback state at the end of setup, restarting the silent loop.
2. **Robustness:** on the LLM-normalization path the silent audio was started only after the async generation `await`, i.e. outside the user-gesture activation window, so autoplay policy could reject it. `play()` now primes the silent audio synchronously within the gesture and releases it if generation fails or returns nothing.

Remaining known caveat: activation of the silent audio depends on autoplay being permitted. Installed PWAs (the target of this issue) allow it; in a non-installed browser tab the first LLM-generated narration may not show controls until a cached replay. Worth a quick on-device sanity check once deployed.

— Claude

PR #1245, opened 2026-07-15
Share target: open Subscribe (pre-filled) for shared feed URLs

brendanlong:
## Summary

When you share a **feed** URL to the PWA (or hit the bookmarklet/save flow with one), we now open the **Subscribe** page pre-filled instead of trying to save it as an article.

This isn't just a nicety — the previous behavior was broken. Feeds are served as XML/JSON, which the article fetch (`fetchHtmlPage`) rejects with an "Invalid content type" error, so sharing a feed to the PWA produced an error toast (or, for the rare feed served as `text/html`, saved Readability garbage).

## How it works

1. The `/save` flow calls `saved.save` → `saveArticle()` as before (single fetch on the common article path — no added latency).
2. If the fetch is rejected on its content type, we check whether the URL is **actually a feed**. If so, `saveArticle` throws the machine-readable `URL_IS_FEED` error.
3. The `/save` page catches it and redirects to `/subscribe?url=…`, which pre-fills the field and auto-runs the preview.

Per discussion, detection is deliberately **scoped to direct feeds only**: sharing an article page always **saves the article**, even if the site has a discoverable feed — that matches intent and avoids hijacking the common save case. The extra "is this a feed" sniff fetch only runs on the feed-rejection path, and only for *ambiguous* content types (`application/xml`, `text/xml`, `application/json`); unambiguous `application/rss+xml` / `application/atom+xml` skip even that.

## Changes

- `http/fetch.ts`: promote the content-type rejection to a typed `InvalidContentTypeError` carrying the content type (still an `Error`; existing callers unaffected).
- `trpc/errors.ts`: new `URL_IS_FEED` (400). Message is the machine token, since tRPC doesn't ship the error `cause` to clients — same pattern as the existing `NEEDS_GOOGLE_*` codes.
- `services/saved.ts`: `urlIsDirectFeed()` + detection in `acquireArticleContent`'s fetch catch.
- `app/save/page.tsx`: redirect to Subscribe on `URL_IS_FEED`, showing a spinner ("This is a feed — opening Subscribe…") instead of an error card.
- `components/subscribe/SubscribeContent.tsx`: read `?url=`, pre-fill, auto-preview once.

## Testing

New integration test `tests/integration/saved-feed-detection.test.ts` drives real HTTP against a loopback server:
- feed served as `application/rss+xml` → `URL_IS_FEED` (no second fetch)
- ambiguous `application/xml` whose body sniffs as a feed → `URL_IS_FEED`
- ambiguous `application/xml` sitemap (not a feed) → normal save-fetch error, **not** `URL_IS_FEED`
- normal HTML article → saves as before

`pnpm typecheck` + `eslint` clean; full integration suite (666 tests) passes.

### DIFF
--- a/src/app/save/page.tsx
+++ b/src/app/save/page.tsx
@@ -44,6 +44,15 @@ function SaveContent() {
     onSuccess: (data) => {
       setArticleTitle(data.article.title);
     },
+    onError: (error) => {
+      // The shared URL is actually a feed, not an article — send the user to the
+      // Subscribe page pre-filled with it instead of failing the save. The
+      // server signals this via the machine-readable `URL_IS_FEED` message (tRPC
+      // doesn't ship the error `cause` to the client).
+      if (error.message === "URL_IS_FEED" && urlToSave) {
+        window.location.replace(`/subscribe?url=${encodeURIComponent(urlToSave)}&shared=true`);
+      }
+    },
   });
 
   const uploadFileMutation = trpc.saved.uploadFile.useMutation({
@@ -197,6 +206,10 @@ function SaveContent() {
     requestGoogleDocsAccessMutation.mutate({});
   };
 
+  // The shared URL turned out to be a feed; we're redirecting to /subscribe.
+  // Keep showing a spinner rather than flashing the error card.
+  const isFeedRedirect = activeMutation.error?.message === "URL_IS_FEED";
+
   // Check if error is an auth error (session expired, revoked, etc.)
   const isAuthError = activeMutation.error?.data?.code === "UNAUTHORIZED";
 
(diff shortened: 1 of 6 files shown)


brendanlong:
Ran an Opus code review. Addressed in 2b81eafd:

- **M1 (correctness, fixed):** a plain JSON API served as `application/json` was misrouted to Subscribe because `detectFeedType` returns `"json"` for any `{...}` body. The ambiguous sniff now confirms by actually running `parseFeed()`, which throws for non-feeds (sitemaps, plain JSON objects with no JSON Feed `version`/`items`). Added tests for a valid JSON Feed (routes) vs. a plain JSON API (doesn't).
- **L1 (cleanup):** dropped the unused `shared=true` query param.

Reviewer also verified clean: the auto-preview effect runs exactly once (ref set before the async call), `useSearchParams` is SSR/hydration-safe, the sniff fetch is SSRF-guarded + size-limited, and promoting `fetchHtmlPage`'s error to a typed subclass doesn't affect its other callers (they catch generically).

**Known minor tradeoff (not changed):** `saveArticle` is shared by MCP/Discord/Wallabag, which will surface the raw `URL_IS_FEED` token as the error message rather than a friendly sentence (the token is the signal the `/save` page matches on, mirroring the existing `NEEDS_GOOGLE_*` pattern). Feed-saving via those surfaces is rare, and the token is arguably clearer than the previous "Invalid content type" error — but happy to add per-surface friendly messages if you'd prefer.

— Claude

PR #1246, opened 2026-07-15
docs: runbook for migrating MPG → unmanaged Fly Postgres (shared-cpu-8x)

brendanlong:
## Summary

Production Postgres is moving from **Fly Managed Postgres Basic** (~$41/mo, shared-cpu-2x/1GB — throttled to a 12.5%-of-a-core sustained baseline) to a **single-node unmanaged flex cluster on shared-cpu-8x/4GB** (~$28/mo incl. volume).

Rationale (researched 2026-07-14): Fly shared-CPU quotas are pooled per machine ("Quotas are shared between a Machine's vCPUs" — [CPU performance docs](https://fly.io/docs/machines/cpu-performance/)), so shared-8x gives Postgres a 50%-of-a-core sustained floor, burst to 8 cores on the 500 CPU-second balance, and 4× the RAM for cache — cheaper than both MPG Basic and a dedicated performance-1x/2GB ($32/mo). The first dedicated-CPU MPG tier is $282/mo. Single node is deliberate (flex multi-node repmgr has a poor failure-mode track record); `--enable-backups` WAL archiving to Tigris + daily volume snapshots are the safety net.

## Changes

- **New: `docs/runbooks/migrate-mpg-to-unmanaged-postgres.md`** — step-by-step cutover: create cluster → stop machines → `fly postgres attach` → `pg_dump | psql` over proxies → verify → destroy MPG. Includes rollback, post-cutover checks (incl. restoring the second app machine — currently only one is running despite `min_machines_running = 2`), and the ongoing ops we take on (throttle monitoring via `fly_instance_cpu_balance`, upgrades, restore drill). All flyctl flags verified against the current CLI.
- **`docs/DEPLOYMENT.md`** — provisioning section and cost table updated to the real production shape (they still described the original $7/mo MVP sizing and the old `lion-reader-db` name).

Docs only — no code changes. The actual migration is run manually per the runbook (needs WireGuard-capable credentials).

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -91,31 +91,44 @@ You should see your new app listed.
 
 ## Database Provisioning (Postgres)
 
-Lion Reader uses PostgreSQL for all persistent data.
+Lion Reader uses PostgreSQL for all persistent data. Production runs **unmanaged
+Fly Postgres (flex), single node** — chosen over Fly Managed Postgres in July 2026
+for performance-per-dollar (see `docs/runbooks/migrate-mpg-to-unmanaged-postgres.md`
+for the rationale, the migration procedure, and the operational duties this implies:
+Fly does not support or upgrade unmanaged clusters).
 
 ### 1. Create a Postgres Cluster
 
 ```bash
 flyctl postgres create \
-  --name lion-reader-db \
+  --name lion-reader-pg \
   --region lax \
-  --vm-size shared-cpu-1x \
+  --flex \
+  --vm-size shared-cpu-8x \
+  --vm-memory 4096 \
   --initial-cluster-size 1 \
-  --volume-size 10
+  --volume-size 10 \
+  --enable-backups
 ```
 
 **Options explained:**
 
 - `--name`: Name for your Postgres app (must be unique)
 - `--region`: Should match your app's `primary_region` in `fly.toml`
-- `--vm-size`: `shared-cpu-1x` is sufficient for MVP (~$7/month)
-- `--initial-cluster-size`: 1 for MVP, increase for HA
-- `--volume-size`: 10GB is plenty for MVP
+- `--vm-size`/`--vm-memory`: shared CPU quotas are pooled per machine, so
+  `shared-cpu-8x` gives Postgres a 50%-of-a-core sustained floor with burst to 8
+  cores (~$27/month at 4GB) — better burst behavior than a dedicated
+  `performance-1x` core at ~$32/month. Save the superuser password it prints.
+- `--initial-cluster-size`: 1 (deliberate — flex multi-node uses repmgr, which has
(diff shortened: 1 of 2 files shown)


brendanlong:
The old database is dump. To re-load, I needed to filter out some fly.io extensions: `grep -v -iE 'pg_stat_monitor|pgaudit' dump.sql > dump.clean.sql`. The restore on the new host is in progress.

brendanlong:
The restore finished. Trying to bring the app back up.

brendanlong:
Everything seems to be working, and I see higher CPU util on the new host and ~0 util on the old one.

brendanlong:
The MPG → unmanaged cutover described here is **done** — production now runs on `lion-reader-pg` (unmanaged flex, db `lion_reader`, `shared-cpu-8x`/2GB in `lax`, PG 18.3). So the step-by-step cutover runbook is now historical, and this PR probably shouldn't merge as-is.

I've **salvaged the durable half** into #1374 so it isn't lost when this closes:
- The `docs/DEPLOYMENT.md` corrections (real `lion-reader-pg` shape, cost table, provisioning flags) — updated to the *current* state (ended at 2GB, not the 4GB this PR was written against).
- The "Ongoing operations you now own" duties → folded into a new **"Operating unmanaged Postgres"** section in DEPLOYMENT.md.
- Plus a `migrations/CLAUDE.md` section on the temporarily-scale-for-expensive-migrations pattern.

**Suggest: close this PR** once #1374 merges. The one-time cutover procedure (stop machines → `fly postgres attach` → `pg_dump | psql` → destroy MPG) is the part not worth keeping in the tree — but logging it here in the thread so it's findable if we ever repeat it.

### Operational log from the `0105_entries_search_vector` run (2026-07-18) — recorded here, deliberately not checked in
Ephemeral specifics that would just rot in the repo:
- Ran the heavy migration by temporarily scaling `lion-reader-pg` → `performance-4x` (dedicated CPU; resize is a ~few-second restart), applying `ADD COLUMN … GENERATED STORED` + `CREATE INDEX … USING gin` **by hand**, then deploying so the guarded migration no-op'd column+index and only created the view. `VACUUM ANALYZE entries`, then back to `shared-cpu-8x`.
- **Memory right-sized 8GB → 2GB.** DB is ~4.9GB on disk (grew ~0.9GB from the tsvector column + GIN index). `pg_stat_database` cache-hit was 98.4% and it had run fine on 1GB, so the working set is small — 2GB gives headroom for the GIN index + growth without paying to cache the whole (mostly-cold) DB.
- Machine id at the time: `e820e60b401158`.
- flex re-derives `shared_buffers` from VM memory at each boot (confirmed: shmem 1→2GB when RAM doubled), so RAM scale-down is safe and needs no config edits.
- Open follow-up (there's already an issue): better monitoring now that we own this DB; also a one-time backup restore drill.

— Claude

PR #1251, opened 2026-07-15
Add announcement banner and maintenance mode (admin-controlled)

brendanlong:
Adds two operator tools for downtime/incidents, both driven from **/admin → Status** and backed by **Redis** (DB-independent, so maintenance mode works during a database migration).

## Announcement banner
A closeable bar at the top of **every** page (info/warning). Rendered SSR from the root layout via `getAnnouncement()`, so it shows on the app, the demo, and logged-out pages. Dismissal is stored in localStorage keyed by a **message-derived id**, so changing the announcement re-shows it to everyone who dismissed the old one.

Announcement banner on the demo

## Maintenance mode
A global kill switch for use during a heavier/non-backward-compatible DB migration. The **custom HTTP server** (`scripts/server.ts`) is the single choke point every request passes through, so the gate lives there (`src/server/maintenance/server-gate.ts`). When on, everything returns a self-contained **503** (HTML page for pages, JSON for API) **except**:

- `/demo` (never touches Postgres — stays fully usable)
- `/admin` + `/api/admin` (so you can turn it back off)
- `/api/health`, legal pages, `/`, and static assets
- `/api/trpc` **only** when carrying a valid `admin_session` cookie

The **worker** stops claiming jobs (no Postgres access; loop just sleeps) and the **Discord bot** stops responding while it's on — all three process groups read the same Redis flag and resume automatically when it clears. A `MAINTENANCE_MODE` env var is an additional force-on fallback.

Maintenance page

## Admin UI
New **Status** tab with a live-preview announcement editor and the maintenance toggle.

Admin Status tab

## Design notes
- **Redis, not Postgres**: maintenance mode is meant to run *while Postgres is being migrated/locked*, so the code that reads the flag must not depend on the DB.
- **Fail-safe**: Redis down ⇒ no banner and maintenance = the env override only; never accidentally takes the site down or shows a stale banner.
- **No per-request Redis**: the server gate reads a background-polled module variable; reads elsewhere are cached in-process for a few seconds.
- No schema migration required.

## Testing
- **Unit** (`tests/unit/maintenance-gate.test.ts`): pure `evaluateRequest` truth table (demo/admin/health/legal/static/`/` allow; `/api/trpc` with/without a valid admin cookie; other `/api/*` → block-api; pages → block-page), HTML escaping, JSON body.
- **Integration** (`tests/integration/site-status.test.ts`, real Redis): flag round-trips, `MAINTENANCE_MODE` env fallback, fail-safe defaults, and the message-specific announcement id.
- **Manual** (Playwright + curl against a local dev server): verified the banner renders on demo/login, maintenance 503s the app/API while exempting demo/admin/health/legal + admin-cookie tRPC, and the worker logs paused→resumed. `pnpm typecheck`, `pnpm test:unit`, `pnpm lint`, `check:colors`/`check:theme-tokens`, and all three esbuild bundles pass.

### DIFF
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -420,6 +420,10 @@ Practically, this means using the expand/contract pattern:
 - **Expand** (safe in one release): add nullable columns or columns with defaults, add tables, add indexes, create views alongside old ones
 - **Contract** (requires two releases): to drop or rename a column/table, first ship a release whose code no longer references it; only then ship the migration that removes it. A rename is an add + dual-write/backfill + drop across releases, never a single `ALTER ... RENAME`.
 
+### Maintenance Mode
+
+For a heavier migration that can't be made backward-compatible (a data backfill, a non-expand/contract change), an admin can flip **maintenance mode** from `/admin` → Status. It is a **Redis** flag (not Postgres — the DB may be the thing being migrated) read by every process group. The custom server (`scripts/server.ts`) then serves a self-contained 503 maintenance page/JSON for all traffic except the demo (no DB), the admin panel, and the health check; the worker stops claiming jobs and the Discord bot stops responding. Turning it off resumes everything with no redeploy. A `MAINTENANCE_MODE` env var is an additional force-on fallback. There's also a Redis-backed, closeable **announcement banner** (info/warning) shown site-wide for known-issue notices. See "Site Status" in `src/server/CLAUDE.md`.
+
 ### Local Development
 
 Docker Compose provides Postgres and Redis for local development. See README for setup instructions (and "Local Services" in the root CLAUDE.md for the no-Docker path).
--- a/scripts/raw-color-baseline.json
+++ b/scripts/raw-color-baseline.json
@@ -22,6 +22,7 @@
   "src/components/admin/AdminApp.tsx\thover:border-zinc-300",
   "src/components/admin/AdminFeedsContent.tsx\tdark:bg-zinc-800",
   "src/components/admin/AdminInvitesContent.tsx\tdark:bg-zinc-800/60",
+  "src/components/admin/AdminStatusContent.tsx\tdark:bg-zinc-800",
   "src/components/auth/OAuthSignInButton.tsx\tdark:hover:bg-zinc-200",
   "src/components/auth/OAuthSignInButton.tsx\thover:bg-zinc-800",
   "src/components/entries/EntryArticle.tsx\tactive:text-zinc-800",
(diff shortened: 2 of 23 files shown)


brendanlong:
Ran an automated Opus code review on the branch. It caught one real ship-blocker, now fixed in `8969239c`:

- **Maintenance gate let static-suffixed API routes through.** `evaluateRequest` checked the `STATIC_EXT` (`.json`/`.xml`/`.txt`/…) allow-list *before* the API decision, so authenticated, Postgres-touching routes like the whole Wallabag compat API (`/api/wallabag/api/entries.json`, `…/tags.json`, `/api/entries.json`, etc.) were treated as static assets and stayed up during maintenance — defeating the purpose during a DB migration. Fixed by evaluating the API/`/api/trpc` decisions first and scoping the static-asset check to non-`/api/` paths. Added regression tests for `.json`-suffixed API paths.

Also addressed a minor note from the review: the gate now accepts a `Bearer ALLOWLIST_SECRET` in addition to the `admin_session` cookie (mirroring `adminMiddleware`), so a programmatic admin isn't locked out of `/api/trpc` while maintenance is on.

The review confirmed the rest — worker/bot pauses (all DB paths covered, incl. the raw-gateway DM recovery), SSR/Redis fail-safety, the admin-cookie check, and the UI state handling — as correct.

— Claude

brendanlong:
Added live push for the announcement banner so open tabs update without a reload — via a new SSE event, not polling (zero extra traffic when nothing changes).

- New global `site-status:events` Redis channel (broadcast, not per-user). `admin.setAnnouncement` publishes `announcement_changed` with the resolved announcement (or null when cleared).
- The SSE endpoint subscribes every connection to that one channel (ref-counted on the shared subscriber → a single Redis SUBSCRIBE per process regardless of connection count) and forwards the event.
- Client: `announcement_changed` flows through the existing `parseSyncEvent`/`handleSyncEvent` pipeline into a small `useSyncExternalStore` module store (`@/lib/site-status/announcement-store`). The root-layout banner reads the live override and falls back to its SSR value until an event arrives (`undefined` = no update yet, `null` = cleared, so a live "clear" actually hides it rather than reverting to the stale SSR value).

Maintenance mode is intentionally **not** pushed — the SSE endpoint itself is blocked by the gate once maintenance is on.

New e2e test publishes a real Redis `announcement_changed` and asserts the banner shows then clears with zero `entries.*`/`tags.*`/`subscriptions.*` refetches; all existing realtime e2e specs still pass.

— Claude

PR #1252, opened 2026-07-15
Quiet the Docker build: suppress benign otel warnings, bake git SHA

brendanlong:
## What

The clean `docker build` emitted "hundreds of lines" of import/build warnings. The build was already **succeeding** — these were all warnings, not fatal errors — but two sources were worth cleaning up:

### 1. Sentry → OpenTelemetry "Critical dependency" warnings (the bulk)
`@sentry/node` pulls in OpenTelemetry auto-instrumentation, which hooks module loads via dynamic `require()` (`require-in-the-middle`, `@prisma/instrumentation`). webpack can't statically resolve those, so it emits a multi-line `Critical dependency: the request of a dependency is an expression` warning **with a full import trace for each otel instrumentation version** — dozens of lines of noise for third-party code we can't change.

Fixed with a **scoped** `webpack.ignoreWarnings` in `next.config.ts` matched to those specific packages, so genuine "Critical dependency" warnings from our own code still surface.

### 2. `git: not found` (a real functional gap, not just noise)
The alpine builder has no `git`, so `next.config.ts`'s `git rev-parse` fallback fails and `GIT_COMMIT_SHA` was baked into prod as **`undefined`** — and it's surfaced in the outgoing `User-Agent`. Now passed explicitly as a Docker build arg from the deploy workflow (`flyctl deploy --build-arg GIT_COMMIT_SHA=$(git rev-parse --short HEAD)`), fixing both the noise and the missing value.

### Left as-is (harmless, single lines)
- esbuild bundle-size `⚠️` markers (server/worker/discord bundles)
- Workbox `26 MB wasm won't be precached` (intended — we don't want to precache the 26 MB ONNX wasm)

## Verification
- `pnpm typecheck` passes
- Local `pnpm build` completes with **zero** `Critical dependency` lines (previously ~6 multi-line blocks); only the intentional WASM-precache warning remains

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -47,6 +47,9 @@ jobs:
         uses: superfly/flyctl-actions/setup-flyctl@master
 
       - name: Deploy to Fly.io
-        run: flyctl deploy --remote-only
+        # Pass the deployed commit SHA as a build arg: the alpine builder has no
+        # git, so next.config.ts's `git rev-parse` fallback can't resolve it.
+        # Short SHA to match the previous git-derived value baked into User-Agent.
+        run: flyctl deploy --remote-only --build-arg GIT_COMMIT_SHA="$(git rev-parse --short HEAD)"
         env:
           FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,14 @@ RUN node scripts/copy-onnx-wasm.mjs
 # Set environment for build
 ENV NEXT_TELEMETRY_DISABLED=1
 ENV NODE_ENV=production
+
+# Git commit SHA for the build. The alpine builder has no git, so next.config.ts's
+# `git rev-parse` fallback fails ("git: not found") and the SHA is baked in as
+# undefined — it's surfaced in the outgoing User-Agent. Pass it explicitly from CI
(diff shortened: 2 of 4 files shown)


brendanlong:
Added a third fix to this PR: **the ~thousands of `[ioredis] Unhandled error event: AggregateError` lines** in the build.

I initially mis-diagnosed these as a local-sandbox-only artifact — I'd read a truncated CI log (the job was still in progress). The completed log has **630 of these blocks (~1,900 lines)** during `next build`, all starting at "Generating static pages".

**Root cause:** the root layout (`src/app/layout.tsx`) calls `getAnnouncement()` during static prerender → `getSiteStatus()` → `getRedisClient()`, which eagerly connects (`lazyConnect: false`) to the dummy `REDIS_URL` the Dockerfile sets. Nothing listens there, so the connect fails on both `::1` and `127.0.0.1` (AggregateError), and since the `error` listener is gated behind `NODE_ENV === "development"` (build runs as production), ioredis prints its raw fallback per retry across the render workers.

**Fix** (`src/server/redis/index.ts`): return `null` from `getRedisClient()` when `NEXT_PHASE === "phase-production-build"`, matching the Dockerfile's "modules check these exist but don't connect" intent. Every consumer already treats a null client as "Redis unavailable" and falls back safely (no announcement). `NEXT_PHASE` is unset at runtime, so the server/worker connect normally.

Verified with a full local production build: **0 ioredis lines (was 630)**, static generation completes cleanly, and `NEXT_PHASE` does reach the render workers.

— Claude

PR #1258, opened 2026-07-15
Narration: live pause/skip during chunk generation + cap chunk length

brendanlong:
## Summary

Two narration playback improvements requested from the [fork-around-and-find-out-part-2](https://www.lesswrong.com/posts/6reCnPYeopThEFQxN/fork-around-and-find-out-part-2-one-head-does-the-summing) post:

### 1. Skip forward/backward and pause while a chunk is generating

Previously play/pause and skip were disabled whenever a chunk was being synthesized (`showLoadingState = isLoading || isBufferingMidPlayback`), so mid-playback buffering locked the whole control bar.

Now only the **initial** pre-playback generation locks the controls. Once playback has started, pause and skip stay live while the next chunk generates:

- `NarrationControls` splits the old `showLoadingState` into `isInitialLoading` (locks controls, nothing to control yet) vs `isBufferingMidPlayback` (playback active, keep controls live — spinner still shows as an activity indicator but the button pauses).
- `StreamingAudioPlayer.pause()` now handles the `buffering` state: it cancels the in-flight generation and records `pausedWhileBuffering` so resume restarts from the current position (there's no audio source to resume mid-generation — it replays the cached chunk if generation finished while paused, else regenerates).
- Keyboard shortcuts (`p`, `Shift+N`, `Shift+P`) get the same treatment.

Browser voices are unaffected — `ArticleNarrator` never enters a `loading` status mid-playback, so `isBufferingMidPlayback` is Piper-only.

### 2. Cap synthesis chunk length

Neural TTS rushes/muddles very long sentences. `splitIntoSentences` now breaks any sentence over `MAX_CHUNK_CHARS` (180) at clause boundaries (commas/semicolons/colons/dashes), falling back to word boundaries only when a lone clause is still too long, and greedily recombines short clauses so it doesn't over-split. Sentences within the limit are untouched.

The reported 395-char single-sentence paragraph now reads as four clause-sized chunks (~90–106 chars each) instead of one rushed blob, while the shorter neighboring paragraphs (e.g. 133 chars) stay whole.

## Testing

- Added unit tests in `tests/unit/frontend/sentence-splitter.test.ts` covering the long-sentence chunking (clause split, word-boundary fallback, word-preservation, and the real 395-char example).
- `pnpm typecheck`, `pnpm test:unit`, and eslint pass.

### DIFF
--- a/src/components/narration/NarrationControls.tsx
+++ b/src/components/narration/NarrationControls.tsx
@@ -115,16 +115,20 @@ export function NarrationControlsImpl({
   const { status, currentParagraph, totalParagraphs } = state;
   const isPlaying = status === "playing";
   const isPaused = status === "paused";
-  // Consider "loading" as active when we already have paragraphs (buffering mid-playback)
-  // vs initial generation (no paragraphs yet)
+  // "loading" once we already have paragraphs means we're generating the next
+  // chunk mid-playback. The controls stay live here so the user can pause or
+  // skip while a chunk generates. Before any paragraphs exist we're still doing
+  // the initial narration generation, where there's nothing yet to control.
   const isBufferingMidPlayback = status === "loading" && totalParagraphs > 0;
+  const isInitialLoading = (isLoading || status === "loading") && !isBufferingMidPlayback;
   const isActive = isPlaying || isPaused || isBufferingMidPlayback;
 
   /**
-   * Handle play/pause button click.
+   * Handle play/pause button click. Pausing works while a chunk is generating
+   * (buffering) as well as during normal playback.
    */
   const handlePlayPause = () => {
-    if (isPlaying) {
+    if (isPlaying || isBufferingMidPlayback) {
       pause();
     } else {
       play();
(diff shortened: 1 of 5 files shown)


brendanlong:
## Follow-up: fixed a generation race surfaced by review

An automated review flagged a real correctness bug that the new "control-while-generating" behavior makes easy to hit, plus a related pre-existing rough edge. Both are now fixed (2nd commit):

1. **Generation race on fast pause→resume / skip mid-chunk.** The player used a single shared `bufferingCancelled` boolean checked *after* the `await generateAudio`. When a newly-started generation reset that flag to `false`, an already-cancelled in-flight generation would be resurrected — caching and playing its chunk anyway. Result: the sentence audibly restarts, plus double CPU and a duplicate background buffer loop after a skip. Replaced the post-await check with a monotonic `bufferGeneration` id that `cancelBuffering()` bumps; each generation discards its result if the id changed while it was in flight. (Concurrent *non-cancelled* generations capture the same id, so ordinary buffer-ahead isn't disturbed — no stalls.)

2. **Skip-while-paused then resume could play nothing** (pre-existing on `master`, since skip was already enabled while paused). Skipping stops/discards the paused audio buffer, but the resume path still tried to resume that (now-cleared) source and did nothing. Generalized the flag (`pausedWhileBuffering` → `resumeByRestart`) so both "paused mid-generation" and "skipped while paused" resume by restarting from the current position.

`pnpm typecheck`, full `pnpm test:unit` (2408 passing), and eslint remain green.

— Claude

PR #1274, opened 2026-07-16
Security: bind OAuth state to the initiating browser (#1263)

### LINKED ISSUE #1263: OAuth login CSRF: bind OAuth `state` to the initiating browser
From the security review of the auth surfaces (companion to #1262).

## Problem
The OAuth `state` (and Google PKCE verifier) is stored only in Redis keyed by the state value; callback validation just checks "does this state exist in Redis." It is never tied to the initiating browser (no state cookie / session binding). See `src/server/auth/oauth/google.ts`, `apple.ts`, `discord.ts`.

An attacker can start their own OAuth flow, obtain a valid `code`+`state` for the attacker's provider account, and deliver the callback URL to a victim. The victim's browser then completes login into the **attacker's** account (login CSRF / session fixation — a data-capture vector). `SameSite=Lax` + one-time state + 10-min TTL reduce but do not close this (the callback is a top-level GET, permitted by Lax).

## Fix
Set a short-lived `HttpOnly; SameSite=Lax` state cookie when generating the auth URL and require it to match the callback `state` (the arctic/Lucia double-submit pattern).

Severity: Medium.

— Claude

brendanlong:
Fixes #1263.

## Problem

The OAuth `state` (and Google PKCE verifier) was stored only in Redis keyed by the state value, so callback validation only proved *"we issued this state"* — never that it belonged to the browser that started the flow. That left a **login-CSRF / session-fixation** hole: an attacker runs their own OAuth flow, captures a valid `code`+`state` for their account, and delivers the callback URL to a victim, whose browser then logs into the **attacker's** account (a data-capture vector). `SameSite=Lax` + one-time state + the 10-min TTL narrow but don't close it — the callback is a top-level GET, which Lax permits.

## Fix

The arctic/Lucia **double-submit** pattern:

- Generating an authorization URL now sets a short-lived `HttpOnly` `oauth_state` cookie holding the same `state`, on the browser tRPC path via `resHeaders` (no-op for REST/OpenAPI callers, exactly like the session cookie). Applies to `googleAuthUrl` / `appleAuthUrl` / `discordAuthUrl` and `requestGoogleDocsAccess`.
- The three browser-facing callback routes (`/api/v1/auth/oauth/{google,apple,discord}/callback`) require `oauthStateCookieMatches(cookie, state)` — **fail-closed on a missing cookie** — before touching Redis, and clear the one-time cookie on every response.
- The binding (an attacker can't set/read the victim's `HttpOnly` cookie), **not** `SameSite`, is the protection.

### Apple needs `SameSite=None`

Apple uses `response_mode=form_post` — a cross-site **POST** from `appleid.apple.com`, from which `SameSite=Lax` cookies are withheld. A Lax cookie would never arrive and would break every Apple login, so the Apple state cookie is `SameSite=None; Secure` (still `HttpOnly`, still bound the same way). Google/Discord redirect back with a top-level GET and stay `SameSite=Lax`.

### Not cookie-guarded (by design)

The tRPC callback mutations (`googleCallback` / `appleCallback` / `linkGoogle` / `linkApple`) are JSON-only tRPC (CSRF-resistant) and double as non-browser REST endpoints returning the token in the body; the client `/auth/oauth/callback` page they back is not a provider redirect target (all `redirect_uri`s point at the REST routes).

## Tests

- New `tests/unit/oauth-state-cookie.test.ts` covers the pure match logic (accept / mismatch / **missing-cookie fail-closed** / empty) and the cookie attributes (Lax vs. None; Secure).
- Full unit suite green; OAuth integration suites (`google-oauth`, `apple-oauth`) green — the service-layer functions are unchanged, so binding lives entirely at the HTTP boundary.
- `pnpm typecheck` + `pnpm lint` clean.

Docs updated: `SECURITY.md` §4 and `src/server/auth/CLAUDE.md`.


brendanlong:
Ran an independent Opus review after opening this PR. It confirmed the core CSRF fix is sound (cookie reliably set before navigation, Apple `SameSite=None; Secure` correct, fail-closed on missing cookie, no double-Set-Cookie conflicts) and caught one regression I've now fixed:

- **Google `extension-save` flow would break with `invalid_state`.** That flow builds its auth URL in a Server Component (`src/app/extension/save/page.tsx`), and Next.js forbids setting cookies during a Server Component render — so no `oauth_state` cookie is ever set, and an unconditional check rejected the callback.

Fix (98a13a3): the Google callback now enforces the cookie for every mode **except** `extension-save`. That mode re-authorizes an already-logged-in user's own Google scopes (identified by their session), so it isn't the login-CSRF vector the issue is about. The login/save modes (tRPC-generated) still set and verify the cookie. Documented the exemption in `SECURITY.md` and `src/server/auth/CLAUDE.md`, with a note to drop it if extension-save is ever moved to a Route Handler.

— Claude

Issue #1264, opened 2026-07-16
Add security response headers (CSP, frame-ancestors, HSTS, nosniff, Referrer-Policy)


brendanlong:
From the security review (companion to #1262). Defense-in-depth.

## Problem
`next.config.ts` `headers()` and the custom server (`scripts/server.ts`) set only caching/maintenance headers — no CSP, `X-Frame-Options`/`frame-ancestors`, HSTS, `X-Content-Type-Options`, or `Referrer-Policy`. Entry bodies render via `dangerouslySetInnerHTML`, so the server-side sanitizer is the **sole** XSS gate with no CSP backstop, and nothing prevents the app from being framed (clickjacking).

## Fix
Add response headers (in `next.config.ts headers()` or the custom server):
- CSP — start with `frame-ancestors 'none'`, `object-src 'none'`, `base-uri 'self'`; a `script-src` lockdown needs a nonce/hash for the two inline scripts in `layout.tsx`.
- `X-Content-Type-Options: nosniff`
- `Referrer-Policy: strict-origin-when-cross-origin`
- HSTS in production.

This turns any future single-point sanitizer regression into a non-event.

Severity: Low (defense-in-depth).

— Claude

PR #1278, opened 2026-07-16
Security: lock down CSP script-src/default-src with a per-request nonce (#1275)

### LINKED ISSUE #1275: Lock down CSP script-src/default-src with a nonce for the inline layout scripts
Follow-up to #1264 / #1269. Those added a baseline CSP (`frame-ancestors 'none'; object-src 'none'; base-uri 'self'`) but **intentionally left `script-src`/`default-src` unset**, so the CSP does not yet restrict script execution. The server-side HTML sanitizer therefore remains the **sole** gate against injected `<script>` — a single sanitizer regression is still directly exploitable as XSS. This issue tracks closing that gap.

## Why it was deferred

Setting `script-src` would block the two inline `<script>`s in `src/app/layout.tsx`:
- `textAppearanceScript` — blocking `<head>` script that applies text size/font before first paint (avoids FOUC)
- `swScript` — service-worker registration

A static `next.config.ts` `headers()` value is the same string on every response, so it can't carry a per-request nonce. Doing this right needs the nonce generated per-request and threaded to both the CSP header and the inline `<script nonce>` attributes.

## Proposed approach

1. Generate a random nonce per request in `middleware.ts` (Next's documented CSP-nonce pattern: create the nonce, set it on an outgoing request header, and set the `Content-Security-Policy` response header there).
2. Read the nonce in `app/layout.tsx` via `headers()` and pass it to both `<script nonce={nonce}>` tags.
3. Tighten the CSP to something like:
   - `script-src 'self' 'nonce-<value>' 'strict-dynamic'`
   - `default-src 'self'`
   - keep the existing `object-src 'none'`, `base-uri 'self'`, `frame-ancestors 'none'`
   - add whatever `connect-src`/`style-src`/`font-src`/`img-src`/`worker-src` the app actually needs (Sentry `/monitoring` tunnel, SSE, tRPC, Google Fonts `fonts.googleapis.com`/`fonts.gstatic.com`, `data:`/`https:` images, the service worker, next-pwa).

## Watch out for

- **Static generation vs. per-request nonce** — a nonce forces dynamic rendering. Confirm this is acceptable for the routes involved (or scope the nonce'd CSP to the dynamic app shell).
- **next-pwa / precached HTML** — a nonce baked into a cached document goes stale; make sure the CSP interacts correctly with the service worker and precache.
- **The custom server + maintenance short-circuit** (`scripts/server.ts`) sets its own static CSP on 503s; that path has no inline scripts, so it can keep a script-less policy — just keep the two in sync conceptually.
- **Report-only rollout** — consider shipping `Content-Security-Policy-Report-Only` with the tightened `script-src` first to catch anything the policy would break before enforcing.
- Update `SECURITY.md` (the note added in #1269 explicitly says `script-src` is not yet covered) once this lands.

## Acceptance

- Injected/inline `<script>` without the current nonce does not execute (verify in a browser: a CSP violation is logged, the script is blocked).
- The app's own inline scripts, Sentry, SSE/tRPC, fonts, and the service worker all still work — zero CSP violations in normal use.

Severity: Low→Medium (this is the meaningful half of the XSS backstop that #1269 could not deliver).

— Claude

brendanlong:
Closes #1275.

Follow-up to #1264 / #1269: those shipped a baseline CSP that intentionally left `script-src`/`default-src` unset. This PR closes that gap, so an injected `<script>` or inline event handler that survives a sanitizer regression is blocked by the browser instead of executing.

## What changed

- **`src/proxy.ts`** now generates a random nonce per request and sets the full CSP on both the forwarded request headers (Next.js extracts the nonce from there and stamps it onto its own framework/chunk scripts) and the response — Next's documented CSP pattern. The static CSP entry is removed from `next.config.ts` (a static `headers()` value can't carry a per-request nonce).
- **`src/server/http/csp.ts`** (new) builds the policy, with rationale for every directive:
  - `script-src 'self' 'nonce-…' 'strict-dynamic' 'wasm-unsafe-eval'` (+ `'unsafe-eval'` in dev for HMR). `wasm-unsafe-eval` is for the TTS narration WebAssembly (ONNX runtime + Piper phonemizer).
  - `default-src 'self'`; `img-src`/`media-src` allow `http:/https:/data:/blob:` (sanitized entry HTML legitimately embeds arbitrary remote images/audio/video); `style-src 'self' 'unsafe-inline'` (React style props, next-themes, sanitized `style` attrs).
  - `connect-src` allows self (tRPC/SSE/Sentry tunnel) plus the TTS downloads: Hugging Face voice models (incl. its LFS CDNs — CSP checks every redirect hop) and the Piper wasm/data on jsdelivr.
  - `frame-src` is derived from the sanitizer's `EMBED_CANONICAL_HOSTNAMES`, so the CSP double-enforces the embed-provider allow-list.
  - `worker-src 'self' blob:` (service worker + ONNX thread workers).
  - `form-action` deliberately omitted: Chrome checks post-submit redirects against it, which would break the OAuth consent flow's 302 to external `redirect_uri`s.
- **`layout.tsx`** reads the nonce from `x-nonce` and puts it on the two inline scripts (with `suppressHydrationWarning` — browsers blank the nonce attribute after parsing, which otherwise trips React hydration warnings); **`ThemeProvider`** gains a `nonce` prop for next-themes' inline theme script.
- **`/sw.js` is deliberately CSP-exempt**: a service worker's fetches are governed by the CSP on the worker script itself, and the runtime image caching fetches cross-origin entry images from *inside* the worker — the app's `connect-src` would silently break it.
- The maintenance short-circuit in `scripts/server.ts` keeps its own static CSP, tightened with `script-src 'none'` (that HTML has no scripts).
- SECURITY.md and `src/server/http/CLAUDE.md` updated.

## Verification

- New e2e spec (`tests/e2e/security-headers.spec.ts`), run against **both dev and the production build**: policy shape + nonce matches the inline scripts, fresh nonce per request, injected `` markup is **blocked** (`script-src-attr` violation fires, handler never runs), and a logged-in `/all` page load produces **zero CSP violations**.
- Full e2e suite passes (50 passed); typecheck/lint/unit green; production build succeeds (all HTML routes were already dynamic, so the nonce forces no rendering change).
- Manually verified in a browser against the prod build: demo page renders entries with zero violations and the service worker registers successfully.

## Notes / residual risk

- **Zod 4 probes eval support** with `new Function("")` in a try/catch and cleanly falls back to its non-JIT validation path when blocked; the caught probe still emits a one-time `securitypolicyviolation` event (filtered as known-benign in the e2e test). No functional impact.
- **TTS narration (Piper voice download + synthesis) couldn't be exercised headless** — the directives were sized from reading the actual load paths (`piper-tts-provider.ts`, `voice-download.ts`, the bundled piper-tts-web/ONNX code), but it's worth a quick manual narration run after deploy. If anything's blocked it'll show as a console CSP violation naming the host to add.
- I went straight to enforcing rather than a `Report-Only` rollout since there's no report collector wired up and the e2e suite + manual checks cover the main surfaces; happy to flip to Report-Only if you'd rather stage it.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -44,14 +44,22 @@ Entry bodies, saved articles, and AI summaries are rendered with
 - The allowlist deliberately excludes `script`/`style`/`on*`/`form`/`base`/`meta`,
   restricts URL schemes, forces `rel=noopener`, and treats SVG/MathML as mXSS-prone.
   Changes here need a security review.
-- **Defense-in-depth headers** (`securityHeaders` in `next.config.ts`, mirrored on
-  the maintenance short-circuit in `scripts/server.ts`) add a CSP
-  (`frame-ancestors 'none'; object-src 'none'; base-uri 'self'`), `X-Frame-Options`,
-  `X-Content-Type-Options: nosniff`, `Referrer-Policy`, and (prod) HSTS. The CSP
-  does **not** yet lock down `script-src`/`default-src` — that needs a per-request
-  nonce/hash for the two inline `<script>`s in `layout.tsx` — so the sanitizer is
-  still the sole gate against injected `<script>`; the CSP only backstops
-  clickjacking, plugin/object injection, and `<base>` hijacking.
+- **Defense-in-depth headers**: `securityHeaders` in `next.config.ts` sets
+  `X-Frame-Options`, `X-Content-Type-Options: nosniff`, `Referrer-Policy`, and
+  (prod) HSTS. The **Content-Security-Policy** is set per-request in
+  `src/proxy.ts` with the policy built in `src/server/http/csp.ts`: it locks
+  `script-src` down to a random per-request nonce (`'strict-dynamic'`), sets
+  `default-src 'self'`, restricts `frame-src` to the sanitizer's allow-listed
+  embed hosts, and keeps `frame-ancestors 'none'` / `object-src 'none'` /
+  `base-uri 'self'`. The nonce is threaded to the inline `<script>`s in
+  `layout.tsx` (and next-themes) via the `x-nonce` request header, so an
+  injected `<script>` (or inline event handler) that survives a sanitizer
+  regression is blocked by the browser instead of executing — the sanitizer is
+  the primary XSS gate, the CSP is the backstop. Directive rationale lives in
+  `csp.ts`; the maintenance short-circuit in `scripts/server.ts` carries its own
+  static, script-less CSP (it bypasses Next and has no scripts). `/sw.js` is
+  deliberately CSP-exempt (the service worker's own fetches for runtime caching
+  of cross-origin images would be broken by the app's `connect-src`).
 
 ## 2. SSRF-safe outbound fetching
 
--- a/next.config.ts
+++ b/next.config.ts
@@ -77,20 +77,15 @@ const withPWAConfig = withPWA({
 
 // Security response headers applied to every route (defense-in-depth). Entry
 // bodies render via `dangerouslySetInnerHTML`, so the server-side sanitizer is
-// the sole XSS gate; these headers make a future sanitizer regression a
-// non-event (CSP blocks plugins/base-tag injection) and stop the app from being
-// framed (clickjacking).
+// the primary XSS gate; these headers make a future sanitizer regression a
+// non-event and stop the app from being framed (clickjacking).
 //
-// The CSP intentionally omits `default-src`/`script-src`: locking those down
-// requires a per-request nonce/hash for the two inline <script>s in layout.tsx,
-// which the static `headers()` config can't emit. We still ship the directives
-// that are safe without a nonce. `frame-ancestors` has no X-Frame-Options
-// equivalent for allow-lists, but we send both so older browsers are covered.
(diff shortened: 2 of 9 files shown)


brendanlong:
Ran an Opus-level reviewer over the diff. No correctness or security bugs found; it confirmed the maintenance short-circuit, /register rewrite, /sw.js exemption, header-spoofing protection, and nonce coverage of all inline scripts.

Its one substantive concern — that `connect-src`'s `*.hf.co` might not cover HuggingFace's Xet CDN hosts (`cas-bridge.xethub.hf.co`) — turned out to be a false alarm: CSP host wildcards suffix-match subdomains at **any** depth (unlike TLS certs), and I verified empirically that the voice-model `resolve/` URL 302s to `cas-bridge.xethub.hf.co`, which `https://*.hf.co` matches. Pushed a follow-up commit (3778e94d) documenting that, the `form-action`-omission fallback semantics (no `default-src` fallback; safe because the sanitizer strips `<form>`), a corrected base64-grammar note, and the e2e coverage gaps (Sentry Replay and the real TTS downloads aren't exercised in the test env).

— Claude

PR #1281, opened 2026-07-16
Rewrite the entry-HTML sanitizer in Rust (one N-API call, SANITIZER_VERSION 9)

brendanlong:
## Summary

Replaces the JS sanitization pipeline (sanitize-html + htmlparser2 transforms) with a native Rust module, `@lion-reader/sanitizer` (`native/sanitizer/`). One call into native code now runs the whole pipeline: MathJax CHTML→MathML conversion → inline-SVG sanitization → the HTML allow-list pass (lol_html) → SVG reinsertion. This is phase 1 of the plan to eventually drop the persisted sanitized-content columns (phase 2 issue to follow).

**Performance** (benchmarked on brendan-desktop): the full pipeline runs ~9× faster than sanitize-html — 53ms → ~6ms on a 700KB body, ~0.1ms on a typical 10KB entry. The async form runs on the libuv thread pool, so sanitization no longer needs the piscina worker pool at all (the pool remains for Readability cleaning and feed parsing).

## Security notes (reviewers: read SECURITY.md §1 + src/server/html/CLAUDE.md first)

- **Rules are direct ports** of `SANITIZE_OPTIONS`, `mathjax-chtml.ts`, `sanitize-svg.ts` (DOMPurify-derived SVG profile), and `embed-providers.ts`. All XSS-vector tests were preserved and pass, plus new Rust unit tests (`cargo test -p lion-reader-sanitizer-core`, 50 tests) covering entity-obfuscated schemes (`&colon;`, numeric entities, tab injection), foreign-content edge cases, and unclosed-element recovery.
- **lol_html tokenizes per the HTML5 spec** — the same tokenization the browser applies to the output — which removes the parser-differential (mXSS) exposure the non-spec htmlparser2 parser had. Untouched markup passes through byte-identical (case, entities, quoting are preserved rather than re-serialized).
- **URL attributes are checked entity-decoded** (the single decode a browser does), fail-closed allow-list.
- **SVG**: html5ever's spec foreign-content parsing gives browser-faithful camelCase adjustment (`viewbox` → `viewBox`), replacing the XML-mode workaround. Same constrained allow-list (no script/foreignObject/use/animation/style; per-element href scheme rules).
- **The worker-bundle staleness hazard is structurally gone**: `SANITIZER_VERSION` (now **9**) is compiled into the .node and re-exported to TS, and the main process + worker threads load the same binary — they cannot disagree. The version probe machinery remains as a load check (candidate for phase-2 removal).
- **v8→v9 has no fast-forward**: serialization changed for every row, so the whole table re-sanitizes via the read-path heal / bulk script.

## Build & deploy

- `native/sanitizer` joins the pnpm workspace; `pnpm build:native` is a plain `cargo build` (no @napi-rs/cli) + artifact copy. Committed `Cargo.lock`; `.node` binary is gitignored.
- CI: unit/integration/e2e jobs build the native module (ubuntu runners have Rust).
- Docker: a rust/cargo install + `pnpm build:native` in the alpine builder stage (with cargo cache mounts); the runner copies the loader + binary (musl build happens in-image).
- The loader resolves the binary bundler-proof (turbopack inlines workspace packages even when marked external), mirroring `resolveWorkerPath`.
- Removed deps: `sanitize-html`, `@types/sanitize-html`, `dom-serializer`, `domhandler`, `domutils`.

## Intentional output differences (why tests changed)

- Kept markup passes through byte-identical instead of being re-serialized: uppercase tags stay uppercase, entities stay encoded, attribute quoting is preserved. Semantically identical in the browser.
- SVG subtrees re-serialize with attributes in sorted order.
- MathML output matches the old dom-serializer conventions (verified against the real mathjax-full 3.2.2 fixtures).

## Testing

- `cargo test`: 50 passed
- `pnpm test:unit`: 2412 passed
- `pnpm test:integration:local`: 688 passed
- `pnpm test:e2e:local`: 46 passed
- `pnpm typecheck` / `lint` / `knip` clean (knip baseline unchanged)
- Production bundles build; `dist/worker-thread.js` loads the native module and reports v9


brendanlong:
Phase 2 (drop the persisted sanitized-content columns and sanitize per read) is tracked in #1282.

— Claude

brendanlong:
Ran an Opus-level review plus my own adversarial probing. Two reviewers found the port faithful with no allow-list regressions, but my post-review fuzzing turned up **one real mutation-XSS bug I'd introduced**, now fixed:

**Rawtext-element mXSS** (fixed in the latest commit): non-allow-listed rawtext/RCDATA elements (`title`, `xmp`, `noembed`, `noframes`, `noscript`, `plaintext`) were being *unwrapped* (content kept). Because lol_html tokenizes their contents as a single text run, the `*` element handler never fired on markup inside them, and unwrapping re-emitted that markup **verbatim** — so `<title></title>` came out as a live `` when the browser re-parsed the sanitized output. sanitize-html escaped such kept text, so this was a regression against the old behavior. These elements now drop their whole subtree (like `script`/`style`). Regression tests added on both the Rust and TS sides.

Also addressed two LOW review notes:
- `img`/`source` `src`/`srcset` schemes tightened to `http`/`https`/`data` only (dropped the inadvertently-widened `mailto`/`tel`), matching the old `allowedSchemesByTag` exactly.
- The main allow-list pass is now wrapped in `catch_unwind`, so an internal panic fails closed (drops the body) like the MathJax/SVG stages, rather than unwinding across the FFI boundary.

`cargo test` 52 passed, `pnpm test:unit` green, native module rebuilt.

— Claude

PR #1283, opened 2026-07-16
Switch article extraction to a native Rust module (dom_smoothie)

brendanlong:
## Summary

Replaces the linkedom + `@mozilla/readability` extraction pipeline with a native Rust N-API module wrapping the [dom_smoothie](https://github.com/niklak/dom_smoothie) crate (a maintained Rust port of Mozilla Readability that tracks upstream), following the same pattern as the recently-landed native sanitizer.

- **New workspace package `native/readability` (`@lion-reader/readability`)** mirroring the sanitizer's layout: plain `cargo build` via `build.mjs` (no `@napi-rs/cli`), hand-maintained `index.d.ts`, fail-loud loader, sync + libuv-async entry points. `pnpm build:native` now builds both modules; Docker/CI/bundler wiring follows the sanitizer everywhere it appears.
- **`cleanContent` keeps its signature and behavior gates** (min raw length, min cleaned-text length, `<base href>` extraction + URL absolutization post-pass, `keepClasses`/`charThreshold` options) — only the extraction engine changed. A new `cleanContentAsync` runs extraction on the libuv thread pool with the same ≤10 KB inline shortcut as `sanitizeEntryHtmlInWorker`.
- **`cleanContentInWorker` no longer uses piscina**: it composes `cleanContentAsync` + `sanitizeEntryHtmlInWorker`. The fused clean+sanitize piscina task existed to avoid structured-cloning the cleaned string across the thread boundary twice; with both steps native, each hop is a single N-API string copy (~1% of the work), so the task and its wire schemas are removed. The piscina pool now serves only `parseFeed`.
- `@mozilla/readability` is dropped; `linkedom` stays (narration input builder still uses it).

## Why

Benchmarked on mozilla/readability's 130 real-world test pages (~26 MB HTML):

| Engine | ms/page | Peak RSS (corpus) | Largest page (1.6 MB) |
|---|---|---|---|
| linkedom + readability.js (current) | 13.5 | 242 MB | 49 ms / 101 MB |
| dom-smoothie-js (official WASM) | 21 | 191 MB | — |
| **dom_smoothie native (this PR)** | **2.7** | **46 MB** | **11 ms / 12 MB** |

The WASM route was evaluated first and rejected — WASM's allocator/codegen tax on the allocation-heavy DOM build makes it *slower* than the JS pipeline. Extraction parity was verified on the same corpus: 129/130 pages within 10% of readability.js's extracted text length (the outlier is the pathological `mathjax` test page). dom_smoothie passes Mozilla's test suite with [documented diffs](https://github.com/niklak/dom_smoothie/tree/main/test-pages/not-matching) (byline/heading dedup, more aggressive single-child-div removal).

## Testing

- `pnpm typecheck`, `pnpm lint` clean; knip findings unchanged from master
- `pnpm test:unit`: 131 files / 2436 tests pass — including the existing content-cleaner behavior tests, unmodified, against the new engine
- `pnpm test:integration:local`: 48 files / 693 tests pass (feed pipeline, saved articles, email ingest all exercise the new extractor)
- Worker-thread test rewritten to cover `cleanContentInWorker`'s sanitizeCleaned contract directly (the piscina task it tested is gone)

## Notes for review

- Extraction output is *not* byte-identical to readability.js — titles/bylines/whitespace can differ on some pages. Feed content stores both original and cleaned variants and full-content serving falls back to sanitized original when cleaning fails, so no fast-forward/migration concern; newly-cleaned entries will just reflect the new engine. `SANITIZER_VERSION` is untouched (sanitization behavior is unchanged).
- The native call happens before sanitization; its output flows through the same `withSanitizedEntryContent` chokepoints as before.


brendanlong:
Ran an independent Opus-level review of this PR (full diff + adversarial testing of the built module against the 130-page Mozilla corpus and crafted inputs). Summary:

**Confirmed correct:** option/null/empty-gate semantics vs the old pipeline, the `<base href>` + absolutize flow, error paths, the worker-pool rewiring (nothing still references the removed piscina task or wire schema; the sanitizer-version probe still makes sense for `parseFeed`), `contentSanitized` reuse semantics in `saved.ts`/`full-content.ts`, and all build/deploy wiring (Docker stages, esbuild externals, `serverExternalPackages`, workspace, CI, gitignore).

**One genuine regression found and fixed** (commit cd6eb583): deeply-nested HTML was a CPU-exhaustion DoS. dom_smoothie's scoring is super-linear in nesting depth — a ~30 KB feed entry with 5000-deep `<div>`s ground for minutes, where the old JS pipeline failed fast in ~1.7 s via a caught stack overflow. Fixed with an iterative depth check (cap 512, Blink's flattening limit) before extraction, failing fast to the existing raw-content fallback; regression tests added (bomb rejected in ~30 ms, depth-400 pages still extract).

Also from the review, two defense-in-depth items in the same commit:
- `catch_unwind` around extraction — dom_smoothie has two `partial_cmp().unwrap()` sites on f32 scores; a panic there would abort the whole Node process across the N-API boundary. No reproducer found (theoretical), now mapped to the ordinary null fallback.
- `process-upload.ts` switched to `cleanContentAsync` — it runs on the saved-articles request path and was using the sync form (pre-existing on master, flagged by the review).

— Claude

PR #1286, opened 2026-07-16
Port feed parsing to a native Rust module and delete piscina entirely

### LINKED ISSUE #1285: Port feed parsing to a native Rust module so piscina can be deleted entirely
## Motivation

After #1283 (native readability extraction), the piscina worker-thread pool serves exactly one task: `parseFeed`. Porting feed parsing to a native module — following the now twice-established `native/` N-API pattern (`@lion-reader/sanitizer`, `@lion-reader/readability`) — would let us delete the pool and everything that exists to support it:

- `src/server/worker-thread/pool.ts` (pool init, `getVerifiedPool`, the sanitizer-version probe + timeout machinery, `parseFeedInWorker`) — `cleanContentInWorker`/`sanitizeEntryHtmlInWorker` would move somewhere more natural, since neither touches piscina anymore
- `src/server/worker-thread/worker.ts` + `types.ts` (task protocol, `SerializedParsedFeed` structured-clone serialization/deserialization of Dates)
- `scripts/build-worker-thread.mjs` and the `dist/worker-thread.js` bundle (and with it the whole stale-worker-bundle class of problems the version probe guards against)
- the `piscina` dependency and its esbuild `external` entries

A native `parseFeedAsync` on the libuv thread pool (same ≤10 KB inline shortcut as the other two modules) would serve the request-path callers (`subscriptions.create`, feed preview/refresh in `feeds.ts`, WebSub notification ingest); the background feed poller keeps a synchronous call like it does today.

## Scope / parity notes

The current parsers (`src/server/feed/streaming/`) are bespoke SAX-style (fast-xml-parser) with Lion-Reader-specific extraction that any port must preserve exactly:

- RSS + Atom with `hubUrl`/`selfUrl` (WebSub discovery), `iconUrl`, `ttlMinutes`, syndication hints (`sy:updatePeriod`/`updateFrequency`), entry field derivation (guid/link/author/content-vs-summary selection, date parsing)
- JSON Feed — probably worth **leaving in JS** (`JSON.parse` is native-fast already; a Rust hop buys nothing)
- OPML import is separate and not pool-related — out of scope

Rust options: [`feed-rs`](https://crates.io/crates/feed-rs) (mature, RSS/Atom/JSON in one model, exposes rel-links and extensions — needs a parity audit against our field extraction) vs. a direct port of our parsers on `quick-xml` (more code, exact semantics). The existing unit tests for the streaming parsers are behavior-based and should run unmodified against the native module, the same way the content-cleaner tests did in #1283 — that's the parity gate.

## Caveat worth benchmarking first

Unlike readability (where jsdom/linkedom overhead made native a 5× win), fast-xml-parser is already reasonably quick, so the *speed* win here may be modest — the primary payoff is deleting the piscina machinery and the worker-bundle staleness surface, plus quick-xml's typical 5-10× on the parse itself as a bonus. Worth a quick corpus benchmark (real feeds, including some large ones) before committing, per the approach used for #1283.

— Claude

brendanlong:
Closes #1285.

## What

Adds **`@lion-reader/feed-parser`** (`native/feed-parser/`), the third native N-API module, following the same pattern as the sanitizer and readability modules (standalone crate + `core` sub-crate for `cargo test`, plain `cargo build` via `build.mjs`, hand-maintained `index.d.ts`, uncommitted `.node` artifact). It contains quick-xml SAX ports of the RSS 2.0/RDF, Atom, and OPML state machines — same streaming model as the old htmlparser2 parsers, no DOM, per our memory-minimizing preference.

With feed parsing native, the piscina worker-thread pool served nothing, so it's deleted along with everything that existed to support it:

- `src/server/worker-thread/{pool,worker,types}.ts` — pool init, `getVerifiedPool`, the sanitizer-version probe + timeout machinery, the task protocol, and the `SerializedParsedFeed` date round-trip
- `scripts/build-worker-thread.mjs`, the `dist/worker-thread.js` bundle, and its Dockerfile build/copy steps — the whole stale-worker-bundle hazard class is gone (every thread loads the same `.node` file)
- the `piscina` dependency and its esbuild/Next `external` entries

## Parity (the hard part)

The ports preserve the old parsers' semantics exactly:

- **Unclosed-tag recovery** (`<link>https://…<pubdate>`), RDF item placement, Atom `<source>` shielding, content-vs-summary selection, `sy:`/`atom:link` extraction — the state machines are line-by-line ports.
- **Entity decoding is XML-strict** (5 predefined + numeric character refs incl. the C1/windows-1252 remap, semicolon required): `&nbsp;` etc. stay literal, matching htmlparser2's xmlMode — critical so escaped HTML in feed bodies isn't double-decoded (the LessWrong `&lt;code&gt;` regression class). CDATA stays literal.
- **Dates stay in TypeScript.** No Rust date library reproduces V8's lenient `new Date()`, so the native side returns raw date strings as ordered candidates and the TS wrappers replay the original pubDate/dc:date (published/updated) selection with the unchanged JS parsers — date behavior is byte-for-byte identical, timezone-abbreviation table and all.
- JS `parseInt`/`String.prototype.trim` semantics are reimplemented for `ttl`/`sy:updateFrequency`/text trimming.

Verification:
- The existing streaming-parser unit tests pass **unmodified** against the native module (the parity gate from #1283); full unit (2438) + integration (693) suites green.
- A real-feed corpus check (YouTube, LessWrong, WordPress, xkcd, rust-lang blog, danluu's 9.9 MB feed) produced **canonically identical output** old vs new on all six.
- `cargo test -p lion-reader-feed-parser-core` covers the Rust-side helpers.

## Benchmark (per the "make sure this isn't pointless" caveat)

| feed | size | old (htmlparser2) | native | speedup |
|---|---|---|---|---|
| xkcd.atom | 3 KB | 0.03 ms | 0.02 ms | 1.3× |
| youtube.atom | 36 KB | 0.32 ms | 0.17 ms | 1.9× |
| wordpress.rss | 199 KB | 0.44 ms | 0.56 ms | 0.8× |
| rustblog.atom | 378 KB | 4.19 ms | 2.07 ms | 2.0× |
| lesswrong.rss | 376 KB | 0.71 ms | 1.44 ms | 0.5× |
| danluu (RSS) | 9.7 MB | 105 ms | 54 ms | 1.9× |

The Rust core alone is faster everywhere (lesswrong 0.57 ms, danluu 25 ms); the two sub-1× rows are N-API string-conversion overhead on CDATA-heavy feeds whose entire body is entry content. That conversion now happens once per parse instead of the old piscina round trip's structured-clone + thread hop, and the parse itself runs on the libuv pool off the event loop — 8 concurrent async parses of the 9.9 MB feed complete in ~270 ms total.

## Call-site changes

- Request paths (`subscriptions.create`, feed preview/discovery in `feeds.ts`, WebSub ingest) use the new `parseFeedAsync` (libuv pool, ≤10 KB inline shortcut); the background poller keeps synchronous `parseFeed`.
- The surviving pool wrappers moved to natural homes: `sanitizeEntryHtmlAsync` (sanitize.ts) now owns the inline threshold and replaces `sanitizeEntryHtmlInWorker`; `cleanContentSanitizedAsync` (content-cleaner.ts) replaces `cleanContentInWorker({sanitizeCleaned})`.
- `resanitize-bulk` drops the worker-bundle preflight (no bundle to be stale) and sizes `UV_THREADPOOL_SIZE` instead of the piscina pool.
- **OPML** is native and async too, and parsing moves off the browser: the import preview UI previously ran `parseOpml` client-side; a new `imports.preview` tRPC endpoint parses server-side and the component just uploads the file content (integration-tested). `importOpml` also parses async now.

## Risk notes

- quick-xml is stricter than htmlparser2 on truly ill-formed markup (e.g. a stray `<` in text). The reader is configured maximally lenient (`check_end_names=false`, `allow_unmatched_ends`, `allow_dangling_amp`, tolerant attribute parsing), and the malformed-feed tests pass, but a marginal feed that htmlparser2 silently accepted could now surface as a parse failure — it would show up as ordinary feed-fetch failures in monitoring.
- Docker: the builder gains a cargo cache mount + native/feed-parser copy lines mirroring the other two modules.


brendanlong:
Ran an Opus-level reviewer over the diff (line-by-line comparison of the Rust state machines against the old TS parsers, plus empirical htmlparser2-vs-quick-xml leniency checks). It confirmed parity for date replay, entity decoding, JS-truthiness attribute handling, RDF/`<source>`/media handling, the OPML validation split, the napi layer, the call-site rewiring, and the build/deploy wiring — and found two genuine gaps, both confined to malformed XML:

1. **Unclosed comments/CDATA made the native parser throw** where htmlparser2 recovered a partial parse (the unclosed construct swallowed the rest of the input and parsing ended silently). A feed with an unclosed `<!--` would have flipped from partial-parse to fetch-failure + backoff.
2. **Truncated/misnested feeds lost content**: htmlparser2 implicitly closes all open elements at EOF (a truncated feed keeps its last partially-written entry), closes intermediate elements on a deep end tag, and *ignores* unmatched end tags entirely (so raw `</p>` inside a description doesn't split the text buffer). The pump previously fired one close per raw end event.

Both fixed in 10e1949f: quick-xml `SyntaxError`s (all of which mean "input ended inside a construct") now map to a graceful end-of-input, and the event pump reproduces htmlparser2's end-tag stack dispatch exactly. New Rust core tests + TS parser tests lock the behavior in, with expectations verified against the old parsers; the six-feed real-world corpus still produces canonically identical output.

One accepted (unfixable without a different tokenizer) divergence remains, as noted in the PR description: a stray unescaped `<` in text truncates that text run where htmlparser2 kept it literal — malformed-input only, and the parse still succeeds.

— Claude

PR #1311, opened 2026-07-17
Ship Next standalone output in the runner image (#1305)

### LINKED ISSUE #1305: Shrink the runner image (Next standalone output) to speed up deploys
Follow-up to #1304 (deploy speedup items 1–2). This is item 3.

## Problem

The production image is **318 MB**, and its size shows up in several places in every deploy:

- `COPY node_modules` + `COPY .next` into the runner: ~41s
- Exporting layers: ~35s; push: ~5s
- Machine image pulls during the canary + rolling update (~2m50s of the deploy is rollout)

The runner ships the full pruned `node_modules`, but `dist/server.js`, `dist/worker.js`, `dist/discord-bot.js`, and `dist/migrate.js` are already self-contained esbuild bundles — the big dependency tree is mostly there for the Next.js runtime that `dist/server.js` wraps.

## Proposal

Use Next's `output: 'standalone'`, which traces the server build and emits a minimal `node_modules` subset (typically a fraction of the full tree), and copy that into the runner instead of the full pruned `node_modules`.

Things to work through:

- The custom server (`scripts/server.ts` → `dist/server.js`) wraps Next rather than using the generated `standalone/server.js` — need to confirm the esbuild bundle resolves `next` from the standalone `node_modules`, or adjust the bundle to keep `next` external and point at it.
- The native module symlinks (`node_modules/@lion-reader/{sanitizer,readability,feed-parser}` → `../../native/...`) must survive — standalone output won't know about them, so they likely need explicit handling.
- `pnpm prune --prod` and the runner's `node_modules` COPY can then be dropped entirely.
- Anything else resolved at runtime from `node_modules` that isn't traced by the Next build (e.g. deps only used by the worker/discord bundles — though those are bundled, so in principle none).

## Expected impact

Rough guess: image drops to ~120–150 MB, saving ~30–60s of copy/export/push in the build plus faster machine pulls during rollout.

---
*Filed by Claude on behalf of Brendan (from the deploy-speed investigation).*

brendanlong:
Closes #1305 (follow-up to #1304, deploy speedup item 3).

## What

Ship Next's `output: "standalone"` traced `node_modules` (~76 MB uncompressed) in the runner instead of the full `pnpm prune --prod` tree, cutting the image size and the copy/export/push/pull time it costs every deploy.

- **`next.config.ts`**: enable `output: "standalone"`.
- **`scripts/server.ts`**: in production, hand `next()` the resolved build-time config from `.next/required-server-files.json` via `__NEXT_PRIVATE_STANDALONE_CONFIG` — the same mechanism the generated standalone `server.js` uses. Without this, `next()` tries to load config at runtime and requires `next/dist/compiled/webpack/webpack-lib`, which the trace deliberately omits. This also removes the need for the generated one-line `next.config.js` in the image.
- **`scripts/fixup-standalone.mjs`** (new, run in the builder after `pnpm build`): patches the gaps `@vercel/nft` misses:
  - `html-rewriter-wasm` — `dist/html_rewriter.js` dynamically requires `./asyncify.js`, which isn't traced (instrumentation failed to load without it).
  - `argon2` — the prebuild is resolved per-platform/libc at runtime; the trace only includes the variant matching the build machine, so copy all `prebuilds/` (1.2 MB) to be safe across glibc/musl.
  - `next`'s top-level subpath shims — `dist/server.js` requires `next/constants`; the shim's *target* is traced but the one-line shim file isn't.
  - `@lion-reader/*` workspace symlinks — the trace resolves them to their real `native/` paths, so no `node_modules/@lion-reader` entries exist; recreate them (the Dockerfile still copies the actual native files/binaries from the native-builder stage).
- **`Dockerfile`**: run the fixup, `mv .next/standalone /standalone` so the runner's `.next` COPY doesn't ship a duplicate node_modules, copy the traced node_modules, drop the `pnpm prune --prod` step.
- **`scripts/start-all.sh`**: plain bash job control instead of `npx concurrently` (concurrently isn't in the traced tree). Forwards SIGTERM/SIGINT and propagates the first exit code. Fly runs each process directly via `[processes]`, so this only affects local/default-CMD runs.

## Verification

Assembled the exact runner layout (standalone node_modules + `.next` + `native/` + `dist/`) in a temp dir against a real Postgres/Redis (`pnpm services`) and verified:

- `dist/migrate.js` runs migrations successfully
- `dist/server.js` boots: `/api/health` 200 (DB+Redis healthy), `/login` SSR 200, `/api/openapi` 200 (trpc-to-openapi), static assets + `_next/static` chunks 200, brotli compression active
- `dist/worker.js` boots and processes jobs
- All runtime externals load from the traced tree: argon2 (hash+verify round-trip), sanitizer/readability/feed-parser natives, html-rewriter-wasm, trpc-to-openapi
- `pnpm typecheck`, `pnpm lint`, unit tests (2467 passed), and the e2e suite with `E2E_PRODUCTION=true` (50 passed) all green

Not verified here: an actual `docker build` (no daemon in this sandbox). The first deploy should confirm the final image size — the standalone node_modules is ~76 MB uncompressed vs the previous pruned tree, so the image should land in the neighborhood of the issue's ~120–150 MB estimate.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,6 +106,13 @@ ENV NEXT_PUBLIC_SENTRY_DSN=$NEXT_PUBLIC_SENTRY_DSN
 RUN --mount=type=cache,id=next-cache,target=/app/.next/cache \
     pnpm build
 
+# `output: "standalone"` emitted a traced, minimal node_modules into
+# .next/standalone — the runner ships that instead of the full pruned tree
+# (issue #1305). Patch the gaps the trace misses (dynamic requires, next's
+# subpath shims, the @lion-reader workspace symlinks), then move it out of
+# .next so the runner's .next COPY doesn't pick up a duplicate node_modules.
+RUN node scripts/fixup-standalone.mjs && mv .next/standalone /standalone
+
 # Build custom server bundle (compression + Next.js wrapper)
 RUN pnpm build:server
 
@@ -117,12 +124,6 @@ RUN pnpm build:discord-bot
 
 # Build migration bundle (single optimized JS file)
 RUN pnpm build:migrate
-
-# Prune dev dependencies after build. CI=true lets pnpm remove/recreate the
-# modules directory without a confirmation prompt (there's no TTY in the
-# builder; without it prune fails with ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY).
-RUN --mount=type=cache,id=pnpm,target=/root/.local/share/pnpm/store \
-    CI=true pnpm prune --prod --ignore-scripts
 
 # =============================================================================
 # Stage 5: Production runner (minimal image, no pnpm needed)
@@ -146,8 +147,12 @@ ENV HOSTNAME="0.0.0.0"
 COPY --from=builder /app/public ./public
 COPY --from=builder /app/package.json ./package.json
 
-# Copy production node_modules (already pruned in builder)
-COPY --from=builder /app/node_modules ./node_modules
+# Copy the traced standalone node_modules (a fraction of the full tree; see
+# the fixup-standalone step in the builder). dist/server.js keeps `next`
+# external and resolves it from here; the worker/discord bundles only need
+# their few runtime externals (argon2, html-rewriter-wasm, @lion-reader/*),
+# which the Next server graph also uses, so the trace covers them.
+COPY --from=builder /standalone/node_modules ./node_modules
 
 # The native modules: node_modules/@lion-reader/{sanitizer,readability,feed-parser}
 # are pnpm workspace symlinks into these directories, so they must exist in the
@@ -181,11 +186,10 @@ COPY --from=builder /app/dist/discord-bot.js ./dist/discord-bot.js
 COPY --from=builder /app/scripts/start-all.sh ./scripts/start-all.sh
 RUN chmod +x scripts/start-all.sh
 
-# Generate minimal next.config.js for runtime.
-# The full next.config.ts requires TypeScript and build-time-only deps (next-pwa,
-# sentry). Only compress:false is needed at runtime — everything else (headers,
-# webpack, etc.) is baked into .next/ at build time.
-RUN echo 'module.exports = { compress: false };' > next.config.js
+# No next.config.js in the image: dist/server.js hands next() the resolved
+# build-time config from .next/required-server-files.json (the standalone
+# mechanism) — the traced node_modules doesn't include the runtime
+# config-loading machinery. See scripts/server.ts.
 
 # Switch to non-root user
 USER nextjs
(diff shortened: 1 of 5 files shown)


brendanlong:
Ran an Opus-level review of this PR. Outcome: **no blockers or majors**. The reviewer verified all 35 bare requires from `.next/server` resolve in the standalone tree, that argon2's musl prebuild + node-gyp-build/@phc/format chain is covered, that `required-server-files.json` carries the full resolved config (including `compress: false`), and that the bash `start-all.sh` correctly propagates exit codes and forwards signals.

Two minor findings, both addressed in e2c1b8ad:
- `fixup-standalone.mjs` now removes the traced package subset before the wholesale copy (`cpSync` with `force` throws `EEXIST` on file-type mismatches like symlink vs regular file, rather than overwriting).
- `server.ts` now fails with a clear message when `.next/required-server-files.json` is missing (e.g. `pnpm start` without a prior build) instead of a raw ENOENT.

Noted for a possible follow-up (not this PR): the `jsdom` externals in `next.config.ts` and `build-server.mjs` are dead code — isomorphic-dompurify is gone and nothing in the runtime graph requires jsdom anymore.

Remaining caveat: no docker daemon was available here, so the real Alpine build wasn't run — verification was against a locally-assembled runner layout (details in the PR description). The fixup's unconditional whole-package copies are specifically designed to be robust to the glibc/musl trace difference, but the first deploy is the final confirmation.

— Claude

brendanlong:
Follow-up: built the image for real with **podman 5.7** (no docker daemon here, but podman handles the BuildKit cache mounts fine), closing the Alpine/musl loop from the earlier review:

**Size**
- **130 MB compressed** (`podman save | gzip`), vs the 318 MB baseline from #1305 — right in the issue's 120–150 MB estimate.
- 450 MB uncompressed, of which the `node:26-alpine` base is ~183 MB. The `node_modules` layer is now **72 MB** (uncompressed); the other big layers are `public/` at 84 MB (mostly ONNX wasm) and `.next` at 93 MB, unchanged by this PR.

**Smoke test of the built image** (against real Postgres/Redis via `pnpm services`, `--network=host`):
- `node dist/migrate.js` (the release command): migrations completed
- `node dist/server.js`: `/api/health` 200 with DB+Redis healthy, `/login` SSR 200, `/api/openapi` 200, zstd compression negotiated
- `node dist/worker.js`: started and polling
- argon2 **musl** prebuild hash+verify round-trip, sanitizer v9 / readability / feed-parser natives, html-rewriter-wasm, trpc-to-openapi all load inside the container
- Default CMD (`start-all.sh`): all processes up, health OK; `podman stop` → SIGTERM forwarded, graceful shutdown completed in 1.7 s, exit 143

That removes the last caveat from the review — the standalone image works end-to-end on the actual Alpine/musl target.

— Claude

PR #1337, opened 2026-07-17
Fix image flash on entry navigation (decode cached images before paint)

brendanlong:
## Problem

Paging through demo entries (and reopening after going back to the list) shows a **flash as images load**, even though the network panel shows service-worker/CDN cache hits in ~0 ms. The images are cached, so why do they visibly reload?

## Root cause

It's not a caching problem — it's DOM identity + first-paint decode:

- Each entry gets its own **keyed** `ScrollContainer` (`DemoRouter.tsx`; the real app keys `EntryContent` the same way in `UnifiedEntriesContent.tsx`), and content is injected via `dangerouslySetInnerHTML`. So every navigation mounts **brand-new `` nodes**.
- A fresh `` never paints its bytes on the first frame, **even when cached**: the browser reserves the box, decodes `decoding="async"` off the critical path, and paints a frame later. For sanitizer-stamped `loading="lazy"` images the deferral is worse. That empty→decode gap is the flash. The HTTP cache removes network time, not the decode-and-repaint of a new element.

(Removing the `key` was considered and rejected: it's what resets scroll-to-top per entry, and it wouldn't help anyway — different entries swap the innerHTML and recreate the img nodes regardless.)

## Fix

`useImagePrefetch` now runs in a **layout effect (before paint)** and, for every image already in/near the viewport, sets `loading="eager"` + `decoding="sync"` so the browser decodes a cached image and presents it **atomically on the first paint**. Lazy images below the fold keep the IntersectionObserver upgrade-on-approach path.

Shared across both readers: the hook call moves into `EntryContentRenderer` (the leaf the real app **and** the demo both render through) instead of `EntryContentBody`. The demo's raw, unsanitized HTML has plain-eager images with no `loading="lazy"`, so the selector is broadened from `img[loading="lazy"]` to all `img` to cover them too.

## Verification

- New unit test `tests/unit/frontend/hooks/useImagePrefetch.test.ts` covers the in-viewport eager+sync flip, the 50% near-margin, and that far-below-fold images are left alone.
- Full unit suite green (2526 passed); typecheck + lint clean.
- Driven in a real browser via the demo: the in-viewport hero image gets `eager` + `sync` decode on first load **and again after a back-to-list → reopen round-trip** (the exact reported scenario), while below-fold images stay untouched.

### DIFF
--- a/src/components/entries/EntryArticle.tsx
+++ b/src/components/entries/EntryArticle.tsx
@@ -157,7 +157,7 @@ export function EntryArticle({
         <EntryContentRenderer
           sanitizedContent={contentHtml}
           fallbackContent={fallbackContent}
-          contentRef={contentRef ?? { current: null }}
+          contentRef={contentRef}
           textSizeClass={textSizeClass}
           textStyle={textStyle ?? {}}
         />
--- a/src/components/entries/EntryContentBody.tsx
+++ b/src/components/entries/EntryContentBody.tsx
@@ -21,7 +21,6 @@ import { processHtmlForHighlighting } from "@/lib/narration/client-paragraph-ids
 import { selectDisplayedContent } from "@/lib/narration/select-content";
 import { useNarrationSettings } from "@/lib/narration/settings";
 import { useEntryTextStyles } from "@/lib/appearance/AppearanceProvider";
-import { useImagePrefetch } from "@/lib/hooks/useImagePrefetch";
 import { useSwipeGesture } from "@/lib/hooks/useSwipeGesture";
 import { EntryArticle } from "./EntryArticle";
 import { StickyEntryControls } from "./StickyEntryControls";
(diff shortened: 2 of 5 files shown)


brendanlong:
Ran an independent code review. No correctness bugs or regressions found — ref forwarding (narration highlighting still works), per-entry remount, and SSR safety all check out.

Addressed the one substantive caveat it raised: the synchronous eager + `decoding="sync"` pass now only targets images **actually in the viewport**, not everything within the observer's 50% margin. Using the wide margin there would mass-eager-load and sync-decode images just below the fold on long articles (feed images often lack `width`/`height`, so they collapse to height 0 and stack near the top before loading), defeating lazy loading. Below-fold prefetch-ahead stays with the IntersectionObserver, which upgrades lazy images to eager as they approach — without sync decode, since they aren't visible yet and have no flash to prevent.

Re-verified in a browser: the visible hero gets `eager`/`sync`; below-fold images stay `auto`/`auto`.

— Claude

PR #1349, opened 2026-07-18
Recover from ChunkLoadError by reloading once

brendanlong:
## Problem

`https://lionreader.b-cdn.net/demo/all` intermittently throws:

```
ChunkLoadError: Loading chunk 3241 failed.
(error: .../_next/static/chunks/3241-0c8c6c0909f9cc2f.js)
```

I confirmed with live requests that the failing chunk **404s on both the CDN and the origin** — it isn't a CDN reliability problem, it's version skew. The demo HTML is CDN-cached (`s-maxage=3600, stale-while-revalidate=604800`, see `src/server/http/page-cache.ts`), so a visitor can be served HTML from an older build that references a chunk hash a later deploy replaced. Even though `/demo/all` is fully SSR'd, App Router still ships a client bundle to *hydrate* it into an interactive tree, so it must load those chunks — the fetch is hydration, not deferred rendering.

## Fix

A blocking `<head>` script (`src/lib/chunk-reload.ts`, mounted in the root layout next to the theme/appearance scripts) listens for `ChunkLoadError` via the `error` / `unhandledrejection` events and forces **one** full reload to pull the current HTML + chunk hashes. A `sessionStorage` time-window guard (`RELOAD_WINDOW_MS`) prevents a reload loop if the reload still lands on the missing chunk, while a genuinely later mismatch (a deploy mid-session) is outside the window and still recovers.

**Why an inline head script, not a `"use client"` component:** the failure mode is chunk loading itself. A handler that lives in a chunk and registers from a `useEffect` can't reliably catch its own failure to load — hydration can throw before that effect commits. Inline in `<head>`, the listener is live before any chunk loads and covers every route from one mount point.

Pairs with the deploy-time CDN purge (being added separately) that shrinks the stale-HTML window; this makes any tab that still hits skew self-heal.

## Tests

`tests/unit/chunk-reload.test.ts` executes the generated script against a fake `window` and drives its listeners: reloads once on a chunk error (via both event types, and by message without the `ChunkLoadError` name), never on unrelated errors, at most once per guard window, and again after the window elapses.

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -6,6 +6,7 @@ import { appUrl } from "@/server/config/env";
 import { ThemeProvider } from "@/lib/theme/ThemeProvider";
 import { DEFAULT_THEME, THEME_STORAGE_KEY, THEMES } from "@/lib/theme/config";
 import { buildTextAppearanceScript } from "@/lib/appearance/config";
+import { buildChunkReloadScript } from "@/lib/chunk-reload";
 import "./globals.css";
 
 const geistSans = Geist({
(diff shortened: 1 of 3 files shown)


brendanlong:
Waiting to land this since I think it may already be solved by purging the CDN on deploys.

Issue #1318, opened 2026-07-17
Serve static assets via BunnyCDN (assetPrefix) to offload demo + launch bandwidth


brendanlong:
## Motivation

Prep for a possible Hacker News launch. The capacity benchmark (#1313) found that HN traffic is dominated by **anonymous** landing-page + `/demo` traffic, and that **`/demo` is 100% static** — prerendered HTML, static JS, and hard-coded article data (`src/app/demo/articles/*.ts`), zero DB/API calls. So a demo flood is essentially static-asset serving, which a CDN can absorb almost entirely — cutting Fly egress (~$0.02/GB) and offloading request-serving CPU from the app machines.

A BunnyCDN pull zone already exists: **https://lionreader.b-cdn.net/** (origin should be the app, e.g. `https://lionreader.com`).

## Plan: `assetPrefix` (safe, no risk to dynamic routes or SSE)

Point Next.js `assetPrefix` at the pull zone so only immutable, content-hashed bundles under `/_next/static/*` load from the CDN. Every dynamic route, `/api/*`, and the SSE stream stay on the origin, untouched.

1. In `next.config.ts`, set `assetPrefix` from an env/build arg (inlined at build time), e.g. `ASSET_PREFIX=https://lionreader.b-cdn.net`. Leave it unset in dev.
2. Configure the Bunny pull zone: origin = the app; **respect origin `Cache-Control`** (Next serves `/_next/static` as `immutable, max-age=31536000`; images/fonts already get 1-day + SWR from `next.config.ts` `headers()`).

## ⚠️ The gotcha that will silently break it: CSP

`src/server/http/csp.ts` builds a strict per-request CSP (#1275). Two directives must be updated or the browser blocks CDN assets:

- **`script-src`** uses `'nonce' 'strict-dynamic'` → CDN-hosted **JS works as-is** (Next's nonce'd bootstrap dynamically loads chunks; strict-dynamic propagates trust regardless of origin). Optionally add `https://lionreader.b-cdn.net` as the pre-CSP3 fallback (ignored when strict-dynamic is honored).
- **`style-src 'self' 'unsafe-inline'`** → CDN-hosted **CSS (`/_next/static/css/*`) is BLOCKED**. Must add `https://lionreader.b-cdn.net`.
- **`font-src 'self' data:`** → CDN-hosted **fonts (`/_next/static/media/*` via `next/font`) are BLOCKED**. Must add `https://lionreader.b-cdn.net`.
[...]

PR #1351, opened 2026-07-18
Fix public-page background flash: stale CDN HTML + unthemed error pages (#1350)

### LINKED ISSUE #1350: Background flash on public-page navigation: stale CDN HTML references deleted build assets, hydration fails into unthemed global-error
## Symptom

Navigating between the demo, sign-in, registration, and legal pages sometimes shows a brief flash where the background is lighter than the final page in dark mode (darker in light mode). Reproduces most often via the CDN (lionreader.b-cdn.net), more in Firefox than Chromium. Several rendering-side fixes (blocking theme script, SSR/hydration matching, service-worker changes) made no difference.

## Root cause (verified by frame-by-frame capture)

The flash is not a browser paint artifact — it is the app's own **`global-error` page**, which rendered an unthemed white `<html><body>` with a blue button.

The failure chain:

1. `page-cache.ts` marked the public pages `s-maxage=3600, stale-while-revalidate=604800`, so Bunny could serve HTML cached before a deploy for up to an hour (and up to a week while revalidating).
2. Every deploy deletes the previous build's hashed `/_next/static` assets, so that stale HTML references route chunks that now **404** (`text/plain` + `nosniff` → Firefox `NS_ERROR_CORRUPTED_CONTENT`).
3. The SSR HTML still paints correctly for a moment, then hydration hits the missing chunk and Next swaps the whole document to `global-error` — default-white. In dark mode that's a bright flash (or a stuck white error page for cold-cache visitors); Chrome's paint-holding and its own immutable-cached chunks from prior visits made it rarer there.

Captured on production: `/terms` (dark) → navigate → `/login` paints dark for ~60 ms → white "Something went wrong" page, persisting. The console shows the blocked chunk (`/_next/static/chunks/app/(auth)/login/page-*.js` 404).

Frequent deploys are what made this look random: each deploy re-opens the window until the CDN revalidates, and returning browsers were shielded by their own immutable chunk cache.

## Fixes (PR to follow)

- Cut the public-page CDN policy to `s-maxage=300, stale-while-revalidate=60`.
- Purge the Bunny pull zone after each successful deploy (needs `BUNNY_API_KEY` + `BUNNY_PULL_ZONE_ID` secrets).
- Theme `global-error` self-containedly (inline style, prefers-color-scheme + stored theme) and auto-reload once (guarded) before showing it — a plain reload fixes the stale-HTML case since browsers send `max-age=0`.
- Add a themed root `not-found` page (the built-in one is also unthemed white).

## Follow-up (the fully-correct fix)

Make old build assets survive deploys — e.g. upload each build's `/_next/static` to object storage (Bunny storage zone / S3) and serve via `assetPrefix`, keeping the last few builds. Then stale HTML is always hydratable and the CDN TTLs could be raised again.

---
*Filed by Claude 🤖*

brendanlong:
Closes #1350.

## What the "flash" actually was

Two related mechanisms, both diagnosed frame-by-frame (Xvfb framebuffer sampling + headed Firefox):

1. **Stale CDN HTML after deploys** (the error-causing one): the CDN could serve public-page HTML cached before a deploy for up to an hour + a week of `stale-while-revalidate`, but every deploy deletes the previous build's hashed `/_next/static` assets. The 404'd route chunk makes hydration fail and Next swaps the document to the unthemed white `global-error` page.
2. **The one-frame flash** (from the screen recording): ~2 frames (33ms) of the browser's default canvas (`#2b2a33` on dark-themed Firefox) painted between the old document and the new page's first paint. Colored by the *browser* theme — lighter than our dark canvas, darker than our light one. No page CSS can recolor that gap.

## Changes

- **`deploy.yml`**: with `BUNNY_STORAGE_ZONE_NAME`/`BUNNY_STORAGE_ZONE_PASSWORD` secrets set, the deploy splits into build `--build-only --push` → extract `/_next/static` from the exact built image → upload to the Bunny storage zone (`scripts/upload-bunny-static.sh`, additive — old builds' assets stay servable forever) → release → purge the pull zone. Assets are always in storage before HTML referencing them goes live. Without the secrets, the original single-step deploy runs unchanged. Serving the storage zone needs the Terraform-side pull zone + `/_next/static/*` edge rule tracked in #1318.
- **`globals.css`**: `@view-transition { navigation: auto }` — Chrome 126+/Safari 18.2+ hold the old page until the new one renders, eliminating the between-documents canvas flash (verified: navigation becomes a smooth 0.1s blend with zero default-canvas frames). Firefox inherits when its flagged cross-document implementation ships. Reduced-motion users get the hold-and-swap without the fade.
- **`global-error.tsx`**: themes itself via a self-contained inline `<style>` (`prefers-color-scheme` + stored theme; it can't rely on globals.css, which may itself be a dead asset) instead of rendering default-white.
- **`not-found.tsx`** (new): themed root 404 — the built-in one is also unthemed white.
- CDN TTLs in `page-cache.ts` are unchanged (an interim shortening was reverted): the CDN won't take traffic until the asset pipeline + edge rule are live, at which point stale HTML is harmless because old assets remain servable.

## Verification

- Upload script functionally tested against a local PUT server (paths with parentheses, binary payloads, empty-dir failure).
- Themed error/404 pages and the view transition verified on a local production build with headed browsers under Xvfb, sampling the X framebuffer at ~200Hz for flash frames.
- Typecheck, lint, unit tests, raw-color/theme-token checkers pass. Two Opus review passes (findings applied; no blockers).


brendanlong:
Addendum after analyzing a real screen recording of the flash (frame-by-frame): the user-visible flash is **~2 frames (33ms) of the browser's default canvas** (`#2b2a33`, Firefox dark theme) painted between the old document and the new page's first paint — separate from the chunk-404 → global-error failure this PR already fixes (that one was causing real errors, but shows a full error page rather than a one-frame flash). The default canvas is colored by the *browser* theme, not the page, which is exactly why it reads lighter than dark mode and darker than light mode. No page CSS can recolor that gap.

Pushed one more commit: `@view-transition { navigation: auto }` in globals.css. Chrome 126+ / Safari 18.2+ snapshot the old page and hold it until the new page renders — the gap disappears entirely (verified in headed Chromium with X framebuffer sampling: navigation is a smooth 0.1s blend, zero default-canvas frames). Firefox has cross-document view transitions behind a flag and inherits the fix when it ships; until then the residual one-frame flash in Firefox is browser-level behavior with no page-side fix.

— Claude

Issue #1328, opened 2026-07-17
SSR the auth pages (register, complete-signup) so signupConfig-gated content renders on first paint


brendanlong:
## Problem

The auth pages are `"use client"` components whose content depends entirely on a **client-side** tRPC query (`trpc.auth.signupConfig.useQuery()`). Nothing prefetches `signupConfig` on the server or hydrates it into the React Query cache, so during SSR the query has no data and the config-gated UI is missing from the initial HTML. It only appears after hydration once the client query resolves.

The most visible symptom: on **EU-restricted instances** (`EU_RESTRICTED=true`), the "I confirm that I am not located in the European Union" checkbox and its explanatory footnote on `/complete-signup` are **absent from the SSR render** and pop in a moment later.

### Where it comes from

`src/app/complete-signup/page.tsx`:

```ts
const { data: signupConfigData } = trpc.auth.signupConfig.useQuery();
const euRestricted = signupConfigData?.euRestricted ?? false;
```

Both the EU checkbox (`euRestricted && …`) and the `<EuRestrictionReason />` footnote are gated on `euRestricted`, which is `false` during SSR → neither is in the initial HTML.

`src/app/(auth)/register/page.tsx` has the same client-only dependency. It masks the flash with a full `if (isLoadingConfig) return <Loading/>` gate, so you don't see a *partial* form — but the SSR HTML still contains no real config-driven content (EU warning, provider buttons, invite-required state), just a "Loading..." placeholder.

## Proposed fix

SSR these pages so `signupConfig` is resolved on the server and correct in the first paint. Options considered:

1. **Server-prefetch + hydrate** `auth.signupConfig` into the RQ cache for these routes. Most general, but the app currently has no server-prefetch/dehydration wiring for tRPC, so this introduces a new pattern.
2. **Read config server-side and pass as props** — make `page.tsx` a server component that reads `signupConfig` from `@/server/config/env` and renders the client form with the flags as props. `signupConfig` is static server env (not per-user), so this is simple and gives fully correct SSR with no flash. Likely the cleanest fix for the EU flag specifically.
3. Gate on loading like the register page already does — hides the flash but the SSR HTML still lacks the content.

[...]

PR #1313, opened 2026-07-17
Capacity load-test benchmark (bench/)

brendanlong:
## What & why

Ahead of a possible Hacker News launch, this adds **capacity benchmarks** to estimate how many users Lion Reader can support per hardware tier, how a traffic spike behaves given Fly's shared-CPU throttling, and whether the feed-fetching **worker** can keep up as feeds grow.

Full writeup + results: `~/wiki/pages/lion-reader-capacity-benchmark-2026.md`.

## Headline findings (local, real prod per-user params)

- **App: the bottleneck is the single Node event loop (CPU), not Postgres** — pg pool never queued. **2 dedicated cores ≈ ~200 concurrent active users / ~280 rps** → **~6,700 registered** at the observed 3% concurrent fraction. Capacity scales ~linearly with app-machine count.
- **Worker: ~15× headroom.** One worker at `WORKER_CONCURRENCY=1` sustains ~21k feeds at hourly cadence (worst case) vs ~1,444 active feeds today. Concurrency scales it near-linearly to 200k+. The real reason to raise concurrency is slow-feed head-of-line blocking, not throughput.
- **The real HN risk is the CURRENT `shared-cpu-2x` tier** — 6.25%/vCPU throttle collapses sustained throughput to ~1/8–1/16 of dedicated cores after the burst balance drains. **Move the app off shared CPU + `fly scale count` before posting.**

## The harness (`bench/`)

- `characterize-prod.sql` — read-only prod aggregates → `prod-params.json` (gitignored).
- `seed-bench.ts` / `loadtest.ts` / `lib/trpc-client.ts` — request-path load test (wire-identical tRPC, server-side-gated knee).
- **`worker/`** — worker throughput benchmark: mock feed server + backlog-drain timing across a `WORKER_CONCURRENCY` sweep (FRESH + 304 phases).
- `fly-bench/` — turnkey runbook + `performance-2x/4GB` fly.toml for the real cloud test.

Independent review pass + fixes applied (unreadOnly list, starred seed, histogram series-summing, etc. — see comments).

## Status

- [x] Harness (request-path + worker) built, reviewed, fixed
- [x] Real prod params characterized + plugged in
- [x] Local results: 2-core capacity knee + worker throughput sweep
- [ ] **Optional (agent token can't create Fly apps):** real `performance-2x/4GB` run via `bench/fly-bench/` runbook — local 2-core run is a solid proxy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

— Claude

Issue #984, opened 2026-07-04
proxy.ts auth gate is dead code — isPublicPath() returns true for every path


brendanlong:
## Bug
`isPublicPath()` in `src/proxy.ts` returns `true` for **every** path, so the proxy's session-cookie redirect (proxy.ts, the `if (!sessionCookie?.value) redirect('/login')` block) is unreachable.

`PUBLIC_PATHS` contains `"/"` (the landing page), and the prefix rule is:

```ts
if (publicPath.endsWith("/") && pathname.startsWith(publicPath)) return true;
```

`"/"` ends with `/`, and **every** pathname starts with `/`, so the first list entry matches everything. The proxy always falls through to `NextResponse.next()`.

## Impact
No data leak today — the real auth gate is the `(app)` route-group layout (`src/app/(app)/layout.tsx`, `isAuthenticated()` → `redirect('/login')`) plus tRPC session validation — but the proxy is doing nothing. It's meant to provide an early unauthenticated redirect and is pure theater. (This also explains why `/oauth/*` and `/.well-known/*` are reachable while `/settings`/`/all` redirect: the redirects come from the app layout, not the proxy.)

## Fix (needs care)
Make the `"/"` entry exact-match only (don't treat it as a prefix):

```ts
if (publicPath === "/") return publicPath === pathname; // landing page is exact
```

**But** activating the gate would start redirecting any non-public top-level route without a session cookie — audit first. At minimum `/demo`, `/save`, `/terms`, and `/extension/*` are public pages NOT currently in `PUBLIC_PATHS` and would break. So: fix the `/` match, do a full audit of public top-level routes, add them to `PUBLIC_PATHS`, and add unit tests for `isPublicPath` (there are none). The MCP OAuth PR (#981) already added `/authorize`, `/token`, `/oauth/` — those become load-bearing once this is fixed.

Surfaced during a code review of #981; pre-existing.

— Claude

PR #1360, opened 2026-07-18
Split root layout: static, CSP-relaxed public pages (#1359)

### LINKED ISSUE #1359: Split root layout: static, CSP-relaxed public pages (demo, login, signup, ToS, privacy) to cut origin render cost
## Motivation

The capacity benchmark (#1313) measured the unauthenticated SSR pages — the pages an HN flood actually lands on. On 2 dedicated cores the app serves only **~210 renders/s** for them (*lower* than the authenticated API, because rendering a full React page costs more than a tRPC JSON response), and each is CPU-bound at ~11 ms with **zero DB/SSE usage**. They're forced into dynamic (per-request SSR) rendering purely by the shared root layout — nothing on the pages themselves needs it.

If we make these pages statically prerendered, the origin serves them as static files (near-zero CPU), which drastically cuts launch-day load and may make a full-site CDN unnecessary. (Static **assets** are the separate, bigger bandwidth win — #1318.)

## Root cause: the shared root layout poisons every route

`src/app/layout.tsx` reads three Dynamic APIs on every request:
- `headers()` → the per-request CSP nonce (#1275) for the two inline `<head>` scripts (`textAppearanceScript`, `swScript`).
- `cookies()` → the announcement-dismissed cookie.
- `getAnnouncement()` → Redis (admin announcement banner).

Because it's in the tree for **every** route, all of them become dynamic. Today only `favicon/manifest/robots/sitemap` appear in `.next/prerender-manifest.json`.

## Plan (agreed): split into two root layouts via route groups

Give the SPA and the public pages separate root layouts:

- **`(spa)` group** — the authenticated app + auth/oauth/utility routes. Keeps the current layout: per-request **nonce** (strict CSP — it renders sanitized-but-untrusted entry HTML), announcement banner, dynamic.
- **`(public)` group** — `demo`, `login`, `register` (signup), `terms`, `privacy`. New static-friendly layout: **relaxed static CSP**, **no announcement banner**, client-side theme.

### Next.js mechanic + the main cost

Multiple root layouts require **removing `src/app/layout.tsx`** and giving *each* top-level group its own `layout.tsx` (with `<html><body>`). That means **every** current top-level route must move into a group with a root layout: `admin`, `authorize`, `complete-signup`, `extension`, `mcp`, `oauth`, `revoke`, `save`, `token`, and `app/page.tsx` (the `/`→`/demo/all` redirect) go into `(spa)` (or their own group); `api/` is route handlers and is unaffected. This repo-wide regroup is the bulk of the work and the main risk. Crossing between root layouts is a hard navigation (full reload) — fine here, since public↔app transitions are login flows already.

## CSP: relax for the public group; do NOT try global hashing

Global hash-based CSP is impractical with the App Router: it streams per-response inline `self.__next_f.push(...)` flight-data scripts whose bytes vary by page/data, so they can't be hashed at build time — a nonce is the only practical mechanism for dynamic pages (which is why we use it). So keep the nonce for `(spa)`.

For `(public)`, serve a **static relaxed CSP** — `script-src 'self' 'unsafe-inline'` (plus `'wasm-unsafe-eval'`/`'strict-dynamic'` as today). This is defensible because **these pages render zero untrusted/user-supplied HTML** — the strict nonce exists to backstop the sanitizer on entry-rendering pages, and that surface doesn't exist here. Wiring:
- `src/proxy.ts` middleware: skip nonce generation for the public routes.
- Set the static CSP for those paths via `next.config.ts` `headers()` (keep the maintenance-mode static CSP in `scripts/server.ts` consistent).
- The public layout drops `nonce={...}` from its inline scripts (allowed by `'unsafe-inline'`).

> Alternative if we want to preserve the strict guarantee on public pages: hash-CSP with a **post-build step** that extracts every inline-script hash (ours + Next's baked flight data) into the CSP. More machinery, brittle on any content/dep bump. Not recommended unless the relaxed CSP is deemed unacceptable.

## Per-page: what's static vs dynamic (verified)

- **demo** — already no DB, client-only, hardcoded article data → **static**. (Highest-traffic page; the main win.)
- **login / register (signup)** — already `"use client"`: errors (`useState` + tRPC `useMutation.onError`), URL params (`useSearchParams` — client hook, not the server prop), signup config (client `useQuery`), and submission are **all client-side**. Nothing reads a server Dynamic API once out of the nonce layout → **can be static**. This resolves the "they need to show errors" concern — errors keep working client-side.
- **terms / privacy** — static content → **static**.

## Maintenance / "site down" — already handled upstream, no conflict

The site-down page is enforced by the HTTP-server gate (`src/server/maintenance/server-gate.ts`) **before** Next, keyed on pathname + cookie — independent of static vs dynamic. `login`/`register` are **not** exempt, so they correctly return the 503 site-down page during maintenance; `demo`/`terms`/`privacy` **are** exempt (stay up). So making these pages static does **not** lose the maintenance behavior for origin-served requests.

> Caveat: this only holds while the page is **origin-served**. A CDN edge-cached page bypasses the origin gate — so if we later CDN-cache `login`/`register`, they'd stop showing the site-down message (a reason to keep those origin-served-static; `demo` is exempt anyway, so caching it is safe). Tie-in to #1318.

## Announcements

The public layout omits the SSR announcement banner. If we ever want announcements on public pages, fetch client-side (small client component) — probably unnecessary for demo/login.

## Acceptance criteria

- [ ] `demo`, `login`, `register`, `terms`, `privacy` appear in `.next/prerender-manifest.json` (statically generated).
- [ ] No CSP violations in the console on those pages; inline theme + SW-registration scripts run; theme has no flash.
- [ ] login/register still show OAuth + validation errors and honor `?redirect` / `?invite` / `?error` / `?registered`.
- [ ] Maintenance mode: `login`/`register` show the 503 site-down page; `demo`/`terms`/`privacy` stay up.
- [ ] Authenticated app unchanged: per-request nonce, announcement banner, dynamic rendering, strict CSP over entry HTML.
- [ ] `bench/anon-pages.ts` shows the public pages served at much higher rps / near-zero origin CPU after the change.

## Security review

Relaxing the CSP on the public routes touches the XSS-backstop model (#1275 / SECURITY.md). Reviewer should confirm none of `demo`/`login`/`register`/`terms`/`privacy` render user-supplied content as HTML (they don't today: demo articles are dev-authored constants; the forms render user input only as escaped React text). Document this invariant so a future change that adds untrusted HTML to a public page re-triggers the strict-CSP requirement.

## Out of scope

- Static-**asset** CDN (`assetPrefix` → BunnyCDN) — that's #1318, a separate and larger bandwidth win. This issue is only about the origin **render** cost of the public HTML.

---
Context: capacity benchmark #1313 and the wiki page `lion-reader-capacity-benchmark-2026` (anonymous-page load section). Prereq insight for the "cache the HTML" half of #1318.

— Filed by Claude

brendanlong:
Closes #1359.

## What this does

Splits the single root layout into two route groups so the public pages are **statically prerendered** and served from the build cache instead of paying per-request SSR (#1313 measured the anonymous pages CPU-bound at ~11 ms / ~210 rps on 2 cores):

- **`src/app/(spa)/`** — the authenticated app + auth/OAuth/utility routes (`(app)`, `/auth/oauth/*`, admin, complete-signup, extension, oauth, save, the `/` redirect). Unchanged behavior: per-request CSP nonce (#1275), dynamic rendering, strict CSP over entry HTML.
- **`src/app/(public)/`** — demo, login, register, terms, privacy. New root layout that reads no request data. The shared document shell (fonts, metadata, theme/appearance/SW scripts, ThemeProvider) lives in `src/app/root-document.tsx` so the two layouts can't drift.

**Now in `prerender-manifest.json`:** `/login`, `/register`, `/terms`, `/privacy`, `/demo`, `/demo/all`, `/demo/highlights`, all `/demo/entry/*` articles, all `/demo/subscription/*` and `/demo/tag/*` lists, `/_not-found`.

## Two-tier CSP

`src/proxy.ts` now applies (policies in `src/server/http/csp.ts`):
- **Strict nonce'd policy** (default, every dynamic route) — exactly as before.
- **Relaxed static policy** (only `isPublicStaticPath`: the `(public)` routes) — `script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval'`, **no nonce and deliberately no `'strict-dynamic'`** (in CSP3, `'strict-dynamic'` makes browsers ignore `'unsafe-inline'` and the host allowlist, which with no nonce would block every script). All other directives identical.

This is safe **only** because the public pages render zero user-supplied HTML (demo articles are dev-authored constants; the forms render user input as escaped React text). The invariant is documented in SECURITY.md, `(public)/layout.tsx`, and `csp.ts`; a route added to `(public)` without a proxy-list entry fails safe (gets the strict CSP and visibly breaks). The `POST /register` OAuth-DCR rewrite is untouched.

## The one hard part: demo `?entry=` articles

The issue assumed demo was "client-only", but `/demo/all` (and subscription/tag/highlights) read **server `searchParams`** to SSR the selected article for SEO/OG — which forces dynamic rendering no matter what the layout does, and `/` redirects to `/demo/all?entry=welcome`, the flood landing page. Fix without changing any public URL:

- New prerendered **`/demo/entry/[entryId]`** route (`generateStaticParams` over all 25 articles, per-article metadata + the same `/demo/all?entry=` canonical).
- A **`beforeFiles` rewrite**: any `/demo/*?entry=<id>` request is served from that route. Browser URL, canonicals, sitemap, and DemoRouter's client-side routing are unchanged.
- `dynamicParams = false` so attacker-controlled ids can't grow the route cache on demand. Behavior change: an unknown-but-well-formed `?entry=` id now 404s (previously rendered the list; malformed values still fall through to the static list).

## Accepted trade-offs (login/register)

Making these static drops two request-time behaviors from the old `(auth)` layout:
- **No server-side redirect-to-`/all` for already-authenticated visitors** — they see the form; logging in again works, and the app layout keeps its own guards.
- **The #1328 SSR config prefetch is reverted**: the form itself is in the static HTML (no loading flash for the primary content), but OAuth buttons / invite-required state / the signup link are client-fetched and pop in after hydration. `?redirect`/`?invite`/`?error`/`?registered` all still work (client-side; `?redirect` is read at submit time from `window.location`).

The OAuth callback/complete pages keep the old dynamic layout under `(spa)/auth/`.

## 404 handling

With two root layouts there's no root layout for a root `not-found.tsx`, so this uses Next's `global-not-found` convention (**`experimental.globalNotFound`** — the flag exists precisely for the multiple-root-layouts case). The page is script-free (a static document can't carry the strict CSP's nonce): themed via the `prefers-color-scheme` fallback, plain `<a>` home link. Edge cases: explicit-dark-on-light-OS users get a light 404, and the blocked (unneeded) bootstrap scripts log CSP violations in the console on 404s only.

## Maintenance mode

Unchanged — the gate runs in the custom HTTP server before Next, so static vs. dynamic is irrelevant origin-side. `login`/`register` still 503; demo/terms/privacy stay up. Noted in `src/server/http/CLAUDE.md`: if HTML is ever CDN-cached (#1318), keep `login`/`register` origin-served or they'd bypass the gate.

## Verification

- `next build`: all public routes `○`/`●`; smoke-tested the prod server: rewrite serves the right articles, `/`→`/demo/all?entry=welcome` intact, static ~3 ms vs ~8 ms dynamic locally, CSP headers correct per tier.
- Playwright against the prod build: `/demo/all?entry=welcome`, `/login`, `/register` hydrate with **zero console errors**; demo sidebar/article interactive; invite-required renders client-side.
- `pnpm test:unit` (2547 ✓, incl. new proxy CSP-tier tests), `pnpm test:integration:local` (695 ✓), `pnpm test:e2e:local` (60 ✓ — `security-headers.spec.ts` and `auth-ssr.spec.ts` rewritten for the new contract, incl. a new full login-form flow honoring `?redirect`, and zero-CSP-violation checks on the public pages), `pnpm typecheck`, `pnpm lint`.

## Out of scope

Static-asset CDN / HTML edge caching (#1318). `bench/anon-pages.ts` from the issue's acceptance criteria doesn't exist in-repo; the prerender manifest + latency smoke test stand in for it.


brendanlong:
Pushed two follow-ups (discussed in session) that recover the request-time behaviors the split dropped, without giving up static serving:

**1. SSR'd auth config is back — rendered at process startup instead of per request.** The `(public)/(auth)` layout prefetches `auth.signupConfig`/`auth.providers` through a new request-free `createStaticHydrationHelpers` (env-only procedures, anonymous context — no Dynamic APIs, so `/login`/`/register` stay in the prerender manifest). Since `next build` bakes build-machine env, `scripts/server.ts` re-renders both pages once per boot via a per-process-secret-guarded `/api/internal/revalidate-public` route (`revalidatePath` + warm). Verified against the prod build: with `ALLOWED_PUBLIC_SIGNUP_PROVIDERS=email` set **only at runtime**, the served static HTML shows the signup link/form (~3ms per request). OAuth buttons, invite-required state, and the EU notice are all in the served HTML again — the #1328 pop-in trade-off is gone. `/register` statically renders the no-invite variant; hydration threads `?invite=` in.

**2. Signed-in redirects are back — in the proxy, with full validation.** For GET/HEAD `/`, `/login`, `/register`: no session cookie = one header check (and `/` 307s straight to the demo with no Next render, ~3ms); a present cookie is validated (Redis-first, DB fallback, lazily imported) and only a *valid* session redirects to `/all`/`/complete-signup`. Invalid cookies and validation errors fall through to the page — no redirect loop, and deliberately not the #984 presence-gate: it protects nothing, the layout guards stay authoritative. `POST /register` (the DCR rewrite) is explicitly untouched.

Also: `/api/internal` is maintenance-gate exempt (secret-guarded, no DB) so a boot during maintenance doesn't strand stale HTML.

Tests: unit 2549 ✓ (new proxy-redirect + gate cases), integration 695 ✓, e2e 62 ✓ (raw-HTML SSR-config pins restored, signed-in redirect from all three paths, dead-cookie no-loop test), typecheck/lint clean, prod-build smoke verified.

— Claude

PR #1363, opened 2026-07-18
CDN caching: neutralize public-page s-maxage; serve demo images from CDN

brendanlong:
## Summary

CDN-caching fixes for the `(public)` route group, plus making content-hashed assets `immutable`. Evolved over review into the simplest form of each.

### 1. Stop shared-caching public prerendered HTML/RSC (deploy-safety + #1318)

Next stamps `Cache-Control: s-maxage=31536000` on the statically-prerendered public pages (demo/login/register/terms/privacy) **and their RSC payloads**. Our Bunny pull zone wraps the whole site and honors origin `Cache-Control` (keying on `_rsc`/`entry`), so that year-long shared lifetime on build-coupled content is live — cached HTML 404s its per-deploy chunks, RSC Flight payloads version-skew, and an edge-cached `/login`/`/register` bypasses the maintenance gate (#1318).

Fixed **at the source** in `src/proxy.ts`: it sets `Cache-Control: private, no-cache` on the same `isPublicStaticPath` responses where it already sets the static CSP. Next only stamps its default when the response has no Cache-Control yet (`!res.getHeader(...)` in `send-payload`), so the middleware header wins — no custom-server rewrite needed. `private` keeps it out of shared caches; `no-cache` lets the browser revalidate so a deploy can't leave it booting a stale document.

(`?_rsc=<hash>` is a router-state hash, not a deploy id, so it can't make RSC safe to shared-cache across deploys.)

### 2. Serve demo hero/OG images via static imports

Each article now `import`s its hero + OG image from `./images/`, so Next content-hashes them into `/_next/static/media` and serves them from the CDN automatically — **no manifest, no `?v=` buster, no Bunny query-string config**. Static imports also give a build-time guarantee the file exists (missing → build failure, not a runtime 404). Images are emitted as separate hashed files, never inlined as `data:` URIs. Deleted the helper, generated manifest, generator script, CI check, and `NEXT_PUBLIC_ASSET_PREFIX`.

### 3. Make `/_next/static` immutable

Splitting the images out surfaced that the `headers()` image/font-extension rule was matching `/_next/static/media/*` and downgrading **fonts** (and the new imported images) to a 1-day TTL. Added an explicit `immutable` rule for `/_next/static/:path*`; the 1-day + SWR rule now applies only to genuinely-unhashed `public/` assets (favicon/icons, emojis, social-preview).

## Verified on a production build

| Request | Cache-Control |
|---|---|
| `/demo/all`, `/login` HTML + RSC | `private, no-cache` |
| JS chunk, font woff2, demo hero/OG png | `public, max-age=31536000, immutable` |
| `public/` favicon, social-preview | `public, max-age=86400, stale-while-revalidate=604800` |

Hero baked into HTML as `…/_next/static/media/welcome.<hash>.png` (0 `data:` URIs); OG meta as an absolute CDN URL. Full unit suite green.

## Deploy notes

- No Bunny config change needed anymore (previous revision required adding `v` to the cache key — obsolete).
- Fonts move from 1-day to `immutable` as a side effect (they're content-hashed).


brendanlong:
Verified against a real production build (`pnpm build` + `NODE_ENV=production` custom server, `ASSET_PREFIX=https://cdn.example.test`):

| Request | Cache-Control (before → after) |
|---|---|
| `/demo/all` (HTML) | `s-maxage=31536000` → `private, no-cache` |
| `/demo/all` (RSC, `RSC: 1` → `text/x-component`) | `s-maxage=31536000` → `private, no-cache` |
| `/login` (HTML) | `s-maxage=31536000` → `private, no-cache` |

Demo image URLs baked into the prerendered HTML:
- hero → `https://cdn.example.test/demo/welcome.png?v=ae687520`
- OG → `https://cdn.example.test/demo/welcome-og.png?v=3932b102`

(hero and OG hash independently, so an update to one busts only that file). The custom-server interception wins over Next's emitted header, and the CDN prefix + `?v=` hash is baked into the static HTML so server prerender and client re-render stay byte-identical.

— Claude

Issue #1280, opened 2026-07-16
Full-content markdown preference leaks page chrome for whole-page markdown dumps (e.g. turntrout/Quartz)


brendanlong:
## Summary

Our full-content fetch prefers `Accept: text/markdown` and treats any `text/markdown` response as already-clean content, **skipping Readability entirely**. For sites whose markdown endpoint returns a *whole-page dump* (nav, tags, "about", title/byline, table of contents, …) rather than clean article markdown, all of that chrome flows straight through into the stored content.

**Repro:** https://turntrout.com/why-i-left-google-deepmind — the stored full content leads with the site nav ("The Pond", Search), Tags, "About this post", "Links to this page", the title/byline, and the full Table of Contents. Firefox reader mode strips all of it.

## Root cause

- `HTML_ACCEPT_HEADER` (`src/server/http/fetch.ts:238`) is `text/markdown,text/html,…`.
- turntrout honors it — returns `content-type: text/markdown; charset=utf-8` (`vary: accept`), to our UA too.
- `fetchFullContent` (`src/server/services/full-content.ts:139-153`) treats markdown as clean and **skips Readability**, doing only `processMarkdown` → HTML.
- turntrout's markdown is the entire page rendered to markdown. The article body (`In January, Department of Homeland Security…`) doesn't start until ~line 80; everything before it is chrome.

The wrong assumption is **"supports `text/markdown`" ⇒ "returns a clean article."** True for LLM-oriented endpoints, false for Quartz sites (which render the whole page to markdown). The markdown preference was added deliberately in a9a1beff for the clean-markdown case.

## Why the easy fix doesn't work

Running Readability on the markdown→HTML **does not** strip the chrome (verified): converting to markdown flattens everything into top-level `<p>`/`<blockquote>`/`<h2>` blocks and destroys the `<nav>`/`<aside>`/sidebar structure Readability relies on to distinguish chrome from content. Once on the markdown branch, a clean article can't be recovered.

For comparison, the plain **HTML + Readability** path on the same URL is *cleaner* — it strips the nav/tags/about/title and leaks only a small mobile Table-of-Contents callout (a separate, minor issue: that callout is hidden on desktop purely via a CSS class `.mobile-only { display:none }`, and Readability's visibility filter only checks inline `style`/`hidden`/`aria-hidden`, not stylesheets, so our linkedom-parsed DOM — which applies no CSS — keeps it).

## Options

[...]

PR #1385, opened 2026-07-19
Markdown: GFM footnotes + MathML math rendering

brendanlong:
## Summary

Extends our Markdown rendering (the shared `processMarkdown`, used by uploads, saved articles, and the markdown-only full-content fallback) with two features that core `marked` doesn't support. Follows up on the #1280 work now that we prefer HTML but still handle markdown uploads / markdown-only endpoints.

### 1. GFM footnotes (`marked-footnote`)

Core marked leaves `[^1]` references and `[^1]:` definitions as literal text, with definitions dumped inline where written. Now references render as superscript anchors and definitions collect into a `<section class="footnotes">` block. Registered on a dedicated `Marked` instance so it doesn't leak into the other `marked` importers (summarization, GitHub plugin). Verified end-to-end through the read-path sanitizer: `<sup>` markers, `id` anchors, and `#fragment` back-links all survive (`id`/`data-*` allow-listed, same-document fragments preserved).

### 2. Math rendering (`marked-katex-extension`, MathML output)

Renders `$…$` / `$$…$$` TeX to presentation **MathML** — native browser rendering, no client JS/CSS — matching how the sanitizer already handles feed math (MathJax→MathML).

**Sanitizer change (security-critical):** KaTeX wraps output in `<semantics>` alongside an `<annotation encoding="application/x-tex">` holding the raw TeX. The sanitizer already unwraps `<semantics>` (presentation MathML renders) but was *unwrapping* the annotation too, spilling raw TeX as visible text next to every equation. Fix: add `annotation`/`annotation-xml` to the **drop-with-content** set. This is a safe-direction change (drops more) that also:
- **fixes a latent bug**: untrusted feed MathML carrying an `<annotation>` already leaked visible TeX in the reader, and
- **hardens the `<annotation-xml>` mXSS vector**: it now drops the subtree instead of unwrapping (keeping children).

Neither annotation is used for visual rendering or screen-reader a11y (those use the presentation MathML we keep) — dropping them only forgoes copy-as-LaTeX. If we ever want that back, allow-listing `<annotation>` (text-only, still blocking `annotation-xml`) is a separate, security-reviewed change.

**Perf:** ~18 µs/equation at markdown-process time (once per fetch/upload, **not** per read); per-read sanitize is unchanged.

## Tests

- Rust sanitizer: `<semantics>` unwrap + `<annotation>` drop; `<annotation-xml>` payload dropped with subtree.
- TS pipeline (`sanitize-entry-html`): presentation MathML kept, TeX annotation gone (no source leak); strengthened annotation-xml test to assert payload removal.
- `processMarkdown`: footnote rendering/numbering, inline+display math → MathML, malformed TeX doesn't throw.
- Full suite: 2543 unit + 57 Rust pass; typecheck + lint clean. `src/server/html/CLAUDE.md` updated.


brendanlong:
### Rendered output (uploaded Markdown article, viewed in the app)

A saved article run through the real `processMarkdown` pipeline and rendered via the read-path sanitizer. Shows inline math (`μ ± 2σ`, the tail integral, scaling-law exponents), a display equation with under-braces (bias²/variance), superscript footnote references, and the footnotes section with back-links — including **math inside footnotes** (`α ≈ 0.076`, `∫ e^{-x²} dx = √π`). No raw TeX source leaks anywhere.

Markdown footnotes + MathML rendering

_The screenshot lives on a throwaway `pr-1385-screenshots` branch (not in this PR's diff or master)._

— Claude

Issue #1401, opened 2026-07-19
Investigate smarter heuristics for saved-article excerpts (what does Google do?)


brendanlong:
## Motivation

We currently derive a saved-article excerpt by preferring the extracted description/excerpt (usually `og:description` / `<meta name="description">`) when it's substantial, falling back to the article body text (`summarizeCleanedContent` in `src/server/html/strip-html.ts`).

The description preference is usually the better choice, but it fails when a site uses a **single site-wide description** for every page. Example: https://www.brethorsting.com/blog/2026/05/domain-expertise-has-always-been-the-real-moat/ shows "Personal website for Aaron Brethorst - Seattleite, technology leader, photographer,…" as the excerpt — the author's site-wide bio, not the article.

We're treating this as a **site problem** for now (a bad/site-wide meta description), but it's worth investigating heuristics to detect and avoid it.

## Things to investigate

- **What do others do?** Google's search snippets famously *don't* just use `<meta name="description">` — they synthesize a snippet from page content relevant to the query (or fall back to the meta description only when it's a good page-level summary). Readability-style tools, Wallabag, Miniflux, and browser reading modes each have their own excerpt strategies worth comparing.
- **Detect a site-wide/boilerplate description**: e.g. the description is identical across multiple entries from the same feed/host, or it matches the site name / feed description, or it doesn't overlap at all with the article body → prefer body text.
- **Prefer per-article signals**: distinguish `og:description` (often per-article) from a bare `<meta name="description">` (often site-wide). This needs the extractor to report *which* meta tag it used — dom_smoothie/Readability doesn't currently expose that.
- **First-meaningful-paragraph extraction**: skip author/date/byline lines and lead with the first real sentence of the article, rather than a hard truncation of `textContent`.
- Possibly an LLM-generated summary (we already summarize entries via `summarization.ts`) as an opt-in for saved articles.

## Context

- Shared excerpt helper: `summarizeCleanedContent` (`src/server/html/strip-html.ts`).
- Related: the excerpt-pipeline dedup (#1400) and arXiv API abstract (#1399).

Reported by Claude.

brendanlong:
LLM summaries are too expensive/slow for this, and we probably don't want any approach that requires looking at multiple pages, since our save flow only has the contents of one page.

Issue #1400, opened 2026-07-19
Saved-article excerpt shows site-wide meta description; unify upload/saved/full-content excerpt paths


brendanlong:
## Bug

Some saved articles show a **site-wide description** as their excerpt instead of the article text. Example: https://www.brethorsting.com/blog/2026/05/domain-expertise-has-always-been-the-real-moat/ shows:

> "Personal website for Aaron Brethorst - Seattleite, technology leader, photographer,…"

That string is the page's `<meta name="description">` — a site-wide bio, not an article summary. The page has no per-article `og:description`, so Readability falls back to the site description, and the excerpt logic prefers `cleaned.excerpt` (the meta description) over the article body text.

This is distinct from the arXiv table-of-contents bug fixed in PR #1398 — here it's a non-plugin page and Readability ran fine; the problem is preferring the meta description.

## Affected code (duplicated + inconsistent)

Each surface hand-rolls its own cleaned-content excerpt logic and they disagree:

| Surface | Code | Prefers meta-description excerpt? |
|---|---|---|
| Saved articles (cleaned path) | `saved.ts` (`computeSavedArticleExcerpt`, `cleaned.excerpt \|\| textContent`) | **Yes** ← bug |
| Upload: HTML | `process-upload.ts` (`cleaned.excerpt \|\| generateSummary(...)`) | **Yes** ← same bug |
| Upload: docx | `process-upload.ts` (`generateSummary(cleaned.content)`) | No (body text) |
| Email ingest | `process-inbound.ts` (`generateSummary(contentCleaned ?? …)`) | No (body text) |
| Feed entries | `content-utils.ts` (`generateSummary(content)`) | No (body text) |
| `generateCleanedSummary` (intended shared helper) | prefers `cleaned.excerpt` if `>=50` | **Dead code** — only referenced by its own unit test |

## Proposed direction

Two entangled pieces:

1. **Excerpt policy**: prefer the extracted **article body text** over the meta description for the cleaned path, matching what feed entries / email / docx already do. The failure mode of the meta-description path (site tagline) is worse than a truncated first paragraph. (Alternative: only use `cleaned.excerpt` when it came from a per-article `og:description`, but Readability doesn't expose which meta tag it used.)
[...]

PR #1402, opened 2026-07-19
Refactor: unify cleaned-content excerpt generation into summarizeCleanedContent

brendanlong:
## Motivation

Follow-up to #1398 / #1400. Three surfaces that run Readability and then need a preview each hand-rolled their own "cleaned content → excerpt" logic, and they had drifted apart:

| Surface | Old logic | Prefers description? |
|---|---|---|
| Saved articles (`saved-excerpt.ts`) | `cleaned.excerpt \|\| textContent.slice(0,300)` | yes (any non-empty) |
| Upload: HTML (`process-upload.ts`) | `cleaned.excerpt \|\| generateSummary(cleaned.content)` | yes (any non-empty) |
| Upload: docx (`process-upload.ts`) | `generateSummary(cleaned.content)` | **no** (body text) |
| `generateCleanedSummary` (`content-cleaner.ts`) | `cleaned.excerpt` if ≥50 else `truncateText(textContent)` | yes — but **dead code** (only its own test referenced it) |

## Change

- Add one canonical, description-preferring helper `summarizeCleanedContent` to the pure `strip-html.ts` module (no DB/native, so it's directly unit-testable): prefer the extracted description when substantial (≥50 chars), else the article body text, each truncated at a word boundary.
- Wire `saveArticle` (via `computeSavedArticleExcerpt`) and both upload paths to it. Behavior is now identical everywhere.
- Delete the orphaned `generateCleanedSummary` and its private `truncateText`.

## Policy decision

We deliberately **keep** the description preference — it's usually a better summary than a truncated first paragraph. The known failure mode (a site that serves one site-wide `<meta name="description">` for every page, e.g. the brethorsting.com example in #1400) is treated as a site issue rather than worked around here. **#1401** tracks investigating smarter heuristics (what does Google do?) for a future improvement.

Net behavior changes from consolidation (all minor, all toward consistency):
- docx uploads now also prefer the description (previously body-text only).
- Saved articles / HTML uploads now apply the ≥50-char guard (a tiny junk description no longer wins) and truncate the body-text fallback at a word boundary with an ellipsis instead of a hard 300-char slice.

## Tests

- `summarizeCleanedContent` gets its own unit coverage in `entry-summary.test.ts` (the old `generateCleanedSummary` cases moved here).
- `saved-article-excerpt.test.ts` updated to assert delegation + the ≥50 threshold.
- Full unit suite (2555 tests) + typecheck + lint + knip clean.

Part of #1400.

### DIFF
--- a/src/server/feed/content-cleaner.ts
+++ b/src/server/feed/content-cleaner.ts
@@ -467,47 +467,6 @@ function finishCleaned(
   };
 }
 
-/**
- * Generates a summary from cleaned content.
- *
- * Uses the excerpt if available, otherwise truncates the text content.
- *
- * @param cleaned - The cleaned content result
- * @param maxLength - Maximum summary length (default: 300)
- * @returns Summary string
- */
-export function generateCleanedSummary(cleaned: CleanedContent, maxLength = 300): string {
-  // Prefer the excerpt if it's meaningful
-  if (cleaned.excerpt && cleaned.excerpt.length >= 50) {
-    return truncateText(cleaned.excerpt, maxLength);
-  }
-
-  // Fall back to truncated text content
-  return truncateText(cleaned.textContent, maxLength);
-}
-
-/**
- * Truncates text to a maximum length, adding ellipsis if needed.
- * Attempts to break at word boundaries.
- */
-function truncateText(text: string, maxLength: number): string {
-  const trimmed = text.trim();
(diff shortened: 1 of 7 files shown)


brendanlong:
Superseded by #1403, which includes this `summarizeCleanedContent` consolidation as part of the larger save/upload unification. Closing in favor of that PR.

— Claude

PR #1411, opened 2026-07-21
Build arXiv saved-article summary + authors from the arXiv API (#1399)

### LINKED ISSUE #1399: Build arXiv saved-article summary + authors from the arXiv API
## Idea

When saving an arXiv paper, build the excerpt (and ideally title/authors) from the **arXiv API abstract** rather than scraping whatever Readability extracts from the HTML render.

The arXiv Atom API returns the abstract directly as `<summary>` plus structured `<author><name>` entries and the title:

```
http://export.arxiv.org/api/query?id_list=2503.11926
```

For https://arxiv.org/abs/2503.11926 the `<summary>` is the real abstract:

> "Mitigating reward hacking—where AI systems misbehave due to flaws or misspecifications in their learning objectives—remains a key challenge in constructing capable and aligned models. We show that we can monitor a frontier reasoning model, such as OpenAI o3-mini, for reward hacking…"

That is a far better excerpt than the first 300 chars of the extracted body, and the `<author>` list is cleaner than Readability's byline extraction.

## Context

- Related fix (plugin ordering so the excerpt comes from cleaned content, not the raw ToC): PR #1398.
- Today the arXiv plugin (`src/server/plugins/arxiv.ts`) only does a HEAD check to choose `/html/` vs `/abs/` (`src/server/feed/arxiv.ts`); it never calls the API.

## Sketch of work

- Add an arXiv API fetch (Atom, parse `summary` / `title` / `author` — extract paper id via existing `extractPaperId`).
- Thread an optional `excerpt`/`summary` (and authors) through `SavedArticleContent` → `pluginContent` → the excerpt/metadata logic in `saveArticle`, since plugin content currently carries no excerpt field.
- Use the API abstract as the excerpt when available, still falling back to Readability content if the API call fails.
- All outbound fetches must go through `fetchWithSsrfProtection` + our `USER_AGENT`, and honor arXiv's rate limits.

Reported by Claude.

brendanlong:
Closes #1399.

When saving an arXiv paper, we now fetch the paper's **structured metadata from the arXiv Atom API** (`export.arxiv.org/api/query?id_list=<id>`) and prefer it over Readability's scrape of the HTML render.

## What changed

- **New arXiv API client** (`src/server/feed/arxiv.ts`):
  - `fetchArxivMetadata(paperId)` — one request through `fetchWithSsrfProtection` + our `USER_AGENT`, size-capped, returns `null` on any failure.
  - `parseArxivApiResponse(xml)` — SAX (xmlMode) parse of the first `<entry>`, extracting `title`, `<summary>` (the real abstract), and the `<author><name>` list. Ignores the feed-level query-echo `<title>`.
  - `formatArxivAuthors(authors)` — collapses to `"First Author et al."` for 3+ authors (we store a single author string), `"A and B"` for two.
- **arXiv plugin** (`src/server/plugins/arxiv.ts`): fetches the HTML render and the API metadata **concurrently** (different hosts), then returns the API `title` / `author` / `excerpt`. Each falls back to `null`, so Readability/OG metadata still fill them if the API call fails.
- **Threaded an `excerpt` field** through `SavedArticleContent` → `pluginContent` → `computeSavedArticleExcerpt`, where an explicit plugin excerpt now **outranks everything** (including Readability) and is clipped to the 300-char summary length. This is the field the earlier arXiv-ToC excerpt fix (#1398) couldn't reach.

## Result

For `https://arxiv.org/abs/2503.11926`, the saved article gets:
- **Title:** "Monitoring Reasoning Models for Misbehavior and the Risks of Promoting Obfuscation"
- **Author:** "Bowen Baker et al." (9 authors)
- **Excerpt:** the actual abstract ("Mitigating reward hacking…"), instead of the first 300 chars of the extracted body.

## Testing

- New unit tests for `parseArxivApiResponse` (real-format XML, multi-entry, empty), `formatArxivAuthors`, and the plugin-excerpt precedence/clipping in `computeSavedArticleExcerpt`.
- Verified the parser against a live API response for 2503.11926.
- `pnpm typecheck`, `pnpm lint`, and `pnpm test:unit` all pass.

### DIFF
--- a/src/server/feed/arxiv.ts
+++ b/src/server/feed/arxiv.ts
@@ -10,8 +10,9 @@
  * when the HTML version is available, providing a better reading experience.
  */
 
+import { Parser } from "htmlparser2";
 import { logger } from "@/lib/logger";
-import { FEED_FETCH_TIMEOUT_MS } from "@/server/http/fetch";
+import { FEED_FETCH_TIMEOUT_MS, readResponseWithSizeLimit } from "@/server/http/fetch";
 import { fetchWithSsrfProtection } from "@/server/http/ssrf";
 import { USER_AGENT } from "@/server/http/user-agent";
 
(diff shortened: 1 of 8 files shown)


brendanlong:
Ran an independent code review on this branch. No security or correctness defects found; the outbound fetch is SSRF-protected with our UA, and the abstract/title/author render as escaped text (no HTML injection path).

One functional point from the review was worth a real-world check: old-format IDs get percent-encoded (`encodeURIComponent("hep-th/9901001")` → `hep-th%2F9901001`). I verified the arXiv API accepts the encoded slash and returns the correct paper (`hep-th/9901001` → "String Junctions and Their Duals in Heterotic String Theory"), so the injection-safe encoding is also functionally correct for old-format papers.

Minor notes left as-is (all low/nit): the refetch quality guard compares body-text length only, so a refetch while the arXiv API is transiently down could downgrade title/author/excerpt to Readability's scrape until the next refetch — matches existing behavior for those fields and self-heals. Also note the version suffix (`2601.04649v1`) is preserved and pins that specific version in the API request, which is the intended behavior.

— Claude

brendanlong:
Follow-up from the precedence discussion: pushed a commit that ranks **Readability's byline above the raw OG/`meta[name=author]` scrape**, so `author` now matches `title` (Readability above the raw metadata scrape). Rationale: Readability is itself a source-specific extractor reading mostly the same OG/meta tags, and when they disagree we trust it to have done better. It's HTML-only in practice — Readability is skipped for Markdown (frontmatter wins) and skipReadability plugins (`cleaned` is null), so those still prefer their own declared metadata. Existing saved-article integration tests pass unchanged.

Also filed #1413 to add caller-supplied `author` and `summary`/`excerpt` hints to `saveArticle` (mainly for MCP `save_article`, also worth exposing on the save API) — the top-precedence override that only `title` has today.

— Claude

PR #1412, opened 2026-07-21
Improve docx upload extraction: read core.xml metadata, skip Readability (#1404)

### LINKED ISSUE #1404: Investigate docx upload extraction quality (Readability tags / better title/author/description)
## Context

Uploaded `.docx` files are converted to HTML by mammoth and then run through Readability in the shared `buildArticleFields` pipeline (see #1403). This is an extreme edge case (few users upload docx), so it wasn't touched in that refactor, but it's worth a look.

## Things to check

- **Does mammoth→Readability produce garbage?** mammoth output is a flat run of `<p>`/`<h1>` with no site chrome, so Readability may mis-detect the title (e.g. promote the first paragraph), pick a poor byline, or extract a weird excerpt. Compare stored title/author/summary against the source document.
- **Better metadata sources for docx**: `.docx` has real document properties (`core.xml`: `dc:title`, `dc:creator`, `dc:description`). We could read those directly (mammoth doesn't expose them; would need to unzip the docx and parse `docProps/core.xml`) instead of guessing from the rendered HTML — likely far more accurate for title/author/description.
- **Should docx skip Readability entirely?** Like Markdown, docx is already "clean" content (no nav/ads), so running Readability mostly risks over-stripping. Consider treating docx like Markdown (skip Readability, use body as-is) plus core.xml metadata.

## Related

- #1403 (the save/upload unification that routes docx through buildArticleFields)

Reported by Claude.

brendanlong:
Fixes #1404.

## Problem

Uploaded `.docx` files were converted by mammoth and then run through Readability in the shared `buildArticleFields` pipeline (#1403). But mammoth emits a flat, chrome-free run of `<p>`/`<h1>`, so Readability adds no cleaning and only risks mis-detecting the title (promoting the first paragraph), picking a poor byline, or extracting a weird excerpt. And it guessed title/author/excerpt from that rendered HTML instead of the metadata the author actually set in the document.

## Changes

- **Read the real document properties** from `docProps/core.xml` (`dc:title`, `dc:creator`, `dc:description`). A new dependency-free reader (`src/server/file/docx-core-props.ts`) walks the ZIP central directory and inflates **only** that one tiny entry — no full unzip, addressing the efficiency concern raised in the issue. It's fully bounds-checked and fail-safe: a malformed/garbage upload just yields empty properties (we fall back to the filename / mammoth body) rather than throwing.
- **Skip Readability for docx**, treating it like Markdown: store the mammoth body as-is and use core.xml for title/author/summary. The filename remains the last-resort title when core.xml has none (unchanged behavior for property-less docs).
- **Generalize the pipeline field**: the markdown-only `markdownResult` bundle field is renamed to `preCleanedContent`, since it now carries both Markdown frontmatter and docx core.xml — both "already-clean" sources share the one skip-Readability path (DRY).

## Precedence (docx)

- title: caller-provided → `dc:title` → filename
- author: `dc:creator`
- summary: `dc:description` → generated from body

## Testing

- **Unit** (`tests/unit/docx-core-props.test.ts`): the core.xml reader — entity decoding, whitespace trimming, missing props, no core.xml, and garbage/empty buffers (never throws). Uses a new in-memory docx/zip builder (`tests/utils/docx.ts`) so no binary fixture is needed.
- **Integration** (`tests/integration/saved-uploads.test.ts`): the full docx upload → save path — core.xml title/author/summary, Readability skipped (mammoth body preserved), filename fallback, and caller-title precedence.
- `pnpm typecheck`, `pnpm lint`, unit + integration suites all pass.

## Not in scope

The Google Drive `.docx` path (`src/server/google/drive.ts`) is a separate flow (Drive API → plugin) and still uses the Drive filename for the title; plumbing core.xml through there could be a follow-up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "htmlparser2": "^10.1.0",
     "ioredis": "^5.11.1",
     "jose": "6.2.3",
+    "jszip": "^3.10.1",
     "katex": "^0.17.0",
     "linkedom": "^0.18.13",
     "mammoth": "^1.12.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -97,6 +97,9 @@ importers:
       jose:
         specifier: 6.2.3
         version: 6.2.3
+      jszip:
+        specifier: ^3.10.1
+        version: 3.10.1
       katex:
         specifier: ^0.17.0
         version: 0.17.0
--- a/src/server/file/docx-core-props.ts
+++ b/src/server/file/docx-core-props.ts
@@ -1,0 +1,115 @@
+/**
+ * Reads a `.docx`'s real document properties from `docProps/core.xml`.
+ *
+ * A `.docx` is a ZIP whose `docProps/core.xml` holds Dublin Core metadata the
+ * author actually set (`dc:title`, `dc:creator`, `dc:description`) — far more
+ * reliable than guessing title/author/excerpt from the flat run of `<p>` mammoth
+ * renders (mammoth doesn't expose these, and Readability tends to promote the
+ * first paragraph as a title on chrome-free content). See issue #1404.
+ *
+ * We read the ZIP with **jszip** — the same library mammoth already loads to
+ * parse this exact buffer on every upload — so this adds no new parser or attack
+ * surface. `loadAsync` is lazy: it parses the central directory but only inflates
+ * an entry when you read it, so we decompress just the tiny `core.xml` (never the
+ * large `word/document.xml`). Fully wrapped so a malformed upload just yields
+ * empty properties (we fall back to the filename / mammoth body) rather than
+ * throwing.
+ *
+ * No decompression cap: mammoth already inflates the larger `document.xml` from
+ * the same (size-limited) upload with no cap via jszip/pako, so a `core.xml`-only
+ * cap would be theater — zip-bomb hardening is a pipeline/upstream concern, not
+ * something this reader can meaningfully add.
+ */
+
+import JSZip from "jszip";
+import { Parser } from "htmlparser2";
+import { logger } from "@/lib/logger";
+
+export interface DocxCoreProperties {
+  /** `dc:title` — the author-set document title, if any. */
+  title: string | null;
+  /** `dc:creator` — the author. */
+  author: string | null;
+  /** `dc:description` — a short summary/abstract. */
+  description: string | null;
+}
+
+const EMPTY_PROPERTIES: DocxCoreProperties = { title: null, author: null, description: null };
+
+const CORE_PROPS_PATH = "docProps/core.xml";
+
+/**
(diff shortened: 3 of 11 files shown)


brendanlong:
Ran an independent code review of the diff. One real robustness issue found and fixed; a couple of minor edge cases judged acceptable.

**Fixed (commit 517c7ebe):** decompression-bomb / memory-amplification DoS. `inflateRawSync` had no output cap — `compressedSize` is bounded by the (size-limited) upload, but the *decompressed* size wasn't, so a crafted highly-compressible `core.xml` inside an otherwise-valid `.docx` could expand ~1000x and force a multi-GB allocation before zlib gave up. Now capped via `maxOutputLength: 4 MiB` (far above any real `core.xml`); the resulting `ERR_BUFFER_TOO_LARGE` is caught and treated as "no core.xml metadata". Added tests for the cap plus the previously-untested stored (method 0) ZIP branch.

**Considered, left as-is (both real-world safe):**
- The SAX match keys on the literal `dc:`/`dcterms:` tag prefixes rather than resolving the Dublin Core namespace URI. Every Office/LibreOffice producer uses the conventional prefixes, and it's a fail-safe fallback either way.
- A nested child element inside a `dc:*` tag would drop the value — `core.xml` never nests these, so not reachable in practice.

The ZIP central-directory walk itself was verified safe: `entryCount` is a UInt16, `ptr` strictly advances ≥46 bytes/iteration (no infinite loop), every read is bounds-checked before access, and ZIP64/unknown-method entries bail to empty. All failure modes are fail-safe.

— Claude

brendanlong:
Reworked the core.xml reading based on review discussion.

**Replaced the hand-rolled ZIP reader with jszip** (commit 5e9a83df). Rationale: mammoth already loads jszip and already fully parses this exact `.docx` buffer on every upload, so reading one more entry adds no new parser and no new attack surface — it just reuses the library that's already handling this input. jszip is promoted from a transitive dep of mammoth to an explicit direct dependency (deduped to the same 3.10.1). `loadAsync` is lazy, so only the tiny `core.xml` is inflated, never `word/document.xml`. This deletes ~220 lines of custom untrusted-binary parsing.

**Dropped the core.xml decompression cap.** mammoth already inflates the *larger* `document.xml` from the same size-limited upload with no cap (via jszip/pako), so a `core.xml`-only cap was security theater — it protected the small file while the big one stayed open. Real zip-bomb hardening is a pipeline/upstream concern (jszip/pako) that we can't meaningfully close without replacing mammoth, and that isn't worth it for this edge-case feature. The 5 MB upload limit bounds the worst case.

Also evaluated whether to replace mammoth itself — it remains the standard for `.docx → semantic HTML` in Node; the "alternatives" are all reverse-direction (HTML→docx), generation/templating, style-faithful viewers, or heavy external binaries (pandoc/LibreOffice). No modern replacement fits this job.

— Claude

Issue #1282, opened 2026-07-16
Drop the persisted sanitized-content columns: sanitize per read with the native sanitizer


brendanlong:
## Context

Phase 2 of the Rust sanitizer plan (phase 1: #1281). With the native sanitizer, sanitization is fast enough to run on every read instead of being persisted:

| body | old (sanitize-html) | native (lol_html) |
|------|--------------------|--------------------|
| 10 KB | 1.3 ms | 0.09 ms |
| 100 KB | 6.6 ms | 0.9 ms |
| 700 KB | 53 ms | 6.2 ms |

Worst realistic read (Google Reader stream-contents sync, ~100×10KB bodies) is ~10ms of native CPU, off the event loop via the async form. A single 700KB math-heavy body is ~6ms + the (now also native) MathJax conversion.

## What gets deleted

- `entries.content_*_sanitized` / `full_content_*_sanitized` columns + `*_sanitized_version` stamps (roughly halves entry-content storage)
- `withSanitizedEntryContent` / `withSanitizedEntryContentAsync` derivation at every write site (`sanitize-entry.ts`)
- The read-path self-heal (`resolveSanitizedFamily` heal branch), `persistResanitizedFamily` CAS, `isSanitizedFamilyStale`, `shouldMaterializeFullContentOriginal` lazy-materialization rule
- `scripts/resanitize-bulk.ts`, `selectStaleEntriesForResanitize`, `RESANITIZE_STALENESS_KEY`, the `idx_entries_resanitize` index and its EXPLAIN test
- The worker-pool `sanitizerVersion` probe machinery (`getVerifiedPool` version checks, `getWorkerSanitizerVersion`) — already vestigial after #1281
- The "fast-forward" analysis burden on every `SANITIZER_VERSION` bump — a bump becomes just a deploy

The read path becomes: `SELECT raw columns → sanitizeEntryHtmlAsync → respond`.

## Design questions to settle

- **Read paths**: `resolveSanitizedFamily` collapses to a sanitize call; `getEntries` (Google Reader batch) should sanitize concurrently (async form) with a small concurrency bound.
- **Summaries**: AI summaries are sanitized at generation time and stored — decide whether those keep a stored sanitized copy (they're ours, generated once) or also go per-read.
- **Migration**: expand/contract — code stops reading the sanitized columns first, then a later migration drops them (they must survive one release for rollback per the migration-compatibility rule).
[...]

PR #1422, opened 2026-07-25
Fix broken images in saved GitHub Markdown files

brendanlong:
## The bug

Images in saved GitHub Markdown files (e.g. [Why Software Factories Fail](https://github.com/humanlayer/advanced-context-engineering-for-coding-agents/blob/main/wsff.md)) all render broken. 32 of 32 images in that post were dead.

GitHub resolves a Markdown file's relative URLs against **two different bases**:

- `src` (images) → `raw.githubusercontent.com/{owner}/{repo}/{ref}/…` — the file bytes
- `href` (links) → `github.com/{owner}/{repo}/blob/{ref}/…` — a page a reader can follow

We had no GitHub-specific resolution at all. The plugin's HTML kept its relative `images/foo.png` URLs, and the generic single-base absolutizer downstream (`buildArticleFields` in `saved.ts`) resolved them against the article URL, producing `github.com/{owner}/{repo}/blob/{ref}/images/foo.png` — a valid URL that serves `text/html`, not a PNG. Hence a broken image for every relative reference.

(Only relative references were affected; the post's absolute `img.youtube.com` thumbnails worked fine, which is why this wasn't obvious.)

## The fix

- `absolutizeUrls` takes an optional `mediaBaseUrl` for the media attributes (`src`, `poster`, `srcset`), defaulting to `baseUrl`. An explicit `<base href>` still overrides both, per the HTML spec.
- The GitHub plugin resolves each repo file with the raw base for media and the blob base for links, defaulting to the `HEAD` ref for the repo-root README (which we fetch without one).
- The absolutizing happens inside `processFileContent`, which now takes the file's repo location (or `null` for gists, whose flat file list has no relative paths to resolve), so a new URL type can't silently skip the step.

## Verification

Ran the real plugin against the real post: all 32 images now resolve, and a `HEAD` on each returns `200` with an `image/*` content type (0 failures). Relative links resolve to the blob view. Confirmed the 32 images also survive the read-path sanitizer, and CSP already allows any `https:` image.

New unit tests cover the dual-base resolution (`images/x.png`, `./images/x.png`, raw `<div>` blocks GitHub Markdown commonly uses for centered figures, nested file paths, absolute URLs, the `HEAD` default, gists, HTML files, and code blocks) plus the `mediaBaseUrl` option itself. Full unit suite passes (2588 tests); `pnpm typecheck` clean.

## Note

This fixes new saves. Articles already saved keep their broken URLs until refetched.

### DIFF
--- a/src/server/feed/content-cleaner.ts
+++ b/src/server/feed/content-cleaner.ts
@@ -27,9 +27,17 @@ const decoder = new TextDecoder();
  *
  * @param html - The HTML content to process
  * @param baseUrl - The base URL for resolving relative URLs
+ * @param options.mediaBaseUrl - Separate base for the media attributes (src,
+ *   poster, srcset); defaults to `baseUrl`. Needed for sources that serve a
+ *   document and its embedded files from different origins — see
+ *   `absolutizeGitHubUrls` in `src/server/plugins/github.ts`.
  * @returns HTML with all relative URLs converted to absolute
  */
-export function absolutizeUrls(html: string, baseUrl: string): string {
+export function absolutizeUrls(
+  html: string,
+  baseUrl: string,
+  options: { mediaBaseUrl?: string } = {}
+): string {
   try {
     let output = "";
     const rewriter = new HTMLRewriter((chunk) => {
@@ -41,6 +49,7 @@ export function absolutizeUrls(html: string, baseUrl: string): string {
     // <base> in <head> will be seen before any body elements.
     // Per the HTML spec, only the first <base> with an href is used.
     let effectiveBaseUrl = baseUrl;
+    let effectiveMediaBaseUrl = options.mediaBaseUrl ?? baseUrl;
     let baseHrefSet = false;
 
     // Check for <base> tag and use its href as the base URL
@@ -54,7 +63,10 @@ export function absolutizeUrls(html: string, baseUrl: string): string {
           // in case <base href> itself is relative
           const resolved = resolveUrl(href, baseUrl);
           if (resolved) {
+            // Per the HTML spec <base href> governs every relative URL in the
+            // document, so an explicit tag overrides mediaBaseUrl too.
             effectiveBaseUrl = resolved;
+            effectiveMediaBaseUrl = resolved;
             baseHrefSet = true;
           }
         }
@@ -66,7 +78,7 @@ export function absolutizeUrls(html: string, baseUrl: string): string {
       element(el) {
         const value = el.getAttribute("src");
         if (value) {
-          const absolute = resolveUrl(value, effectiveBaseUrl);
+          const absolute = resolveUrl(value, effectiveMediaBaseUrl);
           if (absolute && absolute !== value) {
             el.setAttribute("src", absolute);
           }
@@ -92,7 +104,7 @@ export function absolutizeUrls(html: string, baseUrl: string): string {
       element(el) {
         const value = el.getAttribute("poster");
         if (value) {
-          const absolute = resolveUrl(value, effectiveBaseUrl);
+          const absolute = resolveUrl(value, effectiveMediaBaseUrl);
           if (absolute && absolute !== value) {
             el.setAttribute("poster", absolute);
           }
@@ -104,7 +116,7 @@ export function absolutizeUrls(html: string, baseUrl: string): string {
       element(el) {
         const value = el.getAttribute("srcset");
         if (value) {
-          el.setAttribute("srcset", absolutizeSrcset(value, effectiveBaseUrl));
+          el.setAttribute("srcset", absolutizeSrcset(value, effectiveMediaBaseUrl));
         }
       },
     });
--- a/src/server/feed/content-cleaner.ts
+++ b/src/server/feed/content-cleaner.ts
@@ -27,9 +27,10 @@ const decoder = new TextDecoder();
  *
  * @param html - The HTML content to process
  * @param baseUrl - The base URL for resolving relative URLs
- * @param options.mediaBaseUrl - Separate base for the media attributes (src,
- *   poster, srcset); defaults to `baseUrl`. Needed for sources that serve a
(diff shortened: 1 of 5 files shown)


brendanlong:
---

## Review follow-up

An independent Opus reviewer went over this; no blockers. It verified against live GitHub content that the two bases here match GitHub's own renderer, and cleared the cases I'd worried about: refs containing slashes (`refs/heads/main`), percent-encoded paths, protocol-relative URLs, `www.github.com`, `srcset`, the `HEAD` default, sanitizer/CSP interaction, and origin injection (both hosts are hardcoded; `new URL()` path normalization can't change the host).

It also confirmed GitHub is genuinely the only plugin with a split media origin, so the fix didn't need to be more general — though arXiv relies on `<base href>` handling in the same function, which now has a regression test.

Changes made in response:

- Scoped the `<base href>` comment to "elements after the tag" — the single streaming pass can't retroactively rewrite earlier ones, and marked-rendered Markdown has no `<head>` to guarantee ordering.
- Renamed the concept from "media" to "embedded resources": the `[src]` selector also matches `iframe`/`source`/`embed`.
- Documented that `canonicalUrl` doubles as the relative-URL resolution base — the fact a future plugin author needs in order to avoid this bug class.
- Filed the two gaps this deliberately doesn't cover, each with a skipped test: **#1423** root-relative paths (`/docs/logo.png` — GitHub reads these as repo-root-relative; no choice of base URL can express that, so it needs a separate mechanism) and **#1424** gist sibling references. The #1423 tests are verified failing today, so un-skipping them is the acceptance criterion.
- Added plugin-level `srcset`/`poster` and `HEAD`-ref link coverage, plus the arXiv `<base href>` shape.

Note that #1423 is not a regression — that input was equally broken before this PR.


— Claude

brendanlong:
## Follow-up: one Markdown instance, plus two bugs found in review

**Consolidated the Markdown path.** Every source now shares `src/server/markdown`'s single `Marked` instance — the GitHub plugin's separate one is gone, and `summarization.ts` no longer reconfigures the `marked` global (nothing imports it any more). Added a rule to `CLAUDE.md`'s Parsing section so a second instance doesn't reappear.

My justification for splitting was wrong. I claimed KaTeX would turn `costs $5 and $10` into a math span; it doesn't — standard delimiters require a non-space before the closing `$`, so every prose case I could construct stays text (there's a test). Consolidating is a strict improvement: the humanlayer post now renders its **footnotes** (13 in-page links, up from 3; 40 ids, up from 29) and gets frontmatter stripped, and repo docs get math. A frontmatter `author`/`description` now outranks the repo owner, matching what an upload does with the same file.

**Two real bugs, both found by an Opus reviewer that verified them with a throwaway probe crate:**

- `prefix_func_iris` only matched lowercase `url(`, so `URL(#g)` kept pointing at the pre-rename id and its gradient rendered **unpainted** — precisely the failure the funcIRI handling exists to prevent. `url(` is a CSS function token and case-insensitive.
- The HTML pass wrote `href` back **entity-decoded** while `id` went back raw. lol_html's `set_attribute` escapes `"` but not `&`, so `id="a&amp;b"` and `href="#a&amp;b"` drifted apart on the first pass and peeled another layer on each subsequent one — breaking the link and making the pass non-idempotent, which per-read re-sanitizing of summaries relies on. Everything is now read and written raw, which also removes a double-decode idiom that would become a live scheme bypass if copied to a non-fragment URL attribute.

Both are now verified through the real TS pipeline, and the arXiv paper still shows 289/289 in-page links resolving, idempotent.

**Also from the review:** dropped the four `<color>`-valued attributes that can never hold a funcIRI (the doc's accuracy is load-bearing since `CLAUDE.md` points reviewers at it); dropped the per-tag guards in `namespace_idrefs` that duplicated `attr_allowed`; added loop-driven tests over both constants so a typo in an entry can't ship silently; and added the cross-pass HTML→SVG id-agreement test that was previously only asserted in a commit message.

**Reverted the `SANITIZER_VERSION` bump.** The stale module comment told me to bump on any behavior change, but that machinery went away in #1282 — `src/server/html/CLAUDE.md` already says the constant is informational and a rules change is just a deploy. I'd inherited the contradiction instead of resolving it; the comment is fixed and its accreting per-version changelog is gone. Note the constant now has no consumers at all (knip flags it) — worth deleting separately.

One reviewer finding I did **not** act on: a duplicate `id` attribute (`<p ID="url" id="app">`) leaves the second un-namespaced. Browsers keep the first, so our renderer is unaffected and it isn't exploitable; it would only matter for an external consumer of the sanitized HTML (MCP/Wallabag) with last-wins semantics. Say the word if you want it hardened.

— Claude

brendanlong:
## Deleted `SANITIZER_VERSION`

Removed from the core crate, the napi export, the hand-maintained `index.d.ts` typings, and the TS re-export. The native module's runtime surface is now exactly four functions (`sanitizeEntryHtml`, `sanitizeEntryHtmlAsync`, `embedCanonicalHostnames`, `normalizeEmbed`) — verified by requiring it. knip no longer flags it.

Two comments explained the sanitize-on-read guarantee in terms of "a `SANITIZER_VERSION` bump" (the summarization router and its integration test); both now describe it as a rules change, which is what actually happens. `src/server/html/CLAUDE.md` and `SECURITY.md` now record that there deliberatly *is* no version constant, so it doesn't get reintroduced without a consumer.

### Unrelated breakage found while doing it

`scripts/bench/seed.sql` referenced it via a `:SANVER` psql variable, and inserted into `content_original_sanitized` / `content_cleaned_sanitized` / `content_sanitized_version` — **columns #1282 dropped**. The bench seed could not have run against the current schema; it was broken before this PR, independently of the constant.

Fixed it and actually ran it: created a throwaway database, migrated, and ran `seed.sql` twice (it's documented as an idempotent re-seed) — 0 errors both times, 635k entries and 303k user_entries. Throwaway database dropped.

— Claude

Issue #1082, opened 2026-07-08
Saved articles: read-state bypass, awaited Redis publishes, unbounded upload


brendanlong:
Found in the repo-wide audit (verified in code). All in `src/server/services/saved.ts`.

### 1. Refetch path flips `read=false` outside the read-state machinery (medium, correctness)
`saved.ts:909-913` does a bare `UPDATE user_entries SET read=false`. This bypasses everything `markEntriesRead` maintains: no `read_changed_at` (breaks the changedAt idempotency contract — a stale queued offline "mark read" older than the refetch still wins later), no `updated_at` bump (so `sync.events` classifies it metadata-only and offline clients never learn it went unread), and no `entry_state_changed`/counts publish (other tabs' unread badges go stale). It's also a second statement outside the transaction with the content update at `saved.ts:907` — a crash leaves new content with old read state.
- Fix: route the unread flip through `markEntriesRead` (or at minimum set `readChangedAt`/`updatedAt` and publish state+counts), in one transaction with the content update. Share code with the read-state machinery if feasible.

### 2. Saved mutations `await` Redis publishes (medium, error handling)
`saved.ts:328-334` and `:922-930` `await publishNewEntry(...)` / `publishEntryUpdatedFromEntry(...)`. With Redis configured but down, `client.publish` rejects, so `saveArticle` throws a 500 **after** the entry transaction already committed — the client sees an error for a save that succeeded (and may retry). Every other publish site is fire-and-forget with `.catch()` (see `entry-events.ts:31-43`).
- Fix: `void publish...().catch(() => {})`.

### 3. `saved.uploadFile` has no input size cap (medium, DoS)
`saved.save`'s `html` is capped in the schema, but `uploadFile.content` is unbounded base64 (`src/server/trpc/routers/saved.ts:226`, `z.string().min(1)`), and the size check happens only **after** full decode + mammoth/Readability/marked conversion (`saved.ts:1063-1066`). An authenticated user can submit an arbitrarily large `.docx` and burn memory/CPU first.
- Fix: cap `content` in the schema (`.max(ceil(maxSavedArticleSizeBytes*4/3)+slack)`) and check the decoded buffer length before processing.

PR #1426, opened 2026-07-25
Resolve root-relative URLs in GitHub files against the repo root (#1423)

### LINKED ISSUE #1423: Root-relative image paths in saved GitHub Markdown resolve to the wrong URL
Follow-up to #1422, which fixed *directory*-relative images in saved GitHub Markdown (`images/foo.png`, `./images/foo.png`) by resolving `src` against `raw.githubusercontent.com`. Root-relative paths are still broken.

## Problem

GitHub reads a leading-slash path in a Markdown file as **repo-root**-relative. We resolve it with ordinary URL semantics, which sends it to the **origin** root:

| in the Markdown | GitHub renders | we produce |
|---|---|---|
| `` | `raw.githubusercontent.com/{owner}/{repo}/{ref}/Docs/Logo.png` | `raw.githubusercontent.com/Docs/Logo.png` (404) |

This isn't a regression from #1422 — before it, the same input produced `github.com/Docs/Logo.png`, equally dead — but it's the same bug class and it's common. `acidanthera/OpenCorePkg`'s README is a real example (``), and GitHub code search returns ~1800 files for `"src=\"/docs/" filename:README.md` alone.

## Why the current fix can't cover it

`absolutizeUrls`' `mediaBaseUrl` is a base URL, and `new URL("/x", base)` always jumps to the base's origin root. No choice of base makes `/Docs/Logo.png` land under `/{owner}/{repo}/{ref}/`. It needs either a leading-slash pre-rewrite in the GitHub plugin, or a separate "root-relative base" concept in `absolutizeUrls`.

The same question applies to root-relative **links** (`[docs](/docs/x.md)`) — GitHub also treats those as repo-root-relative, so a fix should cover `href` too, using the blob base.

## Tests

`tests/unit/github-plugin.test.ts` has two `.skip`'d tests describing the intended behavior (search for #1423); un-skip them with the fix.

Reported by Claude.

brendanlong:
Fixes #1423 (follow-up to #1422).

## Problem

GitHub reads a leading slash in a repo file as **repo-root**-relative; ordinary URL resolution sends it to the **origin** root:

| in the Markdown | GitHub renders | we produced |
|---|---|---|
| `` | `raw.githubusercontent.com/{owner}/{repo}/{ref}/Docs/Logo.png` | `raw.githubusercontent.com/Docs/Logo.png` (404) |

`mediaBaseUrl` couldn't fix this — `new URL("/x", base)` always jumps to the base's origin root, so no choice of base works.

## Fix

`absolutizeUrls` gains a `rootBaseUrl` option (plus `mediaRootBaseUrl`, defaulting to it, mirroring `mediaBaseUrl`): a directory URL that root-relative paths resolve against instead of the origin root. Internally the two bases became a small `ResolveBase { url, root }` pair threaded through `resolveUrl`/`absolutizeSrcset`, so `src`/`href`/`poster`/`srcset` all get consistent treatment in the same single streaming pass — no extra rewrite pass.

Behavior preserved:
- Protocol-relative `//host/path` is left to ordinary resolution (it's not root-relative).
- An explicit `<base href>` restores plain HTML semantics (drops the root bases), same as it already overrides `mediaBaseUrl`.
- A `rootBaseUrl` without a trailing slash is treated as a directory.

The GitHub plugin passes the blob root for `href` and the raw root for `src`, both at the file's ref, so root-relative **links** are covered too (GitHub treats those as repo-root-relative as well).

## Testing

- Un-skipped the two `#1423` tests in `tests/unit/github-plugin.test.ts`, plus new cases for a file in a subdirectory and protocol-relative URLs.
- New `rootBaseUrl` describe block in `tests/unit/content-cleaner.test.ts` (all four attribute kinds, directory-relative paths unaffected, `mediaRootBaseUrl` default, missing trailing slash, protocol-relative, `<base href>` override).
- `pnpm test:unit` (2601 passed), `pnpm typecheck`, `pnpm lint` clean.
- End-to-end against the real README from the issue (`acidanthera/OpenCorePkg`): its `/Docs/Logos/OpenCore_with_text_Small.png` now resolves to `https://raw.githubusercontent.com/acidanthera/OpenCorePkg/master/Docs/Logos/OpenCore_with_text_Small.png`, which returns **200**.

### DIFF
--- a/src/server/feed/content-cleaner.ts
+++ b/src/server/feed/content-cleaner.ts
@@ -27,24 +27,28 @@ const decoder = new TextDecoder();
  *
  * @param html - The HTML content to process
  * @param baseUrl - The base URL for resolving relative URLs
- * @param options.mediaBaseUrl - Separate base for the embedded-resource
- *   attributes (`src`, `poster`, `srcset` — so `img`/`video`/`audio`, but also
- *   `iframe`/`source`/`embed`); defaults to `baseUrl`. Needed for sources that
- *   serve a document and the files it embeds from different origins — see
- *   `absolutizeGitHubUrls` in `src/server/plugins/github.ts`.
  * @param options.rootBaseUrl - Directory URL that root-relative paths
  *   (`/a/b.png`) resolve against, instead of the origin root that ordinary URL
  *   semantics would give. For sources that serve a document tree under a path
  *   prefix and read a leading slash as relative to that tree's root rather than
  *   the origin's (a GitHub repo — see `absolutizeGitHubUrls`).
- * @param options.mediaRootBaseUrl - Root base for the embedded-resource
- *   attributes; defaults to `rootBaseUrl`.
+ * @param options.media - Separate bases for the embedded-resource attributes
+ *   (`src`, `poster`, `srcset` — so `img`/`video`/`audio`, but also
+ *   `iframe`/`source`/`embed`); the whole pair defaults to the document's.
+ *   Needed for sources that serve a document and the files it embeds from
+ *   different origins — see `absolutizeGitHubUrls` in
+ *   `src/server/plugins/github.ts`. The two bases travel together so a caller
+ *   can't point media at another origin and leave its root base behind on this
+ *   one, which is the bug this option exists to prevent.
  * @returns HTML with all relative URLs converted to absolute
  */
 export function absolutizeUrls(
   html: string,
   baseUrl: string,
-  options: { mediaBaseUrl?: string; rootBaseUrl?: string; mediaRootBaseUrl?: string } = {}
+  options: {
+    rootBaseUrl?: string;
+    media?: { baseUrl: string; rootBaseUrl?: string };
+  } = {}
 ): string {
   try {
     let output = "";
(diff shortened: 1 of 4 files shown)


brendanlong:
Pushed 9257239d addressing review findings.

**Multi-segment refs (the substantive one).** `parseGitHubUrl` took the first segment as the ref and relied on the ambiguity cancelling out in `${ref}/${path}` re-joins — which the new root bases broke, since they use the ref *alone*. So `raw.githubusercontent.com/o/r/refs/heads/main/README.md` (what GitHub's "Raw" button emits today, and a live 200) resolved `/Docs/Logo.png` to `.../o/r/refs/Docs/Logo.png`. Now `refs/heads/…` and `refs/tags/…` — the unambiguous forms — are absorbed into the ref. I verified all three consumers accept the qualified form: raw (`image/png`), blob (200), and the Contents API (200, where the old mis-split 404'd and silently fell back to raw). A branch like `feature/x` remains ambiguous and unfixable without an API round-trip; it's now called out on the helper.

**Half-configured media bases.** Rather than document the footgun, the four flat bases became `media: { baseUrl, rootBaseUrl }` — the pair travels together, so pointing media at another origin while leaving its root base on the document's (exactly the #1422 bug) isn't expressible. Test added for it.

**Leading whitespace.** The `/`-prefix and scheme checks now classify the trimmed value, matching what the URL parser does with surrounding whitespace. I deliberately left the leading-backslash form alone: browsers normalize `\` to `/`, but GitHub's server-side rewriter doesn't treat `\Docs\Logo.png` as repo-root-relative, and matching GitHub is the goal here. Stale JSDoc fixed.

**Verification.** `pnpm test:unit` (2607), typecheck, lint clean; full integration suite passes locally 3/3. Re-ran the real `acidanthera/OpenCorePkg` README through both ref shapes — the root-relative logo resolves to a URL serving `content-type: image/png` in each case. Good call on asserting content-type rather than status: a 200 alone proves nothing here, since the blob page also 200s as `text/html`.

**Unrelated CI failure.** The one red check was `entry-state-events.test.ts > "does not publish when an idempotent replay changes nothing"`. It's a pre-existing timing flake, not this change: the test's setup call mutates through the service, whose SSE publish is fire-and-forget (`void publishEntryStateChanged`, #1082), so that event can land *after* the test subscribes and get attributed to the replay. Filed as #1427 rather than papering over it with a longer quiet window.

Claude

Issue #1424, opened 2026-07-25
Relative sibling-file references in saved gists aren't resolved


brendanlong:
Noticed while fixing #1422 (relative image URLs in saved GitHub repo files).

A gist's files are a flat list, but a Markdown file in one can still reference a sibling — `![chart](chart.png)`. `processFileContent` passes `location: null` for gist files, so no GitHub-specific resolution happens and the reference falls through to the generic single-base absolutizer in `saved.ts`, which resolves it against `gist.github.com/{user}/{id}` — an HTML page, so the image renders broken (the same failure mode #1422 fixed for repos).

The working form is `https://gist.githubusercontent.com/{user}/{id}/raw/{filename}`.

Low priority: sibling image references in gists are much rarer than in repo READMEs, and a gist can't hold a real image file anyway (gists are text-only), so in practice this only affects a gist referencing a name that doesn't exist. Filing so the `location: null` decision in `src/server/plugins/github.ts` is a recorded tradeoff rather than an oversight.

Reported by Claude.

PR #1430, opened 2026-07-25
Make in-page links work in entry content (#1425)

### LINKED ISSUE #1425: In-page links in entry content are absolutized at write time, so they navigate off-site instead of jumping
Noticed while fixing #1422 (relative image URLs in saved GitHub Markdown).

Fragment-only links in entry content (`<a href="#bib.bib19">`) get rewritten to a full off-site URL at write time, so clicking a citation, footnote, or TOC entry leaves the reader instead of jumping to the target — which is usually **already in the same document**.

## Cause

`absolutizeUrls` (`src/server/feed/content-cleaner.ts`) has no fragment-only skip. `resolveUrl("#bib.bib19", base)` returns `${base}#bib.bib19`, which differs from the input, so the `[href]` handler writes it back. This runs at **write** time (`buildArticleFields` in `saved.ts`, `processEntryContent` in `feed/content-utils.ts`), so the stored `contentCleaned` has the off-site URL baked in.

It compounds at read time: because the href is now absolute, the sanitizer's external-link transform adds `target="_blank" rel="noopener noreferrer"` — so a citation click opens a **new tab**. The sanitizer already intends the opposite; the comment at `native/sanitizer/core/src/sanitize.rs:293` reads "Relative/in-page links are left alone." It never sees a relative one because absolutization got there first.

## Evidence

A saved arXiv paper (`2603.05706`, "Reasoning Models Struggle to Control their Chains of Thought"):

| | `contentOriginal` | `contentCleaned` |
|---|---|---|
| `id` attributes | 1559 | **1467** |
| bare `#frag` hrefs | 290 | **0** |
| absolutized frag hrefs | — | **166** → `https://arxiv.org/html/2603.05706#bib.bib19` |

So the 1467 anchor targets are sitting right there in the rendered document, and all 166 links that would reach them point off-site. arXiv papers are the worst case (citations, section refs, figure refs), but this hits LessWrong footnotes and any feed entry with a TOC too.

## Correction to a wrong premise

I previously told Brendan the sanitizer strips `id`, which is **false** — I'd grepped `native/sanitizer/src/lib.rs` (N-API bindings only) instead of the core crate. `GLOBAL_ATTRS` at `native/sanitizer/core/src/sanitize.rs:103` is `["class", "id", "title", "dir", "lang", "role"]`. Untrusted `id`s already flow through to the DOM unprefixed today.

That matters for scoping: **id namespacing is not a prerequisite for this fix.** Whatever collision risk a prefix would address already exists in production; fixing in-page links doesn't add to it.

## Fix

1. **Skip fragment-only hrefs in `absolutizeUrls`** — if the value starts with `#`, leave it. That alone fixes arXiv/LessWrong/feed content. Existing stored content keeps the baked-in URLs until refetched.
2. **Generate heading ids for GitHub Markdown** — a *second, independent* cause specific to #1422's path. `marked` emits no heading `id`s, so `wsff.md`'s hand-written TOC links (`#front-loading-alignment`) have no targets even after (1). Verified: 0 heading ids in that entry's stored HTML. Needs a GitHub-compatible slugger (`marked-gfm-heading-id` / `github-slugger`) so our slugs match the ones the author wrote against.
3. Check that a hash click behaves inside the SPA — the entry body renders via `dangerouslySetInnerHTML` in `EntryContentRenderer`, inside a scroll container, and `useParams`/pathname routing is custom (see `src/CLAUDE.md`). Native anchor jumps should work, but this needs an e2e test, not reasoning.

## On the nonce idea

Brendan asked whether to prefix ids and in-page links with a nonce rather than stripping. Recommendation: **a static prefix, not a nonce** — and treat it as separate hardening, filed independently of (1)/(2).

Why static is enough: only one entry body is in the DOM at a time (`EntryContentRenderer`), so cross-entry duplicate ids can't collide. A nonce's only advantage is uniqueness across simultaneously-rendered bodies, which doesn't happen — and if it ever does, prefixing with the entry id is deterministic and equally sufficient.

Why determinism is worth keeping: entry HTML is sanitized on **every read**, so a nonce would make the same entry serve different bytes each time (defeats ETag/HTTP caching, and makes sanitizer output untestable as a pure function of input). The sanitizer is deliberately a pure `html -> html` transform.

What a prefix would buy, if done:
- Our UI uses static ids (`custom-title`, `url`, `user-search`, `unsubscribe-title`, …). Entry content containing `id="url"` breaks `aria-labelledby`/`<label for>`, which resolve to the first match in document order.
- Mild DOM-clobbering hardening: any element with an `id` becomes a named property on `window`. A prefix containing `-` yields a non-identifier name, so it can't shadow a bare variable reference.

If we do it, the prefix must be applied to `id` attributes **and** same-document fragment hrefs in the same sanitizer pass, or (1) breaks again.

Reported by Claude.

brendanlong:
Follow-up to #1422, which merged before these commits landed. Rebased onto current master.

Closes #1425.

## The bug

Fragment-only links in entry content (`<a href="#bib.bib19">`) were rewritten to a full off-site URL at **write** time, so clicking a citation, footnote, or table-of-contents entry left the reader — usually to fetch something **already in the same document**.

`absolutizeUrls` had no fragment-only skip: `resolveUrl("#bib.bib19", base)` returns `${base}#bib.bib19`, which differs from the input, so the `[href]` handler wrote it back. It compounded at read time — because the stored href was now absolute, the sanitizer's external-link transform added `target="_blank"`, so a citation click opened a **new tab**. The sanitizer already intended otherwise; the comment in `svg.rs`'s HTML sibling reads *"Relative/in-page links are left alone."* It never saw a relative one because absolutization got there first.

A saved arXiv paper had **1467 ids preserved in the rendered body and all 166 of its in-page links pointing off-site**.

Separately, `marked` emits no heading ids, so a Markdown table of contents had nothing to land on even with the hrefs intact.

## What's here

**1. Fragment-only hrefs stay relative** (`content-cleaner.ts`). Also guards the `<base href>` handler: a fragment-only `<base href>` means "this document", and now that `resolveUrl` passes fragments through it would otherwise have become the base and broken every later resolution.

**2. Heading ids for all Markdown.** `marked-gfm-heading-id` (github-slugger) on the shared renderer, so slugs match what authors write their anchors against.

**3. One Markdown instance everywhere.** Every source — uploads, Markdown URL saves, GitHub repo files, AI summaries — now shares `src/server/markdown`'s single `Marked` instance; `summarization.ts` no longer reconfigures the `marked` global, and nothing imports it directly. A rule in `CLAUDE.md`'s Parsing section keeps a second instance from reappearing. This also gets repo files frontmatter, footnotes and math; a frontmatter `author`/`description` now outranks the repo owner, matching what an upload does with the same file.

**4. Id namespacing** (`native/sanitizer/core/src/idrefs.rs`). Entry content renders into the app's own DOM and `id` is allow-listed, so a feed carrying `id="url"` sat in the same namespace as our subscribe form's `id="url"` — and `getElementById`/`aria-labelledby`/`<label for>` all resolve to the first match in document order. Every surviving id gets a fixed `uc-` prefix.

A prefix rather than a nonce: sanitization is a pure `html -> html` function run on every read, so a nonce would make one entry serve different bytes per request (no ETag, untestable), and uniqueness across entries buys nothing while only one entry body is in the DOM at a time. The `-` also means `window["uc-foo"]` isn't identifier-shaped, so it can't shadow a bare variable reference.

Renaming an id is only safe if **every** reference moves with it, so this covers the whole surface the allow-lists keep: `a[href="#…"]`, `a[name]`, th/td `headers`, the ARIA idrefs, SVG `href` fragment refs, and — the part that would otherwise have broken rendering — `url(#id)` funcIRIs in the SVG paint/clip/mask/filter attributes, since a gradient whose id moves without its `fill="url(#…)"` renders unpainted. `idrefs.rs` also records what is deliberately *not* rewritten (filter `in`/`in2`/`result` name values inside one filter, not document ids) and what is unreachable because the allow-lists drop it. Prefixing is idempotent, because cached summaries are re-sanitized on read and a second `uc-` would break the match.

**5. `SANITIZER_VERSION` deleted.** It had no consumers — the machinery that compared it went away in #1282 — and its stale "bump this on any behavior change" comment was actively misleading. Removed from the core crate, napi export, typings and TS re-export.

## Verification

- Saved arXiv paper: 1559 ids all namespaced, **289/289 in-page links resolve** (was 289 pointing off-site), output idempotent.
- The humanlayer README: 40 heading ids (was 0), 13 in-page links kept relative (was 0), footnotes now render.
- A gradient/clip-path/mask chart round-trips through the split SVG pass with every reference resolving, and content `id="url"` no longer collides with our subscribe form.
- `scripts/bench/seed.sql` was independently broken (it inserted `content_*_sanitized` columns #1282 dropped). Fixed and actually run against a throwaway database — twice, since it's documented as an idempotent re-seed: 0 errors, 635k entries.
- 80 Rust tests, 2606 TS tests, typecheck and lint clean on the rebase.

## Known gaps, each with a skipped test or an issue

- **#1423** root-relative paths (`/docs/logo.png`) — GitHub reads these as repo-root-relative; no choice of base URL can express that. Two skipped tests, verified failing today.
- **#1424** gist sibling references.
- A duplicate `id` attribute (`<p ID="url" id="app">`) leaves the second un-namespaced. Browsers keep the first, so our renderer is unaffected and it isn't exploitable; it would only matter to an external consumer of the sanitized HTML with last-wins semantics.


brendanlong:
## Benchmarked the id change — it was a real regression, now mostly gone

I hadn't measured this, and I should have. `scripts/bench-sanitize.mts` already existed and its corpus happens to put an `id` on every paragraph, so it was the right harness.

**As originally written, namespacing cost +55–64%** of total sanitize time on id-carrying content. Two causes: `namespace_idrefs` probed 12 attributes *by name* on every element (12 case-insensitive linear scans of the attribute list, mostly to find attributes that are absent), and `attr.value()` allocates, so every attribute paid for a copy.

Fixes: piggyback on the attribute pass `handle_element` already runs, test the name before reading the value (with an `aria-` prefix check guarding the 8-entry list scan), and skip no-op writes — an already-prefixed value must not pay for a start-tag re-serialization.

Same corpora, same machine, `origin/master` vs branch:

| corpus | master | branch | delta |
|---|---|---|---|
| small 3.8 KB (11 ids) | 0.039 ms | 0.042 ms | +8% |
| medium 57 KB (150 ids) | 0.542 ms | 0.602 ms | +11% |
| large 463 KB (1200 ids) | 5.15 ms | 5.53 ms | +7% |
| mathHeavy 112 KB | 2.88 ms | 2.85 ms | ~0% |
| **mediumNoIds** 55.8 KB (0 ids) | 0.539 ms | 0.573 ms | +6% |

The remainder is largely irreducible: rewriting an `id` forces lol_html to re-serialize that start tag, which is inherent to renaming ids. I added `mediumNoIds` as a twin of `medium` so that cost stays attributable, and updated the doc's `~0.09 ms / 10 KB` figure — which matched master exactly on this machine, so my change genuinely moved it.

## Other review items

- **`scripts/bench/bench.sql` was broken the same way `seed.sql` was** — still selecting the `*_sanitized` columns #1282 dropped. Realigned with the current `selectFullEntry` and ran migrate + seed + bench against a throwaway database: 0 errors.
- **The shared marked instance is only concurrency-safe because parsing is synchronous.** `marked-gfm-heading-id` keeps its slugger in module-level state and resets it in a `preprocess` hook; an async extension would let two documents interleave and slug one against the other's occurrence table, silently breaking every table of contents under load. `markdownToHtml` now pins `{ async: false }` — marked *throws* if an extension forced async, turning silent corruption into a loud failure — and the misleading `as Promise<string>` cast is gone.
- **A lint rule now enforces the one-instance rule** (`no-restricted-imports` on `marked`, ignoring `src/server/markdown`), verified firing. Trimmed the doc restatements from four places to two.
- **My "rewrite raw values, never decoded" rule was wrong as stated** — `svg.rs` correctly does the opposite, because `scraper` hands it decoded values and `escape_attr` re-escapes `&`. Reworded per-pass so nobody "fixes" svg.rs to match.
- **Tests** for the markdown → read-path-sanitizer seam, which is the actual user-visible claim of #1425 and which neither unit test covered alone, plus the `xlink:href` form.
- **Filed #1431**: KaTeX amplifies ~29x (16 KB → 456 KB measured) and rendering is synchronous, while the size cap applies to raw bytes *before* rendering — so a 5 MB math-dense file renders to ~140 MB and ~13 s of event-loop-blocking CPU before being rejected. Pre-existing on the upload and full-content paths; this branch adds GitHub as a third entry point, which is why I'm flagging rather than quietly inheriting it.

The reviewer also confirmed the rebase is byte-identical to the pre-rebase branch (`git range-diff` reports `=` for all four commits) and found no remaining hole in the idref surface.

— Claude

PR #1437, opened 2026-07-28
Support OpenRouter for summaries and narration cleanup (#1416)

### LINKED ISSUE #1416: Support OpenRouter for summaries and narration cleanup
Rather than chase individual providers, we should just add support for OpenRouter. We might need to change our model selector to be search based, and it will need a way to filter out obviously-useless models (context too short, wrong kind of model).

brendanlong:
Closes #1416.

Rather than integrate providers one at a time, this adds **OpenRouter** — one key reaches models from every major lab. It slots into the existing provider layer as a fourth `provider:model` namespace and works for both article summaries and narration preprocessing.

## What the aggregator forced

**Model refs split on the first colon only.** OpenRouter's native IDs contain their own (`deepseek/deepseek-r1:free`), so `openrouter:deepseek/deepseek-r1:free` has to keep everything after the provider prefix.

**The model picker had to become search-based.** OpenRouter alone contributes ~350 models; a native `<select>` of that is unusable. Both model settings now use a new ARIA combobox (`src/components/ui/combobox.tsx`): type-to-filter with AND-ed terms matched against the label *and* the stored ref, grouped by provider, keyboard-navigable, capped at 50 rendered options with a "keep typing" notice.

Searching the model picker

**The catalog needs real filtering.** Of the 367 models OpenRouter lists we drop anything that can't take or emit text, anything under a 32k context window, and — for narration only, which parses a JSON object out of the response — anything that doesn't report `response_format` support. That leaves **350** for summaries and **299** for narration. The one fail-closed filter is the JSON one: a model that can't do JSON produces garbage rather than a clean failure.

## Notes

- The catalog endpoint is public and identical for every user, so it's cached process-wide with single-flight, and serves the stale copy when a refresh fails rather than emptying the picker.
- OpenRouter sorts **last** in both default-model priority lists — it can reach the other providers anyway, so a direct key wins when both are set. No existing user's effective default changes.
- Privacy policy updated (OpenRouter added to both AI subsections, including that it forwards to whichever upstream provider you picked; the narration section was also stale — it still named only Groq).
- Migration `0107` adds `users.openrouter_api_key`; additive, so it's backward-compatible with the previous release.

## Testing

- `pnpm typecheck`, `pnpm test:unit` (2660 passing), `pnpm test:integration:local` (720 passing) — all green.
- New unit tests: OpenRouter catalog filtering, the HTTP client's two error shapes + catalog caching/stale-fallback, the combobox's filtering/truncation/keyboard/cancel behavior, and colon-preserving ref parsing.
- **Verified end-to-end** against a local server with a live catalog: both pickers list and filter it, narration correctly shows 51 fewer OpenRouter models than summaries, and selecting a model round-trips to `openrouter:anthropic/claude-opus-4.6` in the database.


brendanlong:
Ran an independent Opus reviewer over the diff. It found six real issues, all now fixed in `42e5c72e` + `8aef2365`:

**Server**
- **No timeout on either OpenRouter fetch.** Every other outbound fetch in the repo passes `AbortSignal.timeout`; these inherited undici's ~5-minute default. That mattered most for the catalog, which is shared process-wide by a single-flight promise — one hung request would have hung *every* user's model picker behind it. Now 15s for the catalog, 180s for completions.
- **Catalog parsing was all-or-nothing.** `z.array(modelSchema)` fails wholesale if one of ~370 models changes a field's type, which would have silently emptied the picker of every OpenRouter model while other providers still appeared. Now per-entry `.catch(null)` + filter, which is what the doc comment already claimed.
- **A failed refresh re-attempted on every request.** The stale-cache fallback didn't age the entry forward, so an OpenRouter outage meant a fresh ~1 MB attempt per settings-page load. Now backs off a minute.

**Combobox**
- **Options selected on `pointerdown`.** That fires the instant a touch lands, making a 50-row list impossible to flick-scroll on a phone — the surface this control exists for. Now `preventDefault` on pointerdown (to hold input focus) + select on `click`. I verified in real Chromium that cancelling `pointerdown` still fires `click` *and* leaves `activeElement` on the input, since jsdom can't answer that.
- **`activeIndex` wasn't clamped** against the current list, so an options refetch while the listbox was open could leave Enter pointing past the end.
- **ARIA**: a listbox may only contain groups and options — status messages moved outside it, group headings now `role="group"`, `aria-controls` only while the listbox exists, empty/truncation messages in a live region, plus Home/End.

Also dropped three `summarization.ts` doc comments that still enumerated only the old three providers, and taught `formatModelName` that an OpenRouter variant suffix is a word break (`Deepseek R1 Free`, not `Deepseek R1:free`).

Unit tests are up to 2670 and pin each of these.

Two things the reviewer explicitly cleared that I'd flagged as risky: the single-flight cache has no rejected-promise-caching or unhandled-rejection bug, and **not** bumping the `session:v2:` Redis prefix is correct here — `?? false` on the new `userHasOpenrouterApiKey` field is fail-closed (a stale entry hides the feature for ≤5 min rather than granting anything), which is exactly the condition the prefix-bump rule is written around.

— Claude

brendanlong:
I can't decide if the complexity here is worth it

PR #1438, opened 2026-07-28
Render Markdown natively so it can't block the event loop (#1431)

### LINKED ISSUE #1431: Markdown rendering blocks the event loop and can amplify content ~29x before the size cap applies
Found while reviewing #1430, which consolidated every Markdown source onto one renderer and so added a new entry point to this path.

## Problem

Markdown rendering (`markdownToHtml` / `processMarkdown`, `src/server/markdown/index.ts`) is fully synchronous and unbounded, and KaTeX amplifies input substantially:

| input | output | ratio | time |
|---|---|---|---|
| `"$a_1^2$ "` × 2000 (16 KB) | 456 KB | **28.5x** | 42 ms |
| 200-row `\begin{matrix}` | — | ~19x | — |

The size limit is checked in the wrong order relative to the work. `usageLimitsConfig.maxSavedArticleSizeBytes` (5 MB) is enforced on the **raw fetched bytes** (e.g. `src/server/plugins/github.ts` for repo files), and the limit on rendered HTML is only applied *after* the plugin returns (`src/server/services/saved.ts`). So a 5 MB math-dense Markdown file renders to roughly 140 MB of string and ~13 s of **event-loop-blocking** CPU on the app server before being rejected for being too large.

This contrasts with the deliberate protection the neighbouring stages already have: `src/server/html/CLAUDE.md` documents that both sanitization and Readability extraction offload to the libuv thread pool above ~10 KB precisely so a large body can't block the event loop. Markdown rendering is the most expensive of the three and has no equivalent.

## Affected paths

Pre-existing on:
- Markdown URL saves (`saved.ts`)
- `full-content.ts`

New in #1430 (same code, additional reachable source):
- GitHub repo files / gists — fully third-party content, reachable by saving any URL

## Possible directions

- Check a rendered-size budget *during* or immediately after rendering, inside the renderer, rather than at the caller.
- Cap Markdown input at something well below the 5 MB raw cap, given the measured amplification.
- Offload rendering the way sanitize/Readability do, so at minimum it doesn't block the event loop.

Reported by Claude.

brendanlong:
Fixes #1431.

## Problem

Markdown rendering was the most expensive step in the content pipeline and the only one that could stall the app server. `marked` + KaTeX ran fully synchronously and amplified math-dense input ~29x, and the size limit was checked in the wrong order relative to the work: `maxSavedArticleSizeBytes` guarded the **raw fetched bytes**, and the rendered HTML was only measured *after* the plugin returned. A 5 MB math-dense Markdown file therefore rendered to ~140 MB of string and ~13 s of event-loop-blocking CPU before being rejected for being too large.

The neighbouring stages already had the protection this one lacked — `src/server/html/CLAUDE.md` documents that both sanitization and Readability extraction offload to the libuv thread pool above ~10 KB.

## Approach

A fourth native module, `@lion-reader/markdown` (`native/markdown/`): [comrak](https://github.com/kivikakk/comrak) for GFM, [pulldown-latex](https://crates.io/crates/pulldown-latex) for `$…$` / `$$…$$` TeX → MathML. This is the only option that fixes **both** halves rather than bounding them:

| case | marked + KaTeX | comrak + pulldown-latex |
|---|---|---|
| 16 KB math bomb (`$a_1^2$` ×2000) | 456 KB, **42 ms** | 160 KB, **1.9 ms** |
| 5 MB math bomb (the issue's worst case) | ~140 MB, **~13 s** | 51 MB, **0.67 s** |
| 200-row `\begin{matrix}` | ~19x | 6.7x |
| 1 MB ordinary prose | — | 31 ms |

Being native also means an `AsyncTask` form, so rendering hands off to the libuv thread pool above ~10 KB exactly like `sanitizeEntryHtmlAsync` and `cleanContentAsync` — no new bundle, no Dockerfile worker path, no dev/prod split.

### Budgets moved inside the renderer

Markdown *grows* on the way to HTML, so a raw-bytes check on the way in is the wrong budget:

- **`maxMarkdownInputBytes`** (1 MB, well under the 5 MB raw cap) short-circuits before parsing.
- **`maxRenderedMarkdownBytes`** (5 MB) is enforced by a `fmt::Write` sink that aborts the formatter mid-write, so an amplifying document is rejected *while* it amplifies. The math walk counts against the same budget, so a math bomb can't fill the AST before the formatter ever runs.

Both are configurable via env and surface as the usual `contentTooLarge` error. `saved.uploadFile` now rethrows `TRPCError`s unchanged instead of re-wrapping them, so `CONTENT_TOO_LARGE` isn't buried under a generic `BAD_REQUEST`.

### Dialect

Parity is close to exact. The full suite passed with only the **footnote anchor names** changing (`#fn-1`/`#fnref-1` — GitHub's own, self-consistent so links still resolve). Verified unchanged: heading slugs (`Anchorizer` implements GitHub's algorithm — `## What's *new* in v2.0?` → `whats-new-in-v20`, duplicate `## Intro` → `intro`/`intro-1`), GFM tables/tasklists/strikethrough/autolinks, `breaks: true`, raw-HTML passthrough, and prose `$5 and $10` not firing as math.

Two incidental wins:

- No `<annotation>` copy of the TeX source — the read-path sanitizer dropped it anyway, so KaTeX was emitting pure amplification.
- The heading occurrence table is per-render state inside the Rust call, so the module-global slugger hazard `marked-gfm-heading-id` carried (flagged in the issue discussion) is gone by construction.

Drops `marked`, `marked-footnote`, `marked-gfm-heading-id`, `marked-katex-extension` and `katex`. `markdownToHtml` is no longer exported — every Markdown source in the app is reached from a request, so no caller should be picking the blocking form.

## Verification

Not just tests — the change was driven end to end against a running app:

- **Unit** 2631 passed · **integration** 720 passed · **e2e** 64 passed · **Rust core** 20 new tests (dialect, budgets, deep-nesting stack safety)
- **Real app, dev server**: uploaded a math + footnotes + table + TOC document; read it back through `entries.get` and confirmed MathML, working footnote back-links, and **zero dangling `#fragment` anchors** after the read-path sanitizer's id namespacing.
- **Real app, production bundle** (`next build` → `node dist/server.js`): same upload worked, which is what exercises the ESM/CJS native-loader interop that only breaks outside dev. `fixup-standalone.mjs` creates the `@lion-reader/markdown` symlink; `marked`/`katex` are gone from the standalone trace.
- **Budgets, live**: 1.5 MB of math → rejected on the input cap; 896 KB of math (inside the input cap) → rejected on the output budget as it amplified. Both in ~150 ms round-trip.
- **Event loop, live**: a 480 KB math-dense upload that renders to ~4.8 MB, with `/api/health` polled throughout — median **15 ms**, p95 18 ms while it rendered.


brendanlong:
Pushed a follow-up commit addressing an independent review of this branch. Two findings were serious enough that they'd have shipped a regression:

**The byte budgets couldn't see the worst cost.** `pulldown-latex` is quadratic in TeX nesting depth, and the work happens inside a single call that returns a *small* result — so neither the input cap nor the output budget could stop it. Measured on `$$\frac{1}{\frac{1}{…}}$$`: depth 1k/2k/4k/8k/16k → 17/66/268/1060/4351 ms, and a 1 MB document (i.e. right at the input cap) of `$${{{{…}}}}$$` pinned a thread for **~3 minutes**. With `UV_THREADPOOL_SIZE` defaulting to 4, a handful of those starve sanitization, Readability *and* `dns.lookup` — strictly worse than the event-loop stall this issue set out to fix, and invisible to event-loop-delay alarms because the work is off-thread.

Fixed by capping group nesting at 64 (real math lives in the low tens) and degrading past it. That 1 MB document now renders in **5.5 ms**, and end-to-end through the app the upload round-trips in 314 ms.

**Unescaped TeX was reaching the output.** `pulldown-latex` writes the offending source into `<merror><mtext>` unescaped, and `render.unsafe` passed it straight through. `$\badcmd{<style>x}$` injected a live element, and the read-path sanitizer — doing exactly the right thing with an unclosed `<style>` — dropped the **entire rest of the entry**. Reachable by accident, not just by attack: `$\bm{a}<b$` is ordinary TeX using a macro `pulldown-latex` doesn't support. No XSS was achievable (the sanitizer held on everything tried), but handing it malformed markup is not a posture worth keeping.

Fixed by running the parse to completion before rendering and handling the error ourselves, so nothing attacker-controlled reaches the output except through `escape`. A side benefit: unsupported macros (`\newcommand`, `\bm`, `\DeclareMathOperator`, …) now degrade to readable escaped source instead of a multi-line box-drawing error blob dropped into the article body.

Also fixed: `full-content.ts` does render Markdown from the feed worker, so `processMarkdown` now takes an `offload` flag that `fetchFullContent` threads its existing `offloadClean` through (the "no background-job caller" claim was wrong in both the code and the docs); the output-budget test wasn't testing the output budget (its input tripped the *input* cap, so it passed identically with the output budget deleted); the budgets are clamped before crossing the N-API `u32` boundary, where a malformed env var became `0` and silently rejected every document; plus dead-code and doc cleanups.

One known dialect delta I decided **not** to chase: `comrak` omits `marked-footnote`'s visually-hidden "Footnotes" heading and its `aria-describedby`. comrak matches GitHub's own footnote markup exactly, which is the stated goal of the dialect, and restoring the heading would mean either string surgery on generated HTML or forking comrak's footnote rendering. Flagging it as a real if minor a11y difference rather than an oversight.

Re-verified after the fixes: 2631 unit · 720 integration · 64 e2e · 26 Rust tests, plus both fixes driven end-to-end through the running app.

— Claude

Issue #1443, opened 2026-07-29
Saved articles discard the source's publish date


brendanlong:
## Problem

`insertSavedEntry` hardcodes `publishedAt: null` (`src/server/services/saved.ts:514`), so every saved article shows and sorts by its save time regardless of when it was actually published.

Several sources hand us a real publish date that we then throw away:

- `SavedArticleContent.publishedAt` is part of the plugin contract (`src/server/plugins/types.ts:153`) and Bluesky already returns the post's `createdAt` (`bluesky.ts:436`)
- The LinkedIn plugin returns JSON-LD `datePublished`, which is the one piece of metadata LinkedIn gives us that Threads doesn't
- ArXiv and LessWrong both have the date available from their APIs

For a saved LinkedIn post or a saved blog entry, "this is from 2019" is genuinely useful information that we currently can't show.

## Why it isn't a one-line fix

The timeline sort key is `COALESCE(published_at, fetched_at)`, denormalized onto `user_entries.published_or_fetched_at`. If we simply stop hardcoding `null`, an old article saved today sinks to the bottom of the saved list and is effectively never seen. **Sort must stay pinned to save time.**

The sort is separable: the `user_entries_fill_denormalized` trigger `COALESCE`s an explicitly-passed `published_or_fetched_at` before falling back to the entry's (`migrations/schema.sql:145-153`), so `insertSavedEntry` can pass `now` and keep save-order sorting while storing a real `published_at`.

Display is the harder part. `EntryListItem.tsx:112` and `EntryContent.tsx:348` both render `publishedAt ?? fetchedAt`, so storing the date changes what the saved list shows — and a list sorted by save time but *displaying* publish dates reads as broken ("Apr 2025, Jan 2020, Jul 2026" going down the page).

## Probable shape

Store `published_at`, pin `published_or_fetched_at` to save time, keep the saved list showing save date, and surface the real publish date in the opened article's header. That needs the entry-content date to stop being a blind `publishedAt ?? fetchedAt` fallback, plus a decision about what feed entries (where the two coincide) should do.

Deferred deliberately — the display/sort split has enough edge cases to deserve its own change rather than riding along with the LinkedIn/Threads plugins.

---
Filed by Claude.

brendanlong:
Not going to do this for now since it would be confusing

Issue #1442, opened 2026-07-29
Browser extension should submit the rendered page source when saving


brendanlong:
## Problem

Server-side fetching can't see login-walled or JS-rendered pages. Today we work around that one site at a time with plugins that hit an API (Bluesky) or scrape whatever structured metadata the public page leaks (YouTube, and now LinkedIn/Threads — see #TBD). Each of those is brittle by construction: it breaks when the site changes its markup, when the post is private, or when the site blocks our Fly egress IP. LinkedIn in particular is well known for serving HTTP 999 to datacenter ranges, and both `linkedin.com/robots.txt` and `threads.com/robots.txt` are `Disallow: /` for a generic user agent.

The general fix is to let the client — which is already logged in and has already rendered the page — hand us the DOM.

## What already exists

The server side is done and unused:

- `SaveArticleParams.html` (`src/server/services/saved.ts:78`) — "Pre-fetched HTML (e.g. a bookmarklet capturing the rendered DOM)… useful for JavaScript-rendered pages where a server-side fetch would miss content"
- Consumed at the top of `acquireArticleContent` (`src/server/services/saved.ts:883`), ahead of plugins and the plain fetch
- Exposed on tRPC/REST `saved.save` with a size cap (`src/server/trpc/routers/saved.ts:126-131`)

No client produces it. The extension (`extension/src/background.js:44`) and the `/save` page (`src/app/(spa)/save/page.tsx:145`) both send URL + title only.

## Proposal

Have the extension capture `document.documentElement.outerHTML` from the active tab and send it as `html` on `/api/v1/saved`.

Worth thinking through before building:

- **Permissions.** This needs content-script/`activeTab` access to page DOM. That's a meaningful escalation of what the extension asks for at install time; it should be obvious to the user what it's for, and it may warrant being opt-in.
- **Size.** Rendered DOM of a heavy page is far bigger than what the server fetches. The existing cap applies, but we should measure typical sizes and decide whether to strip `<script>`/`<style>`/inline data before sending rather than just failing large pages.
- **Trust.** Caller-supplied HTML already flows into `contentOriginal` and gets sanitized on every read like any other content, so this isn't a new XSS surface — but it is user-controlled input taking precedence over everything else, which is worth a second look in review.
[...]

PR #1444, opened 2026-07-29
Add LinkedIn and Threads plugins for saving posts

brendanlong:
Saving a LinkedIn or Threads URL previously ran Readability over a login-walled JS page and stored a sign-up prompt. These plugins render the post instead.

## Why scraping and not an API

I checked both. Neither has a usable public read API:

| | Status |
|---|---|
| **LinkedIn** | Open API access ended in 2015. The Community Management / Posts API is partner-gated and scoped to content the authenticated member owns or administers — nothing reads a third party's post. No public oEmbed. |
| **Threads** | `threads_basic` reads only your own posts. Reading anyone else's public post needs the Keyword Search API with `threads_keyword_search`, which requires Meta App Review **and** business verification. |

Threads' oEmbed (`graph.threads.net/v1.0/oembed`) does work without a token, but the `html` it returns is just the `<blockquote>` placeholder `embed.js` hydrates client-side — no post text.

What both *do* serve to a logged-out client is structured metadata, which is what these plugins read.

## What the plugins extract

**LinkedIn** (`src/server/plugins/linkedin.ts`) — the public post page carries a JSON-LD `SocialMediaPosting` with `articleBody` (the full body, past the "see more" fold), `datePublished`, `author`, and the attached `image`. A video post is a `VideoObject` whose commentary is in `description`; its poster frame links back to the post, since LinkedIn video is a streaming manifest a bare `<video>` can't play. Both public URL forms serve the same page, so both are matched:

- `/posts/{slug}-activity-{id}-{code}`
- `/feed/update/urn:li:{activity,share,ugcPost}:{id}`

Falls back to `og:description` (minus LinkedIn's trailing `| 21 comments on LinkedIn` chrome) when there's no usable JSON-LD. `/pulse/` articles, profiles, and company pages deliberately don't match — they fall through to normal handling.

**Threads** (`src/server/plugins/threads.ts`) — `og:description` carries the complete post text; because Threads posts are short it isn't elided the way `twitter:description` is. `og:title` is always `"Display Name (@handle) on Threads"`, which gives the author. Matches `/@handle/post/{code}` (with or without the SEO slug Threads appends) and the `/t/{code}` short permalink.

Two things Threads doesn't give us, documented in the plugin:
- **No publish date** anywhere — no `<time>`, no JSON-LD, no timestamp in the inline payload, and the shortcode doesn't encode one.
- **No media.** `og:image` is the author's avatar on a text post and the attached media on an image post, with nothing in the markup to distinguish them, so rendering it would decorate every text save with an avatar. Neither is rendered.

## Failure behavior

Both are best-effort by construction. Unusable markup, a post gone private, or a blocked fetch returns `null` from `fetchContent` and the save falls back to normal handling — no new failure mode, just a better result when it works. This matters because LinkedIn is well known for serving HTTP 999 to datacenter ranges, so it may degrade on Fly even though it works locally.

Both sites' `robots.txt` are `Disallow: /` for a generic user agent. This is a user-initiated single-post fetch on save rather than crawling, which is the distinction every read-it-later app leans on, but it's called out in both plugin headers and in `docs/DESIGN.md` so nobody extends it into something that walks profiles or feeds.

## Shared extractions

Three plugins now render plain-text prose as HTML and three synthesize a title from a post's first line, so both moved out:

- `plainTextToHtml` in `src/server/http/html.ts` (was `youtubeDescriptionToHtml`) — escape, paragraphs, linkify
- `socialPostTitle` in `src/server/plugins/social-post.ts` (was the body of `blueskyPostTitle`, which is now a wrapper)

## Verification

Beyond the unit tests, I drove both plugins against **live** pages and then through the full `saveArticle` path against a local Postgres:

- LinkedIn text post, video post, and a `/feed/update/` permalink → correct title, author, `datePublished`, body, and image figure
- Threads canonical and `/t/` URL forms → correct title, author, full untruncated body
- A LinkedIn profile URL correctly does **not** match
- End-to-end saves store the right `title` / `author` / `siteName` / `summary` / `content_cleaned`

`pnpm typecheck`, `pnpm test:unit` (2665 passed), and `pnpm test:integration` (720 passed, 15 skipped) all pass.

## Known gaps, filed separately

- **#1443** — plugin `publishedAt` is still discarded (`insertSavedEntry` hardcodes `null`), so LinkedIn's `datePublished` goes nowhere for now. Deferred deliberately: storing it naively would sink an old article you save today to the bottom of the saved list, and the display-vs-sort split needs its own change.
- **#1442** — the extension should submit the rendered page source. That's the general fix for login-walled sources and would make these scrapers a fallback rather than the only path.
- `SavedArticleContent` has no image field, so a plugin's hero image only survives if inlined in the body (LinkedIn does; `imageUrl` on the row stays null).


brendanlong:
Pushed 59048975 fixing findings from an independent review pass. Three of these were real bugs I'd have shipped:

**A 500 on save, not a fallback.** `matchUrl` called `decodeURIComponent` on the path, which throws `URIError` on a malformed escape like `/%` or `/feed/update/%E0%A4%A`. Both `new URL()` and the save path's `z.string().url()` accept those, and the registry calls `matchUrl` *outside* the try that wraps `fetchContent` — so it surfaced as an internal error rather than falling back. Added `safeDecodeURIComponent` in `src/lib/url.ts`. **The Bluesky plugin has had the same latent bug**, so it's fixed there too.

**LinkedIn could save the wrong text as the post.** I picked the first JSON-LD entity with *any* body field. A page can carry an `Organization` block — whose `description` is "LinkedIn is the world's largest professional network" — before the post, and it would win on ordering alone. Now prefers any entity with `articleBody` before falling back to `description`. My original test for this used a `BreadcrumbList`, which has neither field, so it gave false confidence.

**`@graph` was documented as handled and wasn't** — a `@graph`-wrapped post rendered as `null`. Implemented the unwrap rather than deleting the claim.

Plus: rate-limit errors were swallowed, so a 429 fell through to a plain fetch of the same URL (a second request to a host that just throttled us, and a generic error instead of `upstreamRateLimited`); `headline` bypassed the title length cap, so a remote-controlled 5000-char headline became a 5000-char title; schema.org's single-or-array and bare-string-URL shapes weren't handled for `author`/`creator`/`image`; and Threads' `parser.pause()` at `</head>` made og tags emitted after it invisible for no measurable gain.

Both plugins now share `fetchPluginPage`, which encodes the fall-back-vs-fail decision in one place: rethrow a rate limit, keep falling back on a plain block (403, or LinkedIn's nonstandard 999 for datacenter IPs, which isn't a retry-later signal).

That classification is exactly the part unit tests can't reach and review is worst at catching, so it's covered by a new **integration** test driving a loopback server — 429 rethrows; 403/999/404/non-HTML/unreachable all return null. Every other finding got a unit test, plus a `socialPostTitle` file pinning the elision boundary all three social plugins now share.

Re-verified against the live LinkedIn and Threads pages after the refactor: same output as before, no regression. Typecheck, unit (2681), and integration (728) all pass.

The review found no XSS — it built a malicious JSON-LD payload and ran it through `renderLinkedInPost` → `sanitizeEntryHtml` with everything escaped at both stages, and disproved two hypotheses I'd have worried about (htmlparser2 does not decode entities inside `<script type="application/ld+json">` even with `decodeEntities: true`, and non-object JSON-LD is correctly filtered).

One finding I deliberately did **not** change: the LinkedIn `og:description` fallback can in principle save a non-post interstitial's description as the body, if LinkedIn ever serves a 200 interstitial on a URL that matches `/posts/…-activity-…`. Live probes were reassuring — `/authwall` and a 404 post both serve no `og:description` and correctly return null — and removing the fallback would lose the only path that survives LinkedIn dropping JSON-LD. I reworded the misleading comment instead. Happy to drop the fallback entirely if you'd rather not carry that risk.

Claude

brendanlong:
Pushed 8871bde0. All three points taken — and the Readability question turned up that **my stated rationale for the LinkedIn plugin was wrong**, which is worth correcting in the record.

## Readability is not broken on LinkedIn

I asserted it "extracts the auth wall." I hadn't measured it. Running `cleanContentAsync` over the real captured pages:

| | Readability | Reality (JSON-LD) |
|---|---|---|
| Text post body | 421 chars (includes the link-card chrome) | 230 chars, exactly the post |
| Video post body | **2754 chars** | **834 chars** — the rest is the comment thread |
| Video post byline | **"Jon Rosemberg"** | "Jason Feifer" — Readability picked a *commenter* |
| Title | `How to write attention-grabbing LinkedIn posts \| Jason Feifer posted on the topic \| LinkedIn` | `Want to get more attention on LinkedIn?` |

So it returns a genuinely usable article; it's just consistently worse. Readability *does* fail outright on Threads (returns null), so that plugin's rationale stands unchanged.

That reframes what the plugin owes the caller: **falling back is a good outcome, not a failure** — so guessing is never worth it. Which is exactly your point, and it makes both other changes obvious rather than defensive.

## Match on `@type`, don't sniff field shape

You're right that they give us an explicit tag and I was ignoring it. Now:

```ts
const POST_TYPES = {
  SocialMediaPosting: ["articleBody"],
  DiscussionForumPosting: ["articleBody"],
  VideoObject: ["description"],   // video posts have no articleBody
};
```

Keying off the type LinkedIn already declares **removes** the guess rather than patching its ordering — the `Organization` bug becomes structurally impossible instead of fixed-by-precedence. An unrecognized `@type` now falls back.

One subtlety this made expressible: `description` is deliberately *not* read for a text post, because there it describes the **shared article**, not the post. The old field-shape sniff couldn't distinguish those.

## Dropped the og:description fallback

Same mistake in another place, and you called it. LinkedIn serves an `og:description` on non-post pages too, and on a link-share post it describes the linked article — so it could store something that isn't the post, silently, since `skipReadability` means nothing downstream re-checks the extraction. Gone. Threads keeps reading `og:description` because there it genuinely *is* the post text, not a description of something the post links to.

## Removed the dead image option

`SavedArticleFetchOptions.uploadImages` was declared in the plugin contract and read by nobody. No plugin used the `options` parameter at all and both call sites passed `{}`, so the whole parameter went with it.

---

Re-verified against the live pages after the refactor: all six cases produce identical output to before. Typecheck, unit (2682), integration (728), lint, and knip all clean.

Net effect on the risk I flagged earlier: the "silently extract the wrong thing" class is now much narrower — the only remaining way in is LinkedIn emitting a *different* entity that declares one of the three post types on the same page, rather than any entity that happens to have a body-shaped field.

Claude

brendanlong:
Second review pass done; pushed 3f1a5656. It found five findings that all trace to **one root cause I'd left in place**: JSON-LD is remote JSON, but I typed it as strings and read it without validating.

## The contract was broken in four ways

| Input | Old behavior |
|---|---|
| `@type: "constructor"` / `"toString"` / `"__proto__"` | resolved on `Object.prototype` through the `POST_TYPES` record lookup and **threw** |
| non-string `articleBody` / `headline` / `author.name` | **threw** on `.trim()` |
| `image: {url: ["https://ok.jpg", "javascript:…"]}` | **passed the `^https?://` gate** — `RegExp.test` string-coerces the array |
| `datePublished: 123456789` | stored `1970-01-02T10:17:36.789Z` as the publish date |

The throws are the ones that bother me most: this plugin's whole stated contract is *decline, never guess*, and a throw on the decline path defeats the point of declining. (No XSS from the URL one — escaping threw first and the read-path sanitizer runs regardless — but the gate was wrong as written.)

Narrowing now happens once in Zod schemas, so the render path only ever sees validated values. That also picked up shapes my hand-rolled checks had missed: array-wrapped `@type`, bare author-name strings, `thumbnailUrl` arrays.

## The `@type` gate had a hole I'd just introduced

Selection was positional first-match over a flatten order that **isn't document order** (`@graph` members append last). A post that shares a video carries *both* a `SocialMediaPosting` (the member's commentary) and a `VideoObject` (the shared media) — so first-match could save the video's description instead of what the member actually wrote. That's the same mis-selection class the `@type` gate was introduced to eliminate, reintroduced through ordering. `POST_TYPES` is now an ordered list; the post's own types outrank the media's.

## Mutation-tested rather than assumed

A passing test isn't evidence it would catch the bug, so I reverted each fix and confirmed the new tests fail:

- positional first-match → `× prefers the post's own entity over a shared VideoObject, in either order`
- unvalidated body read → 2 failures

Also sharpened two tests the review called out as unable to fail: one asserted only absence and was subsumed by an exact-match assertion above it; the other used a fixture where `headline` was byte-identical to the body's first line, so it couldn't tell which was used. And added the missing `DiscussionForumPosting` case.

## Docs

DESIGN.md was restating the Readability numbers and robots.txt constraints that the plugin files already carry. It now states only the cross-cutting rule — claim a page positively or decline; measure the generic path before writing an extractor — and points at the plugins for per-source specifics. Dropped the "the general fix is client-side capture" line, which described something not built (#1442).

## Not changed

`tests/integration/plugin-fetch-page.test.ts` touches no database, and the review suggested moving it to `tests/unit/`. Leaving it: it makes real loopback HTTP requests, which `tests/unit/` is specified as "pure logic, no DB" — and `tests/integration/ssrf-fetch.test.ts` is the existing precedent for exactly this shape.

Re-verified against the live pages after the rewrite: all six cases identical. Typecheck, unit (2703), integration (728), lint, knip clean. Working tree and stash list empty.

Claude

brendanlong:
Transcript support pushed (098bc935), plus a correction to what I said about the iframe.

## Correction: the compact embed excludes comments

I said the iframe renders "the whole post". I'd only fetched the *non*-compact page. `embedUrl` carries `?compact=true`, and that variant is 20 KB with the author, the commentary, and the video player — **no comment thread** (the commenter who pollutes Readability's extraction is absent from it). So it's "the post minus comments", not "just the video" and not "everything".

That still leaves it duplicating the commentary text we already extract, and it would need LinkedIn added to the sanitizer's block-by-default embed provider list plus the CSP that double-enforces those hostnames — a security-surface change. So still not doing it, but the reasoning is narrower than I stated.

## Transcript

```html
<details><summary>Video transcript</summary><p>…</p></details>
```

`<details>`/`<summary>` were already in the sanitizer allow-list, and omitting the `open` attribute is what makes "collapsed" survive the read path.

One thing worth calling out that isn't obvious: I set `excerpt` to the post text explicitly. Without that, the saved article's list summary is derived from the whole body, so a 1000-character machine-generated transcript would bleed into the summary of any video post with short commentary.

Verified end to end against the live video post rather than only in unit tests — 1143 characters of transcript render collapsed, pass through the **real** read-path sanitizer byte-identical (1143 → 1143, `<summary>` intact, no `open` attribute injected), the post's own text still comes first, and the excerpt stays the commentary.

Caveat: I've sampled exactly one video post, so I don't know how reliably LinkedIn populates `transcript`. Absent or wrong-typed, the block is simply omitted — that's covered by a test.

Unit 2706, integration 728, typecheck, lint all pass.

Claude

Issue #1445, opened 2026-07-29
Narration reads blockquote/table text twice when the wrapper holds block children


brendanlong:
Found while fixing #1441 (list items), which is the same bug in a different wrapper. Not caused by that fix — it reproduces on `master`.

Server-side `htmlToNarrationInput` builds a blockquote's (and a table's) narration from its full `textContent`, then the block-element walk descends into it and narrates its block children again:

```js
htmlToNarrationInput("<blockquote><p>A famous quote.</p></blockquote>")
// [{ id: 0, text: "Quote: A famous quote. End quote." }, { id: 1, text: "A famous quote." }]

htmlToNarrationInput("<table><tr><td><p>Cell 1</p></td></tr></table>")
// [{ id: 0, text: "Table: Cell 1 End table." }, { id: 1, text: "Cell 1" }]
```

`<blockquote><p>…</p></blockquote>` is the shape every Markdown renderer emits for a `>` quote, so this is common in feed HTML.

The client-side path (`htmlToClientNarration`) no longer duplicates — #1441's fix made its `processInlineContent` skip block children, and it has no markers to preserve. The server path is harder because the markers wrap the text: `Quote: ` would have to move onto the first block child and ` End quote.` onto the last (`- ` for list items only needed a prefix, which is what #1441 does via `inheritedListMarker`).

Skipped tests for the intended behavior are in `tests/unit/html-to-narration-input.test.ts` (`blockquote handling` / `table handling`).

Reported by Claude.

brendanlong:
Already fixed on `master` — closing.

The fix landed in 2f413391 ("Narrate a quote or table once, not once per wrapper (#1445)"), which is an ancestor of `origin/master`. The marker handoff the issue describes lives in `src/lib/narration/runs.ts:193-194`: the `Quote: ` prefix goes onto the first run and ` End quote.` onto the last, so the wrapper no longer flattens its own `textContent` alongside descending into its block children.

Verified against both reproductions from the issue on current `master`:

```
blockquote: {"paragraphs":[{"id":0,"o":1,"text":"Quote: A famous quote. End quote."}]}
table:      {"paragraphs":[{"id":0,"o":0,"text":"Table: Cell 1 End table."}]}
```

[...]

Issue #1439, opened 2026-07-29
Sanitizer drops GFM task-list checkboxes and table alignment, so done/not-done render identically


brendanlong:
Found while exercising the full Markdown dialect through `upload_article` after #1438 deployed. Both are **pre-existing** — verified that marked 18 emitted the identical markup, so the sanitizer has always dropped it — but the task-list one is user-visible enough to be worth fixing.

## 1. Task lists lose their checked state entirely

The renderer emits GFM's standard task-list markup:

```html
<li><input type="checkbox" checked="" disabled="" /> done</li>
<li><input type="checkbox" disabled="" /> todo</li>
```

`input` isn't in the sanitizer's tag allow-list, so both become:

```html
<li> done</li>
<li> todo</li>
```

**A completed task and an incomplete task render identically** — there is no checkbox, no strikethrough, nothing. A checklist in a saved README or uploaded note reads as a plain bullet list with a stray leading space, and the reader can't tell which items are done. That's the actual information the markup carried.

Verified against marked 18.0.4 (`<input checked="" disabled="" type="checkbox">`) — same shape, same outcome, so this predates the native renderer.

## 2. Table alignment is dropped

`| :- | :-: | -: |` produces `<th align="left">` / `align="center"` / `align="right"`, and `align` isn't in `GLOBAL_ATTRS` or any per-tag list, so every column renders default-aligned. Also identical under marked. Lower impact, but same root cause and same file.

## Possible directions

Both live in the allow-lists in `native/sanitizer/core/src/sanitize.rs`, but they're different calls:

- **Alignment** is just an attribute allow-list entry (`align` on `th`/`td`, value-checked against `left|center|right`). `align` is a presentational attribute with no script surface. Or map it to a class if we'd rather not admit a deprecated attribute.
- **Checkboxes** mean admitting `input`, a form element — more security-sensitive, so probably worth constraining hard (`type="checkbox"` only, `disabled` forced on, `name`/`value`/`form*`/event handlers dropped). Alternatively the renderer could emit something inert instead of an `<input>` — e.g. `<span class="task-list-item-checkbox" aria-hidden="true">` plus the accessible state in text — which keeps the sanitizer's tag list unchanged. Note the entry-content CSS would need a rule either way, since an unstyled disabled checkbox in prose looks out of place.

Reported by Claude.

PR #1446, opened 2026-07-29
Narrate a list item once when it wraps block children (#1441)

### LINKED ISSUE #1441: Narration reads a list item twice when it wraps block children
Found while adding task-list handling to narration in #1440 (not caused by it — reproduces identically on `master`).

`htmlToNarrationInput` emits a paragraph for the `<li>` (using its full `textContent`) *and* separate paragraphs for each block child, so the item's text is narrated twice:

```js
htmlToNarrationInput("<ul><li><p>hello</p></li></ul>")
// [{ id: 1, text: "- hello" }, { id: 2, text: "hello" }]

htmlToNarrationInput("<ul><li><p>a</p><p>b</p></li></ul>")
// [{ id: 1, text: "- ab" }, { id: 2, text: "a" }, { id: 3, text: "b" }]
```

Listening to that, you hear "hello. hello." — and in the multi-paragraph case, "ab" (the concatenation, with no separator) followed by "a" and "b" individually.

This is the **loose list** shape, which is what cmark-gfm/GitHub emit for any list with blank lines between items, so it's common in feed HTML rather than exotic.

`src/lib/narration/html-to-narration-input.ts:88` handles `li` by returning `- ${el.textContent}`, and the block-element walk then descends into the `<li>`'s children and emits them too. The `ul`/`ol` branch just above returns `""` for exactly this reason (containers have no text of their own); an `li` with block children is in the same position but isn't treated that way.

Likely fix: when an `li` contains block-level children, return `"- "` as a marker (or `""`) and let the children narrate themselves, rather than duplicating their concatenated text. Note the paragraph indices must keep matching the client-side highlighting implementation, so this needs care.

Reported by Claude.

brendanlong:
Fixes #1441 and #1445.

## Problem

`htmlToNarrationInput` emitted a paragraph for an `<li>` built from its full `textContent` **and** separate paragraphs for each of its block children, so a loose list item was narrated twice:

```js
htmlToNarrationInput("<ul><li><p>hello</p></li></ul>")
// before: [{ id: 1, text: "- hello" }, { id: 2, text: "hello" }]
// after:  [{ id: 2, text: "- hello" }]
```

Loose lists (`<li><p>…</p></li>`) are what cmark-gfm/GitHub emit for any list with blank lines between items, so this is common in feed HTML rather than exotic.

## Fix

A block element now contributes only **its own** text, leaving out any child that narrates itself. `img` is the exception — a nested image gets no paragraph of its own, so its enclosing block still speaks its alt text.

A loose item then has no text left, so the bullet and the spoken task-list state from #1439 ride along on the item's first block child instead:

```js
htmlToNarrationInput('<ul><li><p><input type="checkbox" checked> done</p></li></ul>')
// [{ id: 2, text: "- Done: done" }]
```

Paragraph ids are unchanged — they are still the document-order index of every block element, so they keep matching the client-side highlighting. The wrapper elements simply contribute no paragraph, exactly like `<ul>`/`<ol>` already did.

Two things fall out:

- A nested list inside an item is no longer repeated in its parent (`<li>Parent<ul><li>Child</li></ul></li>` → `- Parent`, `- Child`).
- Client-side, a `<blockquote>` wrapping a `<p>` no longer double-narrates either; that path has no markers to move, so the same skip fixes it outright.

List items now go through the same inline handling as paragraphs, so a link, inline code, or image inside one is narrated the way it is everywhere else instead of being flattened to raw `textContent`.

## The same bug in quotes and tables (#1445)

`<blockquote><p>…</p></blockquote>` — what every Markdown renderer emits for a `>` quote — was read as "Quote: x End quote." followed by a bare "x", and a `<p>` in a table cell was read twice as well.

Those markers **wrap** the text, so they can't migrate onto a child the way a bullet can: they'd have to split across the first and last child, and an attribution trailing the quote (`<footer>`, `<cite>`) would be read *before* the quote it follows, since a wrapper always precedes its children in document order. So the wrapper keeps speaking and the blocks inside it fall silent — keeping their paragraph index, contributing nothing, exactly like an empty element.

A quote now narrates its content structurally instead of flattening it to `textContent`:

```js
"<blockquote><p>a</p><p>b</p></blockquote>"          // "Quote: a\n\nb End quote."  (was "ab" + "a" + "b")
"<blockquote><ul><li>a</li></ul></blockquote>"       // "Quote: - a End quote."
"<blockquote><p>x</p><footer>— Author</footer></blockquote>" // "Quote: x\n\n— Author End quote."
```

The blank lines are the player's paragraph separator, and `buildAlignedNarration` maps each of them back to the quote — the invariant `paragraph-map.ts` documents. An empty quote no longer speaks a bare pair of markers.

Client-side the same silencing applies to tables (a cell's paragraph was repeated after the table) and to `<pre>` (a block inside one narrated the code the client deliberately skips). A blockquote is *not* in that set there — it speaks only its own text, so the paragraphs inside it are what narrate the quote.

**Figures are deliberately left out.** A figure speaks its alt text, not its subtree, so silencing a caption's paragraph would drop text that *complements* the alt rather than repeating it — and a figure wrapping a table would lose the table entirely.

## Testing

`pnpm typecheck` and `pnpm test:unit` (2663 passed) both pass. New cases cover the loose item, a multi-paragraph item, an item with both its own text and block children, nested lists, loose task lists, and inline images in items — server and client side.

### DIFF
--- a/src/lib/narration/client-paragraph-ids.ts
+++ b/src/lib/narration/client-paragraph-ids.ts
@@ -233,6 +233,25 @@ function processInlineContent(el: Element): string {
 }
 
 /**
+ * The first matching descendant an element speaks for itself — one no block in
+ * between has already claimed.
+ */
+function ownDescendant(el: Element, selector: string): Element | null {
+  for (const candidate of Array.from(el.querySelectorAll(selector))) {
+    let owner: Element | null = null;
+    for (let parent = candidate.parentElement; parent; parent = parent.parentElement) {
+      const tagName = parent.tagName.toLowerCase();
+      if (tagName !== "img" && BLOCK_ELEMENT_SET.has(tagName)) {
+        owner = parent;
+        break;
+      }
+    }
+    if (owner === el) return candidate;
+  }
+  return null;
+}
+
+/**
  * Everything inside an element, block children included, with the blocks kept
  * apart by a space.
  *
@@ -241,6 +260,14 @@ function processInlineContent(el: Element): string {
  * come from here — and `textContent` won't do, since it drops image alt text.
  */
 function subtreeNarrationText(el: Element): string {
+  return subtreeText(el).replace(/\s+/g, " ").trim();
+}
+
+/**
+ * The subtree's text, untrimmed — trimming at every level of the walk swallows
+ * the space in `<b>Total:</b><span> 42</span>`.
+ */
+function subtreeText(el: Element): string {
   let text = "";
 
   el.childNodes.forEach((node) => {
@@ -261,12 +288,12 @@ function subtreeNarrationText(el: Element): string {
       }
       return;
     }
-    const inner = subtreeNarrationText(childEl);
+    const inner = subtreeText(childEl);
     // Inline markup continues the run of text; a block starts a new one.
     text += BLOCK_ELEMENT_SET.has(childTag) ? ` ${inner} ` : inner;
   });
 
-  return text.replace(/\s+/g, " ").trim();
+  return text;
 }
 
 /**
@@ -320,31 +347,42 @@ function getElementNarrationText(el: Element): string {
     return processInlineContent(el);
   }
 
-  // Handle figures
+  // Handle figures. Only what the figure itself holds: an image inside a
+  // nested block (a table, a list) is spoken by that block instead — and a
+  // figure around one of those is not an image at all, so it announces none.
   if (tagName === "figure") {
-    const img = el.querySelector("img");
-    const figcaption = el.querySelector("figcaption");
-    const alt = img?.getAttribute("alt") || figcaption?.textContent?.trim();
-    if (alt) {
-      return `Image: ${alt}`;
+    const img = ownDescendant(el, "img");
+    if (!img) {
+      return processInlineContent(el);
     }
-    return "";
+    const figcaption = ownDescendant(el, "figcaption");
+    const alt = img.getAttribute("alt") || figcaption?.textContent?.trim();
+    return alt ? `Image: ${alt}` : "";
   }
 
   // Handle tables
   if (tagName === "table") {
-    // Extract table content in a readable format
-    const rows: string[] = [];
+    // Extract table content in a readable format. The caption counts: like the
+    // cells, the blocks inside it stay silent for the table's sake, so if the
+    // table doesn't say it, nothing does (issue #1445).
+    const parts: string[] = [];
+    const caption = el.querySelector("caption");
+    if (caption?.closest("table") === el) {
+      parts.push(subtreeNarrationText(caption));
+    }
     el.querySelectorAll("tr").forEach((tr) => {
+      // A nested table narrates its own rows — as a cell of this one.
+      if (tr.closest("table") !== el) return;
       const cells: string[] = [];
       tr.querySelectorAll("th, td").forEach((cell) => {
+        if (cell.closest("tr") !== tr) return;
         cells.push(subtreeNarrationText(cell));
       });
       if (cells.length > 0 && cells.some((c) => c.length > 0)) {
-        rows.push(cells.join(", "));
+        parts.push(cells.join(", "));
       }
     });
-    return rows.join(". ");
+    return parts.filter(Boolean).join(". ");
   }
 
   // Handle standalone images
--- a/src/lib/narration/client-paragraph-ids.ts
+++ b/src/lib/narration/client-paragraph-ids.ts
@@ -233,6 +233,43 @@ function processInlineContent(el: Element): string {
 }
 
 /**
+ * Everything inside an element, block children included, with the blocks kept
+ * apart by a space.
+ *
+ * `processInlineContent` stops at block children because each narrates itself;
+ * a table cell's blocks don't (the table speaks for them), so their text has to
+ * come from here — and `textContent` won't do, since it drops image alt text.
+ */
+function subtreeNarrationText(el: Element): string {
+  let text = "";
+
+  el.childNodes.forEach((node) => {
+    if (node.nodeType === Node.TEXT_NODE) {
+      text += node.textContent || "";
+      return;
+    }
+    if (node.nodeType !== Node.ELEMENT_NODE) return;
+
+    const childEl = node as Element;
+    const childTag = childEl.tagName.toLowerCase();
+    if (childTag === "img") {
+      const alt = childEl.getAttribute("alt");
+      if (alt && alt.trim()) {
+        // Padded: nothing here guarantees whitespace around an image the way
+        // the text around an inline one does (a caption butts right up to it).
+        text += ` Image: ${alt.trim()} `;
+      }
+      return;
+    }
+    const inner = subtreeNarrationText(childEl);
+    // Inline markup continues the run of text; a block starts a new one.
+    text += BLOCK_ELEMENT_SET.has(childTag) ? ` ${inner} ` : inner;
+  });
+
+  return text.replace(/\s+/g, " ").trim();
+}
+
+/**
  * Blocks whose narration already accounts for everything inside them: a table
  * speaks all of its cells, and a code block is deliberately not spoken at all.
  * A block nested in one of those has to stay silent, or its text comes back —
@@ -301,7 +338,7 @@ function getElementNarrationText(el: Element): string {
     el.querySelectorAll("tr").forEach((tr) => {
       const cells: string[] = [];
       tr.querySelectorAll("th, td").forEach((cell) => {
-        cells.push(cell.textContent?.trim() || "");
+        cells.push(subtreeNarrationText(cell));
       });
       if (cells.length > 0 && cells.some((c) => c.length > 0)) {
         rows.push(cells.join(", "));
--- a/src/lib/narration/client-paragraph-ids.ts
+++ b/src/lib/narration/client-paragraph-ids.ts
@@ -233,10 +233,25 @@ function processInlineContent(el: Element): string {
 }
 
 /**
+ * Blocks whose narration already accounts for everything inside them: a table
+ * speaks all of its cells, and a code block is deliberately not spoken at all.
+ * A block nested in one of those has to stay silent, or its text comes back —
+ * a second time for the table, and at all for the code block (issue #1445).
+ *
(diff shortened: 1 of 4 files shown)


brendanlong:
Pushed 6c5b35e0 after a review pass found the first version could **lose** a list item's marker rather than just de-duplicate it.

**Marker loss (the real bug).** Handing the marker to the item's *leading* block child dropped it whenever that child narrates nothing of its own. Worst case was a task list, where it silently undid #1439's done/not-done cue:

```js
'<ul><li><p><input type="checkbox" checked></p><p>Ship it</p></li></ul>'
// was: [{ id: 3, text: "Ship it" }]           ← "Done" gone
// now: [{ id: 3, text: "- Done: Ship it" }]
```

Same for a leading empty `<p>`, and for an item that opens with a sublist (`<ul>`/`<ol>` always narrate as empty, since their items mark themselves). The marker now goes to the first child that *actually speaks*, and reaches through `<div>`/`<section>` wrappers — those get no paragraph of their own, so they were a second way to lose it (`<li><div><p>x</p></div></li>` → `- x`).

**Anchor without an href.** Now that items go through the same inline handling as paragraphs, `<a id="fn1"></a>` was narrated `[link to null]` — and, counting as text, it made the item look like it had text of its own and took the marker with it. An anchor with no href is a link target, not a link, so it now speaks whatever text it wraps (nothing, here). Fixes the same noise in paragraphs, where it predates this PR. An empty `<code>` likewise no longer speaks a bare pair of backticks.

Six new cases cover these; `pnpm test:unit` is 2653 passing / 2 skipped.

Two review notes I deliberately left alone:

- An alt-less inline image in a list item now says "Image: image", because items use the paragraph rules. That matches what the server already does for `<p>` and for standalone images (asserted in `standalone-image-narration.test.ts:36`), so suppressing it for lists only would make the server inconsistent with itself.
- Client-side, a `<pre>` inside an `<li>`/`<blockquote>` now narrates nothing. The client deliberately skips code blocks; the wrapper reading the code out on the `<pre>`'s behalf was the bug that skip exists to prevent. Pinned with a test.

Claude

brendanlong:
Pushed 01627f22 after a second review pass. Silencing the blocks inside a quote or table only works if the wrapper really does say everything they would have — three ways it didn't, all **content loss**, which is worse than the duplication being fixed.

**A wrapper between the quote and its paragraphs made the whole quote silent.**

```js
'<blockquote><div class="quote-body"><p>Be excellent.</p></div><cite>Bill</cite></blockquote>'
// was: [{ id: 0, text: "Quote: Bill End quote." }]   ← the quote itself, gone
// now: [{ id: 0, text: "Quote: Be excellent.\n\nBill End quote." }]
```

The content walk treated `<div>`/`<section>` as inline, and inline text deliberately stops at block children (#1441) — so the paragraphs were dropped there *and* silenced as quote descendants. That's WordPress-style pull-quote markup and the sanitizer keeps those tags (`sanitize.rs:39`). Wrappers holding blocks are now transparent, as they already were for list markers; ones holding only text stay part of the run around them (`<blockquote>Hello <span>world</span> again` is still one paragraph).

**A table cell's image lost its alt text.** The cell's blocks are silent for the table's sake, but the table's text came from `textContent`, which has no alt text and no `[link to …]` fallbacks. Cells now narrate the way anything else does, on both sides:

```js
'<table><tr><td><figure><img alt="Sales chart"></figure></td><td><ul><li>a</li><li>b</li></ul></td></tr></table>'
// "Table: Image: Sales chart, - a - b End table."   (was "Table: , End table." + a lost figure)
```

**A `<figure>` inside a quote** was walked into as a generic wrapper, running alt and caption together ("Image: A cat**M**y cat"). It speaks its image, so the walk asks it instead of descending.

**Plus a robustness regression I found while measuring:** feed HTML has no nesting limit and the new walk is recursive, so ~1200 nested `<blockquote>`s overflowed the stack — where the old flatten-to-`textContent` never recursed at all. The walk is now depth-capped at 64 and flattens what's left, so the text is still narrated (3000-deep: 10ms, was a `RangeError`). Sanity-checked that the sanitizer does *not* cap nesting, so this was reachable.

Six new cases cover these. `pnpm test:unit`: 2670 passing, 0 skipped.

Claude

brendanlong:
Pushed b2b9c18e — third review pass. The theme again: **what a wrapper silences, it has to say.**

**`<table><caption>` was narrated nowhere.** The table silences every block inside it but its narration only walks `tr → th, td`, so a caption's text was spoken zero times. `<caption>` is on the sanitizer's allowlist (`sanitize.rs:49`), so this is real content. It's read first now, on both sides:

```js
"<table><caption><p>Table 1. Revenue by quarter</p></caption><tr><th>Q</th></tr><tr><td>1</td></tr></table>"
// "Table: Table 1. Revenue by quarter. Q. 1 End table."
```

**A `<figure>` inside a quote or cell lost its caption — and, wrapping a table, lost the table.** Treating it as self-contained meant `Image: <alt>` was *all* it ever contributed while its inner blocks stayed silenced. The walk descends into it instead, and a figure now announces an image only when it **directly holds one** — reaching through a nested block picked up an image that block already speaks (a new duplicate, now that cells speak their images), and a figure around a table isn't an image at all, so it narrates its own text rather than claiming to be one.

```js
"<blockquote><figure><table><tr><td>Cell A</td></tr></table><figcaption>Data</figcaption></figure></blockquote>"
// was: "Quote: Image: Data End quote."          ← table gone
// now: "Quote: Table: Cell A End table.\n\nData End quote."
```

My own test had asserted the caption loss as intended — corrected, and the server/client figure fixtures are now identical so a divergence would show up.

**Cell text lost spaces inside inline markup** (`<b>Name:</b><span> John</span>` → "Name:John"): both new cell paths trimmed at *every* level of the walk. They trim once, at the end. Same for text around a link, and the server now pads image alt text the way the client does.

**A nested table's cells were read three times** — the outer table's `querySelectorAll("tr")`/`("th, td")` are unscoped. Pre-existing, but this PR is precisely about "once, not once per wrapper" and touches that function. Both walks now skip rows/cells belonging to a table further in, so the inner table narrates once, as a cell of the outer one.

**And the depth-cap test didn't test the cap** — 500 nested quotes survive an uncapped walk, so it passed with the cap removed. Now 5000, which overflows without it.

`pnpm test:unit`: 2677 passing, 0 skipped.

Claude

brendanlong:
Pushed 775c070c — fourth review pass, which caught a **quadratic perf regression** I'd missed and more marker/caption loss.

**Perf: 200× slower than master on feed-controlled input.** `inheritedListMarker` asks "which of my blocks carries the marker?" — an answer that depends only on the `<li>` — but it was asked once per block *inside* that item, and answering walks the item (including a full table walk via `markerCarrier`). Measured, vs master:

| input | before | after | master |
|---|---|---|---|
| `<li>` + 3000 `<p>` (131 KB) | 1393 ms | **22 ms** | 6.5 ms |
| `<li>` + 10000 `<p>` | 15720 ms | **41 ms** | 26 ms |
| `<li><table 100×10>` + 800 `<p>` (**22 KB**) | 1263 ms | **4 ms** | 5.4 ms |

Memoized per item. A regression test narrates a 3000-paragraph item under a 2s bound.

**A `<figure>` around a table lost the list marker.** It announces no image, so it was a block that says nothing yet still claimed its descendants — `<li><figure><table>…</table></figure></li>` dropped the bullet. Like a `<div>`, it now owns nothing and the table inside carries the marker. That's the exact invariant the second review round established, escaping through a new hole.

**A figure's `<figcaption>` was narrated nowhere** at top level or inside a list item (the client used to say it, doubled — so a regression there). It's the image's description when there's no alt text and extra detail when there is, so it's spoken either way now, on both sides and in every context: `<figure>A cat<figcaption>My cat</figcaption></figure>` → `Image: A cat. My cat`, identical at top level, in a quote, and in a table cell.

**The client's paragraph walk still had both flaws the server's did** — trimming at every level (`<b>Name:</b><span> John</span>` → "Name:John") and no separation around image alt text. Fixed, but note whitespace collapses **horizontally only**: a blank line is how a `<br><br>` block becomes two spoken paragraphs, and collapsing `\s+` broke that (caught by the existing `paragraph-map` alignment tests — exactly the regression `paragraph-map.ts` was written to prevent).

**A nested table's cells were glued into "ABC" client-side**, which the test I'd added hid by using a single-cell inner table. Both nested-table tests now use a 2×2 inner table, and a table inside a cell reads as a table on both sides.

Also: the inline walks are depth-capped like the content walk — with list items narrating through them, deep wrapper nesting overflowed where master's `textContent` never recursed.

I verified server and client now agree on narrated text across every shape from all four review rounds. `pnpm test:unit`: 2684 passing, 0 skipped.

Filed #1451 for the one out-of-scope hole the review surfaced: `<dl>`/`<dt>`/`<dd>` and bare `<div>` text are narrated nowhere at all, on master too.

Claude

brendanlong:
Pushed 57b0e9d8. The last commit fixed the quadratic marker handoff and **introduced the same shape one line away**: `enclosingListItem` asks every block under a `<figure>` whether that figure says anything of its own, and answering walks the figure. No list required — a plain `<figure>` wrapping an article's paragraphs was enough:

| shape | before | after | master |
|---|---|---|---|
| `<figure>` + 3000 `<p>` | 1271 ms | **23 ms** | 8 ms |
| `<figure>` + 6000 `<p>` | 5234 ms | **23 ms** | 12 ms |
| `<figure>` + 12000 `<p>` | — | **51 ms** | — |

4× the blocks took 16× the time; now 2× size → 2× time. Memoized like the handoff, and I diffed the narration output across every shape from all five review rounds with and without the memo — byte-identical, so it's behavior-neutral.

The perf test I added last round **would have passed with this bug present** — it bounded a ~20ms case at 2s. It now bounds at 500ms and covers a wrapper with no list around it, which is exactly the shape that slipped through.

Also confirmed by the review, all by running it: both memos are sound (every entry point parses its own document, so no `Element` is ever shared across calls — verified by narrating doc A, doc B, then A again); the marker is right in all 14 figure shapes, never lost or duplicated; `figcaption` is said exactly once at top level, in a list item, in a quote and in a table cell, with server and client producing identical strings; the paragraph-map alignment invariant holds for `<br><br>` content; and `data-para-id` indices still match on both sides.

2684 tests pass.

Claude

PR #1449, opened 2026-07-29
Fix Threads saves: match /share/ URLs, guard redirects, log declines

brendanlong:
Fixes the reported bug where a saved Threads post had the post text as its excerpt but *"Log in or sign up for Threads…"* as its body, plus two related problems found along the way.

## The actual bug: `/share/` URLs were never matched

`https://www.threads.com/share/BAd-r4UCHz/` is what the **Threads app's share sheet produces** — so it's the form most likely to be pasted into the app — and `matchUrl` didn't recognize it. The plugin never ran, the save fell through to Readability, and Readability extracted the login wall.

The excerpt looked *right* the whole time because Readability takes excerpts from `og:description`, which on Threads is the post text. Only the body was wrong. That split is what made it look like a plugin failure rather than a plugin that never fired.

`/share/{code}` 302s to the canonical `/@handle/post/{code}`, and its id lives in its own space rather than the post shortcode's — the redirect guard below re-checks the URL we land on, so nothing has to trust that id.

I probed for the other forms rather than guessing again: `/p/{code}` is a 404 and `m.threads.com` doesn't resolve, so those aren't gaps. `/share/` works across all four host spellings.

Verified end-to-end with the exact reported URL:

```
matchUrl: true
plugin:   OK
  title:  I can barely enjoy Amleth anymore.
  author: SMBC Comics (@smbccomics)
full save -> siteName=Threads
```

## Redirect guard

A deleted or invalid post redirects to `https://www.threads.com/?error=invalid_post`, whose home page carries its own `og:description` — so we saved *"Join Threads to share ideas, ask questions, post random thoughts, find your people and more"* as the post body. Reproduced with `@zuck/post/C7lLcpjRhBz`.

Every Threads page has an `og:description`, so unlike LinkedIn there's no declared type to gate on. Confirming the page we actually landed on is still a post URL is this plugin's equivalent of LinkedIn's `@type` gate — the same "only claim a page you positively recognize" rule from `docs/DESIGN.md`, applied across redirects instead of within a page. It also covers the `/share/` id being opaque.

## Decline paths now log at info

Both plugins are built to degrade — private post, changed markup, blocked IP — but their decline paths logged at `debug` while prod runs at `info`. **This bug produced no production log line at all**, which is why I couldn't diagnose it from logs and guessed wrong about the cause. Raised to `info` so a silent fallback is visible.

---

Unit 2716, integration 728, typecheck, lint all pass.

### DIFF
--- a/src/server/plugins/linkedin.ts
+++ b/src/server/plugins/linkedin.ts
@@ -330,7 +330,11 @@ export const linkedInPlugin: UrlPlugin = {
 
         const content = renderLinkedInPost(page.html, page.finalUrl);
         if (!content) {
-          logger.debug("LinkedIn post page carried no usable body", { url: url.href });
+          // Info, not debug: this is the plugin silently declining a page it
+          // matched, which is exactly the failure a scraper degrades into when
+          // the markup changes or the post goes private. Prod runs at info, so
+          // at debug the degradation is invisible.
+          logger.info("LinkedIn post page carried no usable body", { url: url.href });
         }
         return content;
       },
--- a/src/server/plugins/threads.ts
+++ b/src/server/plugins/threads.ts
@@ -42,9 +42,17 @@ import { logger } from "@/lib/logger";
 const THREADS_HOSTS = new Set(["threads.com", "www.threads.com", "threads.net", "www.threads.net"]);
 
 /**
- * Parse a Threads post URL into its shortcode. Handles the canonical
- * `/@handle/post/{code}` form (with or without the SEO slug Threads appends),
- * and the `/t/{code}` short permalink its embeds use.
+ * Parse a URL that identifies a single Threads post, returning its id. Threads
+ * hands out three forms and a saved URL can be any of them:
+ *
+ * - `/@handle/post/{code}` — canonical, with or without the SEO slug appended
+ * - `/t/{code}` — short permalink, used by embeds
+ * - `/share/{code}` — what the app's share sheet produces, so it's the form
+ *   most likely to be pasted in; note its id is a share id in its own space,
+ *   not the post's shortcode
+ *
+ * The last two redirect to the canonical form, so `fetchContent` re-checks the
+ * URL it actually landed on rather than trusting the id here to be a post's.
  */
 export function parseThreadsPostCode(url: URL): string | null {
   if (!THREADS_HOSTS.has(url.hostname.toLowerCase())) {
@@ -63,8 +71,8 @@ export function parseThreadsPostCode(url: URL): string | null {
     return parts[2];
   }
 
-  // /t/{code} — the short permalink form used in embeds and share sheets.
-  if (parts.length === 2 && parts[0] === "t" && parts[1]) {
+  // /t/{code} and /share/{code} — both redirect to the canonical form.
+  if (parts.length === 2 && (parts[0] === "t" || parts[0] === "share") && parts[1]) {
     return parts[1];
   }
 
@@ -167,8 +175,8 @@ export const threadsPlugin: UrlPlugin = {
       siteName: "Threads",
 
       async fetchContent(url: URL): Promise<SavedArticleContent | null> {
-        // Fetch what the user gave us; `/t/{code}` redirects to the canonical
-        // `/@handle/post/{code}`, and both serve the same Open Graph tags.
+        // Fetch what the user gave us; the short forms redirect to the
+        // canonical `/@handle/post/{code}`, which serves the Open Graph tags.
         const page = await fetchPluginPage(url, "threads");
         if (!page) {
           return null;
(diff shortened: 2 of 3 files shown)


brendanlong:
Added image support (9976bd89).

## Why this wasn't already there, and what changed

`og:image` is **always** set on a Threads post — to the author's **avatar** when the post has no media. That's why the original plugin rendered no image at all: rendering it unconditionally captions every text post with a profile picture.

`twitter:card` is the discriminator, and it's a documented Twitter Card semantic rather than a guess at Meta's CDN path conventions (`t51.2885-19` vs `t51.82787-15` — which I did try first and rejected as an undocumented convention Meta can change). Measured over **14 live posts**:

| `twitter:card` | og:image is… | n |
|---|---|---|
| `summary_large_image` | real post media | 7/7 |
| `summary` | the author's avatar | 7/7 |

The rule is deliberately **one-directional**. 2 of those `summary` posts — a 20-image carousel and one single-image post — *do* have media that Threads simply doesn't advertise, so their images are missed. Missing an image is a much cheaper mistake than captioning every text post with someone's face.

Verified end to end on live posts, through the **real** read-path sanitizer:

| post | image rendered |
|---|---|
| comic (`summary_large_image`) | ✅ `t51.82787-15` |
| Instagram-crossposted photo | ✅ `t39.30808-6` |
| text-only | — (correct) |
| 20-image carousel | — (missed, as documented) |

`alt=""`: Threads exposes no alt text, and inventing one is worse than none.

Also sets `excerpt` to the post text so the list summary is unaffected by the image.

## Titles (answering the question, no change made)

The title is the post's **first line**, elided at 100 code points — so it can't run away, but 100 characters is still a big heading on mobile. Measured on real posts:

| body | first line | title |
|---|---|---|
| 77 | 34 | 34 — `I can barely enjoy Amleth anymore.` |
| 395 | 395 | **100 (elided)** |
| 498 | 319 | **100 (elided)** |
| 481 | 39 | 39 — `Stylized Action Comic-Book Illustration` |
| 254 | 80 | 80 |

So a post that opens with a paragraph rather than a short line gets a 100-character title, and for short posts the title is identical to the body — visible in the screenshot as the same sentence twice. Both are inherent to titleless posts and Bluesky behaves the same way (they share `socialPostTitle`), so I've left it alone rather than diverge one plugin. Happy to lower the cap or suppress the body's first line when it duplicates the title if you'd rather.

## Links

Already working — the orange `smbc-comics.com` link in the screenshot is our `<a>`, via `plainTextToHtml`. What *isn't* linkified is `@mentions` and `#hashtags`, which Threads renders as links. That's a genuine gap; I left it out because naive patterns false-positive on things like `#1` and `foo@bar`. Worth doing if you want it.

Unit 2721, integration 728, typecheck, lint all pass.

Claude

brendanlong:
Capped at 60 (a6a0527c). Shortening it surfaced something worth fixing, so this is slightly more than the one-line change.

## At 60, the cut point started landing badly

At 100 characters the cut rarely hit anywhere awkward. At 60, on real posts:

| before |
|---|
| `2) text-only posts perform *significantly* better than post…` ← mid-word |
| `…is the cemetery where Bob …` ← **space before the ellipsis** |
| `Opened up your Barnes & Noble NOOK just to find it's showin…` ← mid-word |

Three of six elided titles cut mid-word and one had a dangling space. So the cut now drops a trailing partial word:

| after | len |
|---|---|
| `2) text-only posts perform *significantly* better than…` | 55 |
| `Just outside of Orlando, Florida is the cemetery where Bob…` | 59 |
| `Opened up your Barnes & Noble NOOK just to find it's…` | 53 |
| `Some people are complaining that they are seeing videos and…` | 60 |
| `Why does LinkedIn persist in showing relative publish dates…` | 60 |
| `The cookout was an *immense* success, so many new people…` (Bluesky) | 57 |

## One subtlety worth recording

Backing up *unconditionally* is wrong, and my first attempt got it wrong — a test caught it. When the dropped character is itself a space, the text already ends on a complete word, and backing up throws away a whole word for nothing: `…seeing videos and` became `…seeing videos`. So it only backs up when the cut genuinely lands inside a word. A single opening word longer than the cap (a bare URL) still cuts mid-word rather than collapsing the title to a few characters.

## Scope

The cap is in `socialPostTitle`, so it applies to **Bluesky and LinkedIn** too — one convention for how a titleless post reads in the list, rather than three. Verified on live posts from all three.

The numeric cap is now asserted in exactly one place (`tests/unit/social-post.test.ts`); the per-plugin tests assert they route through the shared helper rather than restating the number, so changing it again touches one file instead of four.

Unit 2725, integration 728, typecheck, lint pass.

Note this doesn't address the *other* half of what the screenshot showed — for a short post the title is still identical to the body's first paragraph, so you see the same sentence twice. That needs suppressing the first paragraph when it duplicates the title, which means the stored body would no longer be the whole post. Happy to do it if you want, but it's a different tradeoff than truncation.

Claude

PR #1454, opened 2026-07-29
Render Threads post images, and cap synthesized post titles at 60 chars

brendanlong:
Follow-up to #1449. **These two commits were pushed to that branch after it merged**, so they never appeared in a PR — this is them, rebased onto current `master`.

## Render a Threads post's attached image

`og:image` is **always** set on a Threads post — to the author's **avatar** when the post has no media. That's why the original plugin rendered no image at all: doing it unconditionally captions every text post with a profile picture.

`twitter:card` is the discriminator, and it's a documented Twitter Card semantic rather than a guess at Meta's CDN path conventions (`t51.2885-19` vs `t51.82787-15` — which I tried first and rejected for exactly that reason). Measured over **14 live posts**:

| `twitter:card` | `og:image` is… | n |
|---|---|---|
| `summary_large_image` | real post media | 7/7 |
| `summary` | the author's avatar | 7/7 |

Deliberately **one-directional**: 2 of those `summary` posts — a 20-image carousel and one single-image post — do have media Threads simply doesn't advertise, so their images are missed. Missing an image is a much cheaper mistake than captioning every text post with someone's face.

Verified end to end on live posts through the **real** read-path sanitizer: a comic post (`t51.82787-15`) and an Instagram-crossposted photo (`t39.30808-6`) both render; a text post and the carousel render nothing. `alt=""` because Threads exposes no alt text and inventing one is worse than none. `excerpt` is set to the post text so the list summary is unaffected.

## Cap synthesized post titles at 60 characters

A post has no title, so the title is a heading invented from its opening words. At 100 characters a post whose first line is a whole paragraph filled the heading with body text.

Shortening it made a latent artifact frequent enough to matter — at 60, three of six real posts ended mid-word (`…find it's showin…`) and one put a space before the ellipsis (`…where Bob …`). So the cut now drops a trailing partial word, **but only when it lands inside one**: backing up unconditionally throws away a complete word whenever the dropped character happens to be a space (`…seeing videos and` → `…seeing videos`), which my first attempt did and a test caught. A single opening word longer than the cap (a bare URL) still cuts mid-word rather than collapsing the title.

Real titles now, across all three social plugins:

| len | title |
|---|---|
| 34 | `I can barely enjoy Amleth anymore.` |
| 55 | `2) text-only posts perform *significantly* better than…` |
| 59 | `Just outside of Orlando, Florida is the cemetery where Bob…` |
| 53 | `Opened up your Barnes & Noble NOOK just to find it's…` |
| 60 | `Some people are complaining that they are seeing videos and…` |
| 57 | `The cookout was an *immense* success, so many new people…` (Bluesky) |

The cap lives in `socialPostTitle`, so it applies to **Bluesky and LinkedIn** too — one convention for how a titleless post reads in a list rather than three. The numeric cap is asserted in exactly one place (`tests/unit/social-post.test.ts`); the per-plugin tests assert they route through the shared helper instead of restating the number, so changing it again touches one file rather than four.

## Not addressed

For a short post the title is still identical to the body's first paragraph, so the same sentence renders twice. Fixing that means suppressing the first paragraph when it duplicates the title, which makes the stored body no longer the complete post — a different tradeoff, left for a decision.

---

Verified against current `master` after the rebase: typecheck, unit (2776), integration (741), lint all pass.

### DIFF
--- a/src/server/plugins/linkedin.ts
+++ b/src/server/plugins/linkedin.ts
@@ -1,5 +1,5 @@
 import type { UrlPlugin, SavedArticleContent } from "./types";
-import { socialPostTitle } from "./social-post";
+import { socialPostImage, socialPostTitle } from "./social-post";
 import { fetchPluginPage } from "./fetch-page";
 import { Parser } from "htmlparser2";
 import { z } from "zod";
@@ -248,13 +248,11 @@ export function renderLinkedInPost(html: string, postUrl: string): SavedArticleC
 
   const parts = [plainTextToHtml(body)];
 
-  // The post's attached image (or the shared link's thumbnail). Plugin HTML is a
-  // body fragment, so the save path's `og:image` scrape never sees this page —
-  // inlining it is the only way the image survives the save.
+  // The post's attached image, or the shared link's thumbnail.
   const image = firstMatching(jsonLdImageUrl, post.image);
   const thumbnail = firstMatching(jsonLdHttpUrl, post.thumbnailUrl);
   if (image) {
-    parts.push(`<figure><img src="${escapeHtml(image)}" alt="" loading="lazy"></figure>`);
+    parts.push(socialPostImage(image));
   } else if (thumbnail) {
     // A video post. The JSON-LD `contentUrl` is a progressive MP4, but it is
     // access-gated — it 403s to anyone but a logged-in LinkedIn session, so a
--- a/src/server/plugins/social-post.ts
+++ b/src/server/plugins/social-post.ts
@@ -3,6 +3,8 @@
  * (Bluesky, LinkedIn, Threads).
  */
 
+import { escapeHtml } from "@/server/http/html";
+
 /**
  * Longest title we synthesize from a post's first line before eliding.
  *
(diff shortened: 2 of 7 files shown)


brendanlong:
Review pass found four real things; fixed in 8bc56bf3. The two worth reading:

## The floor was untested, and my test for it was fake

Mutating `MIN_ELIDED_TITLE_LENGTH` from 40 to **0** left all 2779 tests green. The test that looked like it covered the floor didn't: its input was `"https://example.com/" + "a".repeat(200)`, which contains **no space at all**, so `body.replace(/ [^ ]*$/, "")` was a no-op and `atBoundary === body`. The mid-word cut was kept because there was nothing to back up over, never because the floor rejected it. Same blind spot for `[...atBoundary].length` vs `.length` — the code-point accounting.

Two cases now cover it, and I mutation-tested each:

| mutant | before | after |
|---|---|---|
| `MIN_ELIDED_TITLE_LENGTH` 40 → 0 | survived | **2 tests fail** |
| `MIN_ELIDED_TITLE_LENGTH` 40 → 10 | survived | **2 tests fail** |
| `[...atBoundary].length` → `atBoundary.length` | survived | **1 test fails** |
| whitespace-collapse regex neutered | n/a | **1 test fails** |

The emoji case is the one that kills the UTF-16 mutant: 25 emoji are 25 characters of title but 50 code units, so counting units clears the 40 floor and collapses the title to 26 characters.

## The floor had a second bypass

`socialPostTitle("a" + " ".repeat(70) + "b")` returned **`"a…"`** — when the text ended in a whitespace run, the guard was skipped and the final `trimEnd()` deleted the run unguarded, directly contradicting the floor's own doc comment.

Fixed by collapsing whitespace runs up front, which is the right thing for a one-line heading regardless: that input now titles as `"a b"`, hard-wrapped opening lines and tabs normalize, and with every space one character wide the `trimEnd` can only ever drop one. Verified across the edge cases:

```
long internal space run    len=  3  "a b"
short word then huge word  len= 60  (mid-word cut, floor respected)
25 emoji then huge word    len= 60  (code-point accounting)
CJK, no spaces             len= 60
tab/newline whitespace     len= 13  "one two three"
all whitespace             len= 11  "Post by Sen"
```

## Also

- **Removed a dead condition.** The `!/\s$/.test(body)` half of the mid-word check changed no output — the reviewer brute-forced it over 300k random strings, 0 differences — and my comment's justification for it was simply wrong: when the text ends in whitespace, backing up discards *nothing*, not "a whole word".
- **`socialPostImage` extracted.** The image `<figure>` was byte-identical in both plugins and is the single reason an image survives a save, so it has to stay in sync.
- **Documented that LinkedIn passes a real `headline`** through this helper, not a synthesized first line — so the cap elides genuine LinkedIn headlines. That was a real cross-plugin effect the doc comment's framing hid.
- **Per-plugin title tests tightened** — they now assert the title is a *prefix of the input* plus an ellipsis. The previous `toBeLessThan(input.length)` assertions also passed for a 1-character title.

Not changed: grapheme clusters can still split (`🇺🇸` → lone regional indicator). Pre-existing, and the comment only promises no lone surrogate half, which remains true.

Unit 2779, integration 741, typecheck, lint, knip clean.

Claude

PR #1456, opened 2026-07-30
Resolve relative sibling-file references in saved gists (#1424)

### LINKED ISSUE #1424: Relative sibling-file references in saved gists aren't resolved

brendanlong:
Fixes #1424.

A gist's files are a flat list, but a Markdown file in one can still reference a sibling — `chart`. `processFileContent` passed `location: null` for gist files, so the reference fell through to the generic single-base absolutizer in `saved.ts`, which resolved it against `gist.github.com/{user}/{id}` — the gist's HTML page, so the image rendered broken (the failure mode #1422 fixed for repo files).

## What changed

Gist files now carry a location too (`GistFileLocation`, the file's own `raw_url`), and relative URLs resolve against `https://gist.githubusercontent.com/{user}/{id}/raw/`.

Two things worth calling out:

- **The revision sha in `raw_url` is dropped on purpose.** `raw_url` is `…/raw/{sha}/{filename}`, and that sha names a *blob* — the filename after it is then ignored. So resolving a sibling against it (`…/raw/{sha-of-notes.md}/chart.png`) returns **notes.md's bytes under the sibling's name**, silently the wrong file, which is worse than the broken link we started with. I verified both halves against the live host: the sha form ignores the filename, and the sha-less form resolves by filename and 404s when there's no such file.
- **The base is derived from `raw_url` rather than built from the gist id**, so an ownerless gist needs no guess at the user segment. Anything that isn't `gist.githubusercontent.com/{user}/{id}/raw/…` over https leaves the content untouched — every URL in the document resolves against this base, so an unexpected one must not become it.

Unlike a repo file — whose `href`s go to the github.com blob view, a page a reader can follow — a gist has no per-file page, so links resolve to the raw file too. A gist is flat, so a leading slash can only mean a sibling as well.

## Verification

- Full unit suite green (3025 passed); 10 new cases covering sibling images, sibling links, `./`- and `/`-prefixed forms, absolute URLs, same-document fragments, the dropped sha, and the two reject paths.
- End-to-end against the live API: fetched a real gist through the plugin, took a real `raw_url` from it, and confirmed the URL we produce for a sibling (`…/raw/script.js`) returns exactly that sibling file's content.

As the issue notes, this is low-impact in practice — gists are text-only, so a sibling image reference usually names a file that doesn't exist. The point is that a reference to a sibling that *does* exist now resolves to it.

### DIFF
--- a/src/server/plugins/github.ts
+++ b/src/server/plugins/github.ts
@@ -18,7 +18,7 @@ type GitHubUrlType =
   | { type: "blob"; owner: string; repo: string; ref: string; path: string }
   | { type: "raw"; owner: string; repo: string; ref: string; path: string };
 
-interface GistFile {
+export interface GistFile {
   filename: string;
   language: string | null;
   content: string;
@@ -26,11 +26,13 @@ interface GistFile {
   raw_url: string;
 }
 
-interface GistResponse {
+export interface GistResponse {
   id: string;
   description: string | null;
   owner?: { login: string } | null;
   files: Record<string, GistFile>;
+  /** Revisions, newest first. */
+  history?: { version: string }[];
   created_at: string;
   updated_at: string;
 }
@@ -344,19 +346,25 @@ function isMarkdownLanguage(language: string | null): boolean {
  * raw.githubusercontent.com) for the repo-root README, which we fetch without one.
  */
 export interface RepoFileLocation {
+  kind: "repo";
   owner: string;
   repo: string;
   ref?: string;
   path: string;
 }
 
-/**
- * A gist file's own raw URL (`raw_url` from the gists API), which is where the
- * relative URLs in it resolve from. We take GitHub's URL rather than building one
- * out of the gist id, so an ownerless gist needs no guess at the user segment.
- */
+/** A gist file's location, used to resolve the relative URLs in it. */
 export interface GistFileLocation {
+  kind: "gist";
+  /**
+   * The file's own `raw_url` from the gists API, which the `{user}/{id}` the
+   * gist is served under is read back out of. Taking GitHub's URL avoids
+   * depending on `owner`, which the API leaves null for an ownerless gist —
+   * and the user segment is load-bearing (a wrong one 404s).
+   */
   rawUrl: string;
+  /** The gist's current commit sha, when the response carried its history. */
+  revision?: string;
 }
 
 /** Where a rendered file came from, which is what its relative URLs resolve to. */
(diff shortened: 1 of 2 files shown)


brendanlong:
Pushed b90410b8 after an independent review pass. Three things it caught that were worth acting on:

**1. The base now pins to the gist's revision.** I originally dropped the sha from `raw_url` entirely. But there are *two* different shas in play: the one in `raw_url` names the file's own **blob** (filename after it is ignored — the trap this PR is about), while the gist's **commit** sha, `history[0].version`, does honor the filename. That's the sha GitHub's own links use, and it pins a saved article's URLs to the revision whose contents we rendered — so a later rename by the gist owner can't break an already-saved article. Only a 40-hex git object name is ever interpolated into the path.

**2. `FileLocation` is now tagged with `kind`** rather than discriminated on `"rawUrl" in location`. `ContentsResponse.download_url` is literally a repo file's raw URL, so anyone caching it on the repo location would have silently routed every repo file through the gist absolutizer, with no type error and no failing test.

**3. A claim in my comment was wrong.** I wrote that a gist has no per-file page. It does — GitHub rewrites a relative reference to a `#file-…` anchor on the gist page. Checking that turned up something better: GitHub does this for **`src`** too, so *its own* rendering of a relative sibling image in a gist is broken (an `` pointing at an HTML page). So resolving `src` to raw isn't just a workaround, it's strictly better than upstream. For `href` the anchor would be nicer than raw, but no URL base can express it — recorded as a tradeoff at `absolutizeGistUrls` and filed as #1459.

Also worth correcting from the issue text: **gists are not text-only.** The gist I verified against holds a 345 KB PNG, served `image/png`. So this fix is more useful than #1424 estimated — a sibling image reference in a gist can resolve to a real image.

Verification: full unit suite green (3033), typecheck clean, and end-to-end through `buildGistHtml` against that real gist — the resolved `src` returns `200 image/png`, 345110 bytes. The `raw_url` plumbing that was previously only exercised by hand-written literals now has `buildGistHtml`-level tests.

— Claude

PR #1457, opened 2026-07-30
Number narration against the tree a browser builds (#1453)

### LINKED ISSUE #1453: Narration paragraph numbering desyncs for content foster-parented out of a <table>
Found while reworking the narration walk for #1451 — pre-existing on `master`, and not narration's fault.

Content that isn't allowed inside a `<table>` (bare text, a stray `<div>`, an ``) is **foster-parented** by a browser's HTML parser: it gets hoisted out to just before the table. linkedom does not do this — it leaves the node where it was written. The sanitizer can't normalize it either (`lol_html` is a streaming rewriter, it never builds a tree), so both forms survive to narration.

That gives the server and the client different trees for the same HTML, and therefore different element numbering:

```
<p>x</p><table><div>h</div><tr><td>c</td></tr></table><p>y</p>

server (linkedom):  [p, table, div, p]
client (DOMParser): [p, div, table, p]
```

So a narration paragraph numbered 1 server-side (the table) highlights the client's `<div>`, and everything after it is off by one too. Only highlighting is affected — the spoken text is right on both sides.

Reachable on `master` today via `<table><p>`, `<table>` and friends; it is not specific to any one tag.

Possible fixes, none of them free:
- normalize the tree server-side after parsing (walk tables and hoist disallowed children out, mirroring the spec's foster-parenting rules);
- parse with a spec-compliant parser server-side;
- have the *client* number against HTML the server round-tripped (a bigger change to how content is delivered).

There is a skipped test pinning the expectation in `tests/unit/paragraph-id-consistency.test.ts` ("numbers the same elements for foster-parented content").

Reported by Claude.

brendanlong:
Fixes #1453.

## The bug

Narration is numbered on both sides of the wire: the server says which paragraph is which, the client stamps the `data-para-id`s those numbers name. Both call the same `narrationTargets` walk — but over **different trees**. The server parsed with linkedom, which is not a spec tree builder; the client parses with `DOMParser`, which is. So any markup the spec's tree construction rearranges numbered differently on the two sides, and every paragraph from the first rearrangement onward highlighted the wrong element.

All three shapes from the issue and its comments:

| input | linkedom (was) | browser |
| --- | --- | --- |
| `<p>x</p><table><div>h</div>…</table><p>y</p>` | `p, table, div, p` | `p, div, table, p` |
| `<p>a<table>…</table>c</p>` | `p, table, p` | `p, table, p` — but a *different* `p` |
| `<a href=out><a href=in></a></a>` | nested | siblings |

The sanitizer can't normalize any of it away first: lol_html is a streaming rewriter and never builds a tree.

## The fix

Normalize with **parse5** (the spec tree builder jsdom uses) and hand linkedom the serialized result. A serialized spec tree has every implied end tag explicit and every moved node already where a browser puts it, so linkedom reproduces the browser's tree — and the walk stays the shared, isomorphic code it is, rather than growing a server-side special case.

Costs a second parse (~2× the linkedom-only cost, ~50 ms for a 200 KB body), which is small against the LLM call it feeds. `htmlToPlainText` (AI summaries) gets the same tree, so a summary reads the article in the order the reader sees it.

Numbering changes for affected documents, so `NARRATION_FORMAT_VERSION` goes to 2 and cached paragraph maps regenerate instead of being served against today's `data-para-id`s.

## Tests

- The skipped `foster-parented content` case is unskipped, joined by eight more shapes a tree builder rearranges (hoisted `<div>`/``/text/`<p>`, `<table>` closing a `<p>`, nested `<a>`, misnested `<b>`, nested `<form>`). They're checked twice: same elements numbered, and same paragraphs narrated (the walk over the browser's tree is now a usable oracle).
- `tests/e2e/narration-highlighting.spec.ts` grows a table with a hoisted `<div>`, so the real server/browser pair is held to it, not just linkedom vs jsdom. Verified passing locally against a real server and Chromium.
- One shape can't use jsdom as its oracle: **jsdom appends foster-parented _text_ after the table** where parse5 and Chromium (checked directly) put it before. That's jsdom's bug, and it doesn't affect numbering (text isn't an element), so the server's output is pinned directly for that shape.

`pnpm test:unit` (3053 tests) and `pnpm typecheck` pass.


brendanlong:
Pushed 1eedc366 after an independent review pass. Two real problems it found, both now fixed and verified against Chromium:

**A stack overflow I introduced.** parse5's `serialize()` recurses per element, so markup nested a few thousand levels deep threw `RangeError` where linkedom's own parse copes — a 500 on `narration.generate` from ~26 KB of feed-controlled HTML, and on the LLM path an `errorAt` write that suppresses retries for an hour. Exactly the shape the walk already guards against with an explicit stack and `MAX_DEPTH`. The normalization now degrades to the plain linkedom parse (logged) rather than throwing: that costs the numbering agreement for one pathological document instead of the request. The threshold moves with the environment (10 000 levels overflows here, 2 400 in the reviewer's run), so the test asserts the contract — deep nesting still narrates — rather than a specific depth.

**`<iframe>` was still divergent** — the same bug class the PR is about, and normalizing can't fix it: parse5 correctly reads an embed's fallback markup as rawtext, but serializing writes it back out for linkedom to re-parse into elements. Chromium:

```
<iframe src="…"><p>Fallback</p></iframe><p>After</p>
  chromium: iframe, p        server (before): iframe, p, p   ← off by one
<iframe src="…"><p>Unclosed<p>Rest of article</p>
  chromium: iframe           ← the whole rest is rawtext; the server numbered all of it
```

`iframe` joins `NON_PROSE`, so both walks skip the subtree — which also stops the client speaking an embed's fallback markup aloud. `NARRATION_FORMAT_VERSION` was already going to 2, so this rides along for free.

Also from the review: the module doc's cost claim was wrong on the fallback paths (no LLM call there) and said nothing about the event loop, which narration — being pure JS — can't offload the way the sanitizer and extractor do. Corrected, and the root `CLAUDE.md` entry trimmed to the decision plus a pointer per the docs guidelines.

Findings I did **not** act on: the `expensive` rate limit for `narration.generate` (pre-existing, and a behaviour change beyond this issue), and a lint rule pinning `src/lib/narration/parse-html.ts` as server-only (linkedom was already there unguarded — worth doing, but separately).

The review verified independently that the numbering-change is otherwise complete: `NARRATION_FORMAT_VERSION = 2` covers every index-keyed cache, there's no path where the client renders HTML the server didn't number against, MathML/SVG foreign content and `<template>`/`<noscript>` round-trip correctly or are unreachable behind the sanitizer, and reverting the fix fails 11 of 55 assertions in the consistency suite.

Claude

PR #1458, opened 2026-07-30
Apply the tag allow-list to end tags too (#1455)

### LINKED ISSUE #1455: A stray </body> in entry HTML truncates narration and AI summaries server-side
Found while reviewing the narration rewrite (#1451). Pre-existing on `master`, and it silently drops content rather than mis-rendering it.

The sanitizer passes an unmatched `</body>` straight through (`lol_html` is a streaming rewriter, and `body` isn't in the allowlist, so there's no element to match the end tag against):

```
sanitizeEntryHtml("<p>a</p></body><p>b</p>") === "<p>a</p></body><p>b</p>"
```

Everything server-side then parses with `parseHTML(`…<body>${html}</body>…`)`, and linkedom treats the stray `</body>` as the end of the body — so it drops the rest of the document:

```
<p>a</p></body><p>b</p><p>c</p>

server (linkedom):  narrates ["a"]              ← b and c silently unspoken
client (DOMParser): narrates ["a", "b", "c"]    ← a browser ignores the stray end tag
```

Two consequences:

- **Narration stops mid-article** on the LLM path. Element ids still line up (both sides agree on the elements they *do* see), so highlighting isn't wrong — the speech is just missing.
- **AI summaries are truncated the same way**: `htmlToPlainText` (`src/lib/narration/html-to-narration-input.ts`), which feeds `summarizeEntry`, parses identically. A summary of a truncated article is wrong in a way nothing surfaces.

`</html>` behaves the same.

The cleanest fix is probably in the sanitizer: an end tag for a tag that is not allowlisted (and was never opened) should be dropped like the rest of the tag. That fixes every consumer at once, including the rendered HTML. Alternatively each parse site could neutralize the string first, which is more places to remember.

Note the same *family* as #1453 (server and client building different trees from the same HTML), but a different mechanism and a different fix.

Reported by Claude.

brendanlong:
Fixes #1455.

## The bug

`lol_html` is a streaming rewriter with no tree, so it only removes an end tag when it can pair it with an open element. An **unpaired** one passes straight through:

```
sanitizeEntryHtml("<p>a</p></body><p>b</p>") === "<p>a</p></body><p>b</p>"
```

A browser ignores a stray `</body>`, but linkedom ends the body there and drops the rest — and everything server-side parses with `parseHTML(`…<body>${html}</body>…`)`:

```
<p>a</p></body><p>b</p><p>c</p>

before →  narrates ["a"]              ← b and c silently unspoken
after  →  narrates ["a", "b", "c"]
```

So LLM narration stopped mid-article and **AI summaries silently summarized a truncated one** (`htmlToPlainText` feeds `summarizeEntry`). `</html>`/`</head>` behaved the same.

## The fix

Apply the tag allow-list to end tags too (`drop_disallowed_end_tags` in `native/sanitizer/core/src/sanitize.rs`), which fixes every consumer at once instead of asking each `parseHTML` site to remember.

It runs on the **rewriter's output**, not its input. That's the load-bearing choice: an end tag that reaches it with a non-allow-listed name is unpaired *by construction* (a paired one was already removed with its start tag, or with the whole subtree for `DROP_WITH_CONTENT`), so the pass needs no open element stack of its own and can safely drop even a `</script>`. It therefore has to run **before SVG re-insertion**, since SVG has its own allow-list (`</desc>` and friends).

Deleting a whole end-tag token leaves the tokenizer state identical either side of the cut, so the output re-tokenizes exactly as before — this can mangle content but can never fabricate markup. Two facts hold that up, both documented at the function and pinned by tests:

- quoted attribute values are skipped, so `title="</body>"` is left byte-identical;
- `iframe` is the only raw-text element the allow-list keeps, and it *is* allow-listed — so the end tag closing a raw-text run can never be deleted and spill its contents into the document as live markup.

## Performance

This is a per-read hot path, so the pass takes a fast exit: a SIMD `</` substring search decides "is there anything to do?" before any tag walking (a deliberate over-approximation of the walk, so skipping is safe), and the end-tag allow-list lookup is bucketed by first letter instead of scanning all ~90 entries. Interleaved A/B runs of `scripts/bench-sanitize.mts` against the pre-change binary:

| | base (A) | with fix | base (B) |
|---|---|---|---|
| 45 KB article | 0.497 ms | 0.494 ms | 0.488 ms |
| 379 KB article | 4.282 ms | 4.295 ms | 4.240 ms |

i.e. inside run-to-run noise (the naive first cut was +8%).

## Tests

- Rust: stray `</body>`/`</html>`/`</script>`/`</custom>` dropped, allow-listed strays kept, `title="</body>"` and `<iframe>` raw-text content untouched, no-markup-fabrication, idempotence, and `tag_allowed_bytes` held to `tag_allowed` over every allow-listed tag (so a tag that fell out of the bucketed index can't silently lose its end tags).
- TS (`tests/unit/sanitize-entry-html.test.ts`): the same at the pipeline level, plus the user-visible symptom — narration and `htmlToPlainText` now see the whole article.
- Full `pnpm test:unit`, `typecheck`, `lint` green.


brendanlong:
## Security review found two XSS bypasses in the first cut; a property test found two more

An independent review of this PR broke the invariant it rested on. All four holes are now closed in 09f8c349, but the failure mode is worth recording, because the original argument sounded airtight:

> Deleting a complete end-tag token leaves the tokenizer state identical either side of the cut, so the rest re-tokenizes exactly as before.

That's true, and **not sufficient**. Each of the following was a working `` injection — verified against a real tree builder, and inert on `master` for the same input:

1. **A lone `<` immediately before the cut.** The tokenizer emits `<` as a *character* when the next byte can't start a tag name — which is exactly the case when the next byte is our `<`. Delete our bytes and it's promoted to a real tag-open: `<p>hi</p><</body>img src=x onerror=alert(1)>` went from inert text to a live ``. Now refused, as is a cut in front of an unterminated character reference (`&am</body>p;` → `&amp;`; text-only, but the cut is meant to be invisible).

2. **Attributes on the end tag.** lol_html passes an unpaired end tag through byte-for-byte *including its attributes*, so a feed can park markup in one where the main pass never reaches it — `</body " a="x>">` is a single ignored token. Only bare `</name>` is cut now, which takes attribute scanning out of the deletion path altogether. Cost: an attributed stray `</body>` is left in place. The reported bug is the bare form.

3. **`scan_to_tag_end` wasn't the spec's tag scan.** A quote only opens an attribute value out of *before-attribute-value* state (after `=`); in name position it's a parse error that becomes part of the name. Treating every quote as a delimiter paired the stray quote with the real value's opening quote and ended the tag *inside* that value — which is what made (2) reachable. It also silently truncated whole documents (`</body class="x><p>two</p>…` → cut to EOF) and mislocates MathJax/SVG ranges, so the fix went into the scanner.

4. **Foreign content.** Whether a start tag opens a raw-text run is the one question the tokenizer asks the *tree*: `<iframe>` is raw text in HTML and an ordinary element in SVG. So cutting a `</svg>` re-namespaced everything after it and could turn an embed's (unsanitized — the rewriter saw it as text) fallback content into live markup. Cuts now happen only in HTML content, tracked by name-matched `svg`/`math` depth.

**The testing lesson is the durable part.** Five hand-written cases passed while all four injections were live — including one named `dropping_a_stray_end_tag_cannot_fabricate_markup`. The invariant is now held by `end_tag_drop_never_adds_markup`: generate fragments from tokenizer-significant pieces, tree-build input and output with html5ever, and assert the pass adds **no element and no attribute** (plus idempotence, plus that some inputs were actually cut so it can't pass vacuously). It found holes 3 and 4 within seconds. 8M cases across four seeds are now clean.

Also from the review: docs stated the invariants as settled fact and repeated the mechanism in three places — `src/server/html/CLAUDE.md` now says the one thing a reader outside the function needs (this is the sharpest edge in the sanitizer; change it only against the property test) and points at the function for the rest.

Bench re-run after the scanner rewrite (`scan_to_tag_end` is on the hot path via the MathJax/SVG locator, not just the new pass) — indistinguishable from the A/A control:

| | base (A) | fixed | base (B) |
|---|---|---|---|
| 45 KB | 0.496 ms | 0.494 ms | 0.494 ms |
| 379 KB | 4.264 ms | 4.265 ms | 4.224 ms |

One thing I checked and am *not* filing: the fuzz also flagged a non-idempotence where unwrapping `<svg>` moves its CDATA out of foreign content. It's an artifact of exercising `sanitize_html_pass` in isolation — the shipped pipeline extracts SVG to a placeholder first, so `<svg>` never reaches lol_html and the full pipeline is idempotent on those inputs.

Not yet done: `pnpm test:integration` / `test:e2e` (no DB in this environment). Unit (3021), typecheck, lint, and 95 Rust tests are green.

— Claude

Issue #1460, opened 2026-07-30
A GitHub API failure silently degrades a gist/repo save into a scrape of the HTML page


brendanlong:
Found while testing #1424 in production. The user saved `https://gist.github.com/akiyan/e5f33118322aade40775432ed6b22363` and got GitHub *page chrome* in the article body — "You must be signed in to star a gist", "…use it in GitHub Desktop" — instead of the gist's rendered README.

Cause, from the app machine's log at the moment of the save:

```
02:51:08Z warn "Failed to fetch gist" {"gistId":"e5f33118322aade40775432ed6b22363","status":401}
```

The production `GITHUB_API_TOKEN` is invalid, so `fetchGist` got a 401, returned null, and `fetchSavedArticleContent` (`src/server/services/saved.ts:917`) treated null as "no plugin content" and fell back to a normal fetch + readability of `gist.github.com`. The result *looks* like a Lion Reader bug: page sidebar text, an avatar, and a stray `article:author` of `262588213843476` (GitHub's Facebook page id, via `og:` metadata).

Two problems, only the first of which is a config issue:

**1. The failure was invisible.** A bad token degrades *every* gist and repo save to a page scrape, indefinitely, at `warn` level with no operator signal. Being fixed in the PR for this issue: a 401 now logs at `error` (so it reaches Sentry) and says which env var to check.

**2. Falling back to a page scrape is arguably wrong here.** For a GitHub URL we *know* the scrape is bad — the page is chrome around a rendered README, and readability keeps the chrome. `saved.ts` already refuses to degrade on a rate limit (`errors.upstreamRateLimited`, with the comment "the same site would rate limit us again"); a 401 has the same character (retrying the scrape can't produce the real content). Options:

- Surface the failure to the user instead of saving a scraped page, like the rate-limit path does.
- Retry the API call **unauthenticated** — the gists and contents APIs serve public content without a token at 60 req/hr/IP, so a bad token would degrade to "works, slower" instead of "works, wrong". This would have made the user's save correct despite the broken token.

The second is attractive but changes behavior under load, so it needs a decision rather than a drive-by.

Not filing the token itself as an issue — that's an ops action (rotate `GITHUB_API_TOKEN`).

Reported by Claude.

brendanlong:
Better evidence, and it changes which option is right.

[...]

PR #1461, opened 2026-07-30
Stop dumping a gist's binary files as base64

brendanlong:
Found while verifying #1424 against a real gist in production.

## The bug

A gist can hold binary files (they're git repos — the issue text for #1424 was wrong that gists are text-only), and the gists API returns a binary file's bytes **base64-encoded** in `content`. `processFileContent` sends anything that isn't Markdown or HTML to the code-block fallback, so a gist containing a PNG rendered as screenfuls of base64:

```
<h2>…mixline.png</h2>
<pre><code>iVBORw0KGgoAAAANSUhEUgAAFj0AAAioCAIAAACpBzGhAAEAAElEQVR42uzd…
```

468 KB of it for `gist.github.com/akiyan/e5f33118322aade40775432ed6b22363` — comfortably under `maxSavedArticleSizeBytes`, so it saved without complaint.

## The fix

Classify from GitHub's `type` field: `image/*` → an ``, a named binary type (`audio/*`, `video/*`, `font/*`, `application/{pdf,zip,gzip,mp4,octet-stream}`) → a link, everything else unchanged. That last part is deliberate — an unrecognized `application/*` is far more often source code (`application/x-python`, `application/json`, `application/x-perl` all show up in real gists) than a binary, and rendering those as text is the better guess.

Both new forms are emitted as **relative** references and run through `absolutizeGistUrls`, so there's one URL builder, and they inherit the revision pinning from #1424 for free.

Same gist, before → after: **468464 bytes → 933 bytes**, with the PNG as a real image.

```html
<h2>…mixline.png</h2>
<img src="https://gist.githubusercontent.com/akiyan/…/raw/cbc18f31…/…mixline.png" alt="…mixline.png">
<h2>README.md</h2>
<p><a href="…">Open the full-resolution PNG</a></p>
<p><img src="…" alt="Codec timeline" /></p>
```

## Also: a 401 from the GitHub API now logs at error level

This is how the bug was found, and it's the more serious problem. Production's `GITHUB_API_TOKEN` is invalid, so `fetchGist` got a 401, returned null, and `saved.ts` fell back to scraping `gist.github.com` — giving the user an article body full of GitHub page chrome ("You must be signed in to star a gist"). At `warn` level that had been happening invisibly for every gist and repo save. A 401 is an operator problem, so it now logs at `error` (→ Sentry) and names the env var. The deeper question — whether a GitHub API failure should degrade to a page scrape at all — is #1460.

Requires rotating `GITHUB_API_TOKEN` to actually fix production; that's not something this PR can do.

## Verification

Full unit suite green (3084), typecheck and eslint clean, 6 new cases (image, other-binary, unrecognized-`application/*`-stays-text, lone image gist, requested image file, filename escaping). Verified end-to-end against the live gist above.

### DIFF
--- a/src/server/plugins/github.ts
+++ b/src/server/plugins/github.ts
@@ -182,14 +182,14 @@ export function normalizeFilenameForFragment(filename: string): string {
 /**
  * Build headers for GitHub API requests.
  */
-function getApiHeaders(): HeadersInit {
+function getApiHeaders(authenticated = true): HeadersInit {
   const headers: HeadersInit = {
     Accept: "application/vnd.github+json",
     "User-Agent": USER_AGENT,
     "X-GitHub-Api-Version": "2022-11-28",
   };
 
-  if (githubConfig.apiToken) {
+  if (authenticated && githubConfig.apiToken) {
     headers.Authorization = `Bearer ${githubConfig.apiToken}`;
   }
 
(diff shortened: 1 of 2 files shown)


brendanlong:
Added 447b20cd: **retry the GitHub API unauthenticated on a 401** — closes #1460.

Brendan's counter-example (`github.com/brendanlong/lion-reader/blob/master/CLAUDE.md` renders perfectly) turned out to be the decisive clue rather than a refutation. The logs show that page 401'd too:

```
03:07:41Z warn "Failed to fetch repo content" {path: "README",    status: 401}
03:09:01Z warn "Failed to fetch repo content" {path: "CLAUDE.md", status: 401}
```

It rendered anyway because `fetchRawContent` reads `raw.githubusercontent.com` with **no `Authorization` header**, so the `blob` fallback sailed through the broken token. Gists and repo roots have no such fallback, so they returned null and got scraped. That made the decision in #1460 easy: the "retry unauthenticated" pattern was already in the file and load-bearing — the other two paths just weren't getting it.

So `fetchGitHubApi` now drops the header and retries once, but **only on a 401**. A 403/429 is a rate limit and the unauthenticated limit is the *lower* of the two (verified just now: 5000/hr authenticated vs 60/hr per IP), so retrying there would spend the shared budget to fail again.

Verified with a deliberately invalid `GITHUB_API_TOKEN` against live GitHub — both previously-broken shapes now return real content, and the error still fires once so the operator signal isn't lost:

```
[ERROR] GitHub API rejected our credentials; check GITHUB_API_TOKEN {"url":"https://api.github.com/gists/e5f33118…"}
  -> OK title="SEGA-CD FMV mixed codec/HUD timeline…" (933 bytes)
[ERROR] GitHub API rejected our credentials; check GITHUB_API_TOKEN {"url":"…/repos/brendanlong/lion-reader/contents/README.md"}
  -> OK title="Lion Reader" (10737 bytes)
```

That second title is the tell: `Lion Reader` is the README's H1 via the plugin, where the scrape stored `GitHub - brendanlong/lion-reader: An AI-native, all-in-one reader…`.

The retry predicate is unit-tested (`shouldRetryUnauthenticated`); the fetch wrapper around it isn't, since it's a network call and this repo doesn't mock internals. Suite: 3088 green.

One correction to my earlier comment on #1460: I'd speculated the 401 might be a fine-grained PAT being rejected by the gists API. That's wrong — I tested a `github_pat_` token against `GET /gists/{id}` and it returns 200. Fine-grained tokens read public gists fine.

— Claude

PR #1463, opened 2026-07-31
Link a gist's sibling files to the gist page, not raw (#1459)

### LINKED ISSUE #1459: Relative links in saved gists could point at the gist page anchor, not raw
Noticed while fixing #1424 (relative sibling-file references in saved gists).

#1424 resolves a relative reference in a gist file — `src` and `href` alike — against `gist.githubusercontent.com/{user}/{id}/raw/{revision}/`. For `src` that's the right answer and better than GitHub's own (see below). For `href` it means a reader who clicks `[Setup](setup.md)` lands on raw text: `text/plain`, `nosniff`, CSP `sandbox`, no rendering.

GitHub itself does something nicer. Verified against a live gist (`akiyan/e5f33118322aade40775432ed6b22363`), whose README source is:

```markdown
[Open the full-resolution PNG](./…mixline.png)

![Codec timeline](./…mixline.png)
```

Inside the rendered `markdown-body`, GitHub emits:

```html
<a href="/akiyan/e5f33118322aade40775432ed6b22363#file-…-mixline-png">Open the full-resolution PNG</a>
<img src="/akiyan/e5f33118322aade40775432ed6b22363#file-…-mixline-png" alt="Codec timeline">
```

So a relative reference becomes a `#file-…` anchor on the gist page. Worth noting the `` gets the same treatment — meaning **GitHub's own rendering of a relative sibling image in a gist is broken** (an `` pointing at an HTML page). That's exactly why we resolve `src` to raw instead and shouldn't follow GitHub there.

For links, though, the anchor is a page a reader can follow, which is the same reasoning behind sending repo-file `href`s to the github.com blob view rather than raw (`absolutizeGitHubUrls`).

Both halves already exist in `src/server/plugins/github.ts`: `normalizeFilenameForFragment` builds that fragment, and `parseGitHubUrl` + `parseGistFilenameFromFragment` parse it back — so such a link would even be re-savable in Lion Reader.

What blocks it: `absolutizeUrls` resolves against **URL bases**, and no base can turn `setup.md` into `…/{id}#file-setup-md`. It needs either a per-attribute transform hook in the absolutizer or a second pass keyed on the gist's file list, and the mapping is only correct for references that name a file the gist actually has. That's more machinery than #1424 (explicitly low priority) justified, so it was left as a recorded tradeoff at `absolutizeGistUrls`.

Low priority: raw serves the file's content, so the link works — it's just unrendered.

Reported by Claude.

brendanlong:
Fixes #1459.

## What

A relative link in a saved gist (`[Setup](setup.md)`) resolved to `gist.githubusercontent.com/…/raw/…`, so a reader who clicked it landed on `text/plain` under `nosniff` and a CSP `sandbox` — the file's source, unrendered. GitHub instead turns such a link into `#file-setup-md` on the gist page, which is a page a reader can follow, and it round-trips: `parseGitHubUrl` parses that anchor back, so the link is re-savable in Lion Reader.

Embedded `src` deliberately stays on raw. GitHub gives an `` the same `#file-…` anchor, which means **its own** rendering of a relative sibling image in a gist is broken (an `` pointing at an HTML page), as #1459 notes.

## How

- `absolutizeUrls` grows a `rewriteHref` hook — last say over each already-resolved `href`, for sources where a link's real destination isn't expressible as a URL base at all (which is what blocked this in #1424).
- `absolutizeGistUrls` uses it via `gistSiblingLinkRewriter`, which maps a URL under the gist's raw base back to a filename and swaps it for the gist page anchor — only for a reference naming a file the gist actually has, and never for one carrying its own query/fragment (which the anchor would silently drop). The file list comes from `buildGistHtml`.
- The synthesized link for a **binary** file keeps pointing at raw: that link stands in for the file itself, so it must reach the bytes.

## Fragment normalization was wrong

Getting the anchor right meant fixing `normalizeFilenameForFragment`. It's Rails' `parameterize`, and each rule is now verified against live gist pages:

| filename | before | now / GitHub |
|---|---|---|
| `agent_patch.diff` | `agent-patch-diff` | `agent_patch-diff` |
| `.md` | `-md` | `md` |
| `datenschutzerklärung.md` | `datenschutzerkl-rung-md` | `datenschutzerklarung-md` |

Underscores are everywhere in gist filenames, so this also fixes a pre-existing bug in the other direction: saving a `#file-…` URL copied from GitHub for such a file matched nothing and silently rendered the whole gist instead.

## Verification

Beyond the unit tests, ran the real plugin against the live gist from the issue (`akiyan/e5f33118322aade40775432ed6b22363`). Our output for the README's relative reference:

```
href="https://gist.github.com/akiyan/…#file-20260730-010800-932668_sonic-jam-op_…_mixline-png"
src="https://gist.githubusercontent.com/akiyan/…/raw/cbc18f3…/20260730-010800-932668_sonic-jam-op_…_mixline.png"
```

The `href` is byte-for-byte the href GitHub emits on that page, and matches the `id=` of the file's section there; the `src` stays on raw, where the image actually renders.

Unit tests, typecheck and lint all pass.

### DIFF
--- a/src/server/feed/content-cleaner.ts
+++ b/src/server/feed/content-cleaner.ts
@@ -40,14 +40,14 @@ const decoder = new TextDecoder();
  *   `src/server/plugins/github.ts`. The two bases travel together so a caller
  *   can't point media at another origin and leave its root base behind on this
  *   one, which is the bug this option exists to prevent.
- * @param options.rewriteHref - Last say over each `href` (only), applied to the
- *   already-resolved value. For sources where a link's real destination isn't
- *   expressible as a base at all — a gist's sibling file is a `#file-…` anchor
- *   on the gist page, see `absolutizeGistUrls` in
+ * @param options.rewriteHref - Last say over each `href` attribute (only),
+ *   applied to the already-resolved value. For sources where a link's real
+ *   destination isn't expressible as a base at all — a gist's sibling file is a
+ *   `#file-…` anchor on the gist page, see `absolutizeGistUrls` in
  *   `src/server/plugins/github.ts`. It receives whatever resolution produced,
- *   which for the values resolution passes through (same-document fragments,
- *   `data:`) is still the relative original, so a hook must recognize the
- *   inputs it acts on rather than assume an absolute URL.
+ *   including the values resolution passes through unchanged (same-document
+ *   fragments, `data:`), so a hook must recognize the inputs it acts on rather
+ *   than assume an absolute URL.
  * @returns HTML with all relative URLs converted to absolute
  */
 export function absolutizeUrls(
--- a/src/server/feed/content-cleaner.ts
+++ b/src/server/feed/content-cleaner.ts
@@ -40,6 +40,14 @@ const decoder = new TextDecoder();
  *   `src/server/plugins/github.ts`. The two bases travel together so a caller
  *   can't point media at another origin and leave its root base behind on this
  *   one, which is the bug this option exists to prevent.
+ * @param options.rewriteHref - Last say over each `href` (only), applied to the
+ *   already-resolved value. For sources where a link's real destination isn't
+ *   expressible as a base at all — a gist's sibling file is a `#file-…` anchor
+ *   on the gist page, see `absolutizeGistUrls` in
+ *   `src/server/plugins/github.ts`. It receives whatever resolution produced,
+ *   which for the values resolution passes through (same-document fragments,
+ *   `data:`) is still the relative original, so a hook must recognize the
+ *   inputs it acts on rather than assume an absolute URL.
  * @returns HTML with all relative URLs converted to absolute
  */
 export function absolutizeUrls(
@@ -48,6 +56,7 @@ export function absolutizeUrls(
   options: {
     rootBaseUrl?: string;
     media?: { baseUrl: string; rootBaseUrl?: string };
+    rewriteHref?: (url: string) => string;
   } = {}
 ): string {
   try {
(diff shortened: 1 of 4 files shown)


brendanlong:
Pushed review follow-ups (d816a32):

- **The anchor is now revision-pinned.** I'd claimed in a test comment that the gist page has no per-revision form — that's false, and I verified it: `https://gist.github.com/{user}/{id}/{sha}` returns 200 and carries the same `id="file-…"` anchors. It mattered behaviorally, because `src` pins to the revision we rendered while the link would have drifted to whatever the gist became. Both URLs now derive from one validated sha. `parseGitHubUrl` still parses the pinned anchor back (tested).
- **Rejected resolved paths containing a separator.** `sub/dir.md` names no file in a flat gist, but normalizing dashed the slash out and could land on an unrelated `sub-dir.md`.
- **Corrected three comments** that claimed more than is true: `data:` URLs *do* reach the rewrite hook (only same-document fragments arrive unresolved), the hook fires on every `href` attribute rather than on links specifically, and our transliteration both over- and under-shoots Rails' Latin-1 table rather than being a subset of it.

Deliberately not changed: a link carrying its own `?`/`#` still keeps the raw URL. Mapping it would silently drop the part the author wrote, and neither target honors it.

Live check against the gist from the issue now emits `…/e5f33118322aade40775432ed6b22363/cbc18f3161df2b2dd22f3c4944d67cebb97ae544#file-…_mixline-png`, which resolves to that file's section on the pinned page.

3109 unit tests, typecheck and lint pass.

— Claude

Issue #825, opened 2026-04-20
TOCTOU race in unlinkProvider could lock user out


brendanlong:
## Problem

In `src/server/trpc/routers/auth.ts` (around line 1225), the `unlinkProvider` mutation has a time-of-check-time-of-use race condition:

1. Counts linked OAuth accounts
2. Checks if unlinking is safe (`!hasPassword && linkedAccountsCount <= 1`)
3. Deletes the OAuth account

Between steps 1 and 3, a concurrent request from another tab/session could unlink a different provider.

**Example:** User has Google and Apple linked, no password. Two tabs simultaneously unlink Google and Apple. Both count `linkedAccountsCount = 2`, both pass the check, both delete. User ends up with zero auth methods — locked out.

## Suggested fix

Wrap the check-and-delete in a transaction with a row lock:

```typescript
await ctx.db.transaction(async (tx) => {
  // Lock the user row to prevent concurrent unlinks
  await tx.execute(sql`SELECT 1 FROM users WHERE id = ${userId} FOR UPDATE`);
  
  const linkedAccountsResult = await tx
    .select({ count: sql<number>`count(*)::int` })
    .from(oauthAccounts)
    .where(eq(oauthAccounts.userId, userId));
  
  if (!hasPassword && linkedAccountsResult[0]?.count <= 1) {
    throw errors.cannotUnlinkOnlyAuth();
  }
  
  await tx.delete(oauthAccounts)
    .where(and(eq(oauthAccounts.userId, userId), eq(oauthAccounts.provider, provider)));
});
```

**Severity:** Low — requires precise concurrent timing from the same user.

Found during code review in #823.

— Claude

PR #1483, opened 2026-08-03
Extract OAuth account linking into a service (#1467)

### LINKED ISSUE #1467: Extract OAuth account-linking into a service (triplicated in auth router)
`linkGoogle` (~`src/server/trpc/routers/auth.ts:994`), `linkApple` (~`:1100`) and `linkDiscord` (~`:1198`) each inline the same ~50-line sequence — check existing link for provider → update tokens or throw `oauthAlreadyLinked` → check the account is bound to another user → throw → `db.insert(oauthAccounts)` — with direct DB writes in the router. There is no oauth-accounts service (`services/users.ts` only has `deleteUser`).

This is the same services-layer-erosion class that #956 / #869 fixed for `entries` and `saved`; the register and OAuth-callback paths already do it right (they route through `createUser`/`createSession` and the shared `handleOAuthCallback` helper).

**Proposal:** extract `linkOAuthAccount` / `unlinkOAuthAccount` into `src/server/services/` and collapse the three procedures onto them.

Found during the August 2026 maintainability review.

— Claude

brendanlong:
Fixes #1467.

`linkGoogle` / `linkApple` / `linkDiscord` each inlined the same ~50-line sequence — check the existing link for the provider → update tokens or throw `oauthAlreadyLinked` → check the account is bound to another user → throw → `db.insert(oauthAccounts)` — with direct DB writes in the router, and `unlinkProvider` carried its whole locking transaction inline. All four now go through `src/server/services/oauth-accounts.ts` (`linkOAuthAccount` / `unlinkOAuthAccount`). Net −160 lines in the router.

## Behavior changes that fall out of sharing one implementation

- **Re-linking the same Apple/Discord account refreshes the stored tokens** instead of throwing `oauthAlreadyLinked`. That's what Google already did (it's how incremental authorization for the Docs scope lands), and the erroring version was just an artifact of Apple/Discord checking for *any* existing link before they had the `providerAccountId` to compare against. Linking a **different** account for an already-linked provider is still refused.
  - Consequence: Apple/Discord now validate the OAuth callback before the already-linked check, so the authorization code is consumed before that error. Harmless — the code is single-use and the flow ends either way.
- **"Already linked to another user" is now enforced by the `(provider, provider_account_id)` unique constraint** (`onConflictDoNothing` + empty `returning` → throw) rather than a check-then-insert, closing the window where two concurrent links of the same provider account both pass the SELECT. The conflict deliberately does **not** update, so a losing request can never take over another user's link.

## Tests

New `tests/integration/oauth-account-link.test.ts` covers create, incremental-auth refresh, scope preservation for providers that don't report scopes, both refusal cases, and the concurrent-link race. The existing `oauth-unlink.test.ts` (#825 locking) still passes unchanged through the router → service path.

`pnpm typecheck`, `pnpm lint`, `pnpm knip` and the full integration suite pass locally.


brendanlong:
Ran an independent review pass. It found no hijack path through the new `onConflictDoNothing` insert (the conflict deliberately does not update, and the `uq_oauth_accounts_provider_account` constraint it targets matches exactly), and confirmed `unlinkOAuthAccount` is a line-for-line code move. Addressed everything else in c2b9ac2:

**Real bug it caught** — the update path was nulling `refresh_token` whenever the provider omitted one. A re-consent usually omits it, so this drops access the user never revoked. `refreshGoogleToken` already guards against exactly this (`...(tokens.refresh_token ? {refreshToken} : {})`); the service now does too. `expiresAt` stays unconditional on purpose — it describes the access token in *this* response, so keeping a stale one next to a fresh token re-refreshes forever. Both are now tested. (Inert in practice today: `createGoogleAuthUrl` never sends `access_type=offline`, so Google returns no refresh token on this path — but the service shouldn't depend on that.)

**Non-determinism** — nothing enforces uniqueness on `(user_id, provider)`, and `processOAuthCallback` can insert a second row for a provider the user already has. The existing-link lookup now orders by `created_at` instead of taking an arbitrary row.

**Test gaps** — the one behavior this PR actually changes (Apple/Discord re-link) had no test; all the update-path coverage used Google, which already behaved that way. Added it for both providers. Also swapped `rejects.toThrow(TRPCError)` for `CONFLICT`/`BAD_REQUEST` assertions, since both error paths throw `TRPCError` and the weak assertion would pass if the two got swapped. And the "concurrent" test's comment overclaimed — dispatching two promises together doesn't force them to overlap inside the insert, and the result is identical serially. Reworded to describe it as the invariant it is, naming what actually makes it hold under concurrency.

**Docs** — the docblock now says the Apple/Discord re-link change applies to them too rather than justifying it via Google alone, and records that the absent `emailVerified` check is deliberate: identity here comes from the session, not the provider email (unlike `processOAuthCallback`, where SECURITY.md §4 applies). That's the thing most likely to get "fixed" incorrectly later. Trimmed the `src/server/CLAUDE.md` entry to just the callback.ts disambiguation, dropping the part that restated the file header.

Also folded `OAuthProvider` into the existing `OAuthProviderName` from `auth/oauth/config` instead of adding a fourth copy of the union.

Not addressed: the ordering change for Apple/Discord (the already-linked refusal now happens after the token exchange, consuming the one-time code). That's inherent to needing the `providerAccountId` before you can compare it, and is in the PR description above.

— Claude

PR #1519, opened 2026-08-31
Re-enable CIMD so claude.ai's "hosted client metadata" connector option works (#986)

brendanlong:
## Why

claude.ai's reworked connector UI (August 2026) makes Client ID Metadata Documents (CIMD, MCP auth spec 2025-11-25) an explicit, **recommended** option — "Use Anthropic's hosted client metadata". Per Anthropic's docs, it only engages when the server advertises **both** `client_id_metadata_document_supported: true` and `"none"` in `token_endpoint_auth_methods_supported`; otherwise claude.ai **silently falls back to DCR** — which still hits the old client-side abort (anthropics/claude-ai-mcp#546). That's why picking the CIMD option in the new UI "fails like it used to": with our explicit `false` (a deliberate 2026-07 workaround for the old auto-CIMD flow), claude.ai never actually tried CIMD against us.

## What

- **Flip `client_id_metadata_document_supported` to `true`** (`src/server/oauth/config.ts`), with the history and a revert breadcrumb in the comment.
- **New `src/server/oauth/cimd.ts`** — pure, unit-tested CIMD validation per the MCP 2025-11-25 spec / IETF draft:
  - client_id URL shape: https, non-root path, no fragment/userinfo (dot-segments are collapsed by WHATWG URL parsing; a normalized-but-unequal client_id is caught by the string-equality binding)
  - document validation: `client_id` ↔ URL string binding, well-formed non-empty `redirect_uris`, and rejection of secret-based `token_endpoint_auth_method` (CIMD clients are inherently public)
- **`fetchClientMetadata` hardening** (`service.ts`):
  - 5-minute success cache (matches the `max-age=300` Anthropic serves; errors never cached; capped at 100 entries since client_ids arrive on unauthenticated `/oauth/authorize` requests)
  - **vendored pinned fallbacks** for Anthropic's two hosted documents (`mcp-oauth-client-metadata`, `claude-code-client-metadata`) when the live fetch fails — Cloudflare intermittently serves datacenter egress IPs (Fly included) a managed challenge instead of the JSON (anthropics/claude-ai-mcp#650). Live fetch always takes precedence.
- **Consent screen** identifies CIMD clients by the client_id hostname next to the self-asserted name — "Authorize Claude (claude.ai)" — per the spec's guidance, since anyone can host a document naming itself anything.
- Docs: `src/server/oauth/CLAUDE.md` gets a CIMD section.

## Verification

- `pnpm typecheck`, full `pnpm test:unit` (with new `tests/unit/oauth-cimd.test.ts`), and the full integration suite against local Postgres/Redis all pass.
- **Live end-to-end simulation** of exactly what claude.ai's backend does, against a local server with claude.ai's real hosted document as `client_id`: `/oauth/authorize` (resolves the CIMD client, live fetch of `https://claude.ai/oauth/mcp-oauth-client-metadata`) → consent (renders "Authorize Claude (claude.ai)") → code → **PKCE-only token exchange with no client secret** → refresh-token grant → authenticated `POST /api/mcp` initialize succeeds. A mismatched/bogus CIMD URL is rejected with `invalid_client`.

## After merge

Retry the claude.ai connector with the **"Use Anthropic's hosted client metadata"** option (on either `https://lionreader.com/api/mcp` or `https://mcp.lionreader.com/mcp`). `LOG_MCP_REQUESTS=true` is already set in prod, so if it still fails we'll finally see whether claude.ai reached `/oauth/authorize` with its CIMD client_id. Refs #986.


brendanlong:
An adversarial security review (separate Opus reviewer over the branch) found one high-severity issue and several mediums in the first commit; all are fixed in the second commit (`20d6f1ac`):

- **High — redirect following broke the CIMD binding.** The metadata fetch followed cross-origin redirects (SSRF wrapper default), while the document↔client_id binding compares against the *originally requested* URL. An open redirect on any trusted HTTPS host would have let an attacker's document pass validation and display the trusted hostname on the consent screen. Fixed: the CIMD fetch never follows redirects (3xx = failure).
- **Medium — pinned fallback vs. revocation.** The pin previously covered *every* live-fetch failure, including a 404 — which is CIMD's revocation mechanism. Fixed: the pin now applies only to transport-shaped failures (network error, challenge page, 5xx, redirect); a withdrawn (404/410) or validation-failing document is authoritative and never papered over (pure `selectClientMetadata`, unit-tested).
- **Medium — consent ordering.** "Authorize Claude (claude.ai)" put the attacker-controlled string first; a `client_name` of `"Claude (claude.ai)"` + padding could push the real hostname out of view. Fixed: hostname leads ("Authorize claude.ai" / 'identifies itself as "Claude"'), and `client_name` is sanitized (control/bidi stripped) and capped at 100 chars.
- **Medium — fetch amplification + cache eviction.** Failures were never cached (each unauthenticated `/oauth/authorize` hit = one outbound fetch), and `Map.set` on an existing key doesn't refresh insertion order, so the hot claude.ai entry was permanently first in FIFO eviction line. Fixed: 60s negative cache, delete-before-set eviction, doc size limit 256 KiB → 16 KiB with per-field caps.
- Smaller: CIMD scopes now filtered like DCR (fail-closed), pins deep-frozen + defensive copies handed out, malformed `redirect_uris` entries dropped individually instead of failing the whole document, dead error path replaced with a warn log at the real failure site.

The full flow was re-verified live against claude.ai's real hosted document after the fixes (authorize → consent → PKCE token → authenticated MCP initialize). One pre-existing finding was split out as its own issue: the consent screen promises revocation in Settings but no grant-revocation UI exists (see the newly filed issue).

One behavior note for after deploy: if Claude Code switches from its DCR client_id to its CIMD one (`https://claude.ai/oauth/claude-code-client-metadata`), existing Claude Code users will see a one-time re-consent prompt (different client_id ⇒ no existing consent grant). Existing tokens keep working.

— Claude

PR #1522, opened 2026-09-01
Remove claude.ai-specific OAuth/MCP workarounds; document the working token-based setup (#986)

brendanlong:
## Why

The claude.ai web connector's OAuth flow is broken client-side (upstream anthropics/claude-ai-mcp#546): with CIMD advertised and explicitly selected, it still falls back to DCR and aborts before ever calling `/oauth/authorize` (server-verified 2026-09-01, see #986). The `mcp.lionreader.com` dedicated-host workaround didn't fix it either. What **does** work in the new connector UI is auth "None" + an explicit `Authorization: Bearer <MCP token>` header. So: remove every claude.ai-specific server special case, keep the spec-compliant OAuth surface (which works for Claude Code, Claude Desktop, and `mcp-remote`), and document/streamline the token-based claude.ai setup.

## Commit 1 — remove the special cases

- **Dedicated MCP-host surface** (`MCP_HOST` / `mcp.lionreader.com`): `resolveSurface()` dual-surface logic, the `/mcp` endpoint, its well-known document, `mcpConfig.host`, the fly.toml env, and the mcp-host audiences. OAuth config helpers are no-arg again.
- **Root-path endpoint aliases** (`/authorize`, `/token`, `/revoke`) and the proxy's `POST /register` → `/oauth/register` method-split (claude.ai origin-root-synthesis workarounds).
- **Trailing-slash normalization** for the OAuth surface in the custom server (claude.ai-specific, upstream #324).

**Kept:** path-inserted + root RFC 9728/8414 discovery, DCR, CIMD, PKCE, audience binding (incl. the legacy-origin grace), CORS, and `LOG_MCP_REQUESTS` — with the unserved root paths still in the log list as diagnostics for connectors that ignore advertised metadata. `src/server/oauth/CLAUDE.md` rewritten accordingly, with a "don't reintroduce without new evidence" note.

Tokens minted on the mcp-host surface stop validating (audience no longer accepted) — that surface never worked with claude.ai, so this affects only test grants.

## Commit 2 — working claude.ai instructions with inline token creation

The Integrations settings' claude.ai section now shows the working setup (auth **None** + `Authorization` header) with a one-click **Create MCP token** button that mints an MCP-scoped API token inline via the existing `apiTokens.create` and shows it once, next to the copyable `Bearer …` value.

| Before token | After token |
|---|---|
| fresh state | token created |

## Verification

- `pnpm typecheck`, full unit (3151) and integration (782) suites pass; knip/eslint clean.
- Live against a local server: discovery + 401 + DCR + CIMD authorize unchanged; `/authorize`, `/token`, `/revoke`, `/mcp`, and the mcp-host metadata now 404; `POST /register` serves the signup page like any page URL; trailing-slash `POST /api/mcp/` gets Next's default 308.
- The inline-created token authenticates: `POST /api/mcp` initialize succeeds with the minted Bearer value.

## After merge (optional infra cleanup)

`fly certs remove mcp.lionreader.com` and drop the `mcp` CNAME — the code no longer serves that host.


brendanlong:
An Opus reviewer went over the branch (verdict: approve with minor fixes — no security issues). All actionable findings are addressed in `5d3893b0`:

- **Dangling references**: `.env.example` still documented the removed `MCP_HOST` (silent no-op for a deployer) and described the old logging behavior; `src/server/http/CLAUDE.md` pointed at the deleted `trailing-slash.ts`; `src/server/mcp/README.md` and the demo article still called OAuth "what claude.ai does automatically". All fixed.
- **Logging noise**: `/register` stayed in the log list for DCR-probe diagnostics, but GET there is the human signup page — with `LOG_MCP_REQUESTS=true` in prod every signup-page view would have been logged. Now method-gated: only non-GET `/register` (a misplaced DCR attempt) is logged. The fly.toml comment now records keeping the flag on as a deliberate decision while the upstream bug is open, not a leftover TODO.
- **CSP tier flip pinned**: with the method-split gone, `POST`/`OPTIONS /register` moved from the strict nonce'd CSP to the relaxed static one. The reviewer assessed this safe (no route handler exists — a non-GET is a bodyless 405 where CSP is inert; the page renders no user HTML) and it's now pinned in `tests/unit/proxy.test.ts` so a re-added method-split fails loudly.
- Settings UI: success toast on token creation, matching the API Tokens page.

Noted but intentionally not changed: mcp-host-audienced tokens fail closed post-deploy (401 `oauth_token_audience_mismatch` with a correct `WWW-Authenticate`, prompting re-auth — the affected population is approximately nobody since claude.ai never completed a flow there), and trailing-slash URLs now get Next's default 308, which fetch-based clients follow with method+body preserved. The reviewer confirmed the audience handling and found no token leakage in the settings flow.

Deploy follow-up (also in the PR description): remove the `mcp` CNAME and `fly certs remove mcp.lionreader.com` — with the code gone, that hostname would otherwise keep serving the whole app under a second origin.

— Claude

Issue #1526, opened 2026-09-01
Confine Google Reader serial ids to the compat layer (batch resolver, not per-read fields)


brendanlong:
## Problem

The service layer returns `greaderItemId` / `subscriptionGreaderStreamId` / `feedGreaderStreamId` on **every** entry read (`EntryListItem`, `EntryFull`, ~6 select-list sites in `src/server/services/entries.ts`), even though only the Google Reader/Wallabag compat layer consumes them. Every other surface has to strip them (tRPC/REST via output schemas, MCP now via the shared `entry-output.ts` schemas), and the bigints are a standing serialization hazard for any future consumer (`JSON.stringify` throws on bigint — this already bit MCP once).

## Why filing despite the duplicated-query concern

A confinement does NOT require the compat layer to duplicate the entry queries. The compat formatters already know the entry UUIDs in the page they're rendering; they can resolve serials with **one batch query per response**:

```sql
SELECT ue.entry_id, e.greader_item_id, s.greader_stream_id AS sub_stream, f.greader_stream_id AS feed_stream
FROM user_entries ue
JOIN entries e ON e.id = ue.entry_id
LEFT JOIN subscriptions s ON s.id = ue.subscription_id
JOIN feeds f ON f.id = e.feed_id
WHERE ue.user_id = $user AND ue.entry_id = ANY($ids)
```

(≈ one indexed query for a ≤100-item page; `src/server/google-reader/id.ts` already owns the reverse direction.) Services then drop the three fields from `EntryListItem`/`EntryFull`, the strip logic in the shared output schemas becomes a no-op, and `visible_entries` no longer needs to carry the serials for the *forward* direction (the reverse resolvers still seek through it).

## Suggested scope

- Add a `resolveGreaderIdsForEntries(db, userId, entryIds)` batch helper in `src/server/google-reader/id.ts`; call it from the item formatters (greader `format.ts`, wallabag `format.ts`).
- Remove the three fields from the service entry types and select lists.
- Only worth doing if the helper + call sites stay small (~50 lines); if it starts needing per-endpoint variants, the current design is the cheaper one.

Not urgent — the fields are contained today — but it moves compat-specific data assembly into the compat layer, the same chokepoint principle as the sanitization/services-layer rule.

---
*Filed by Claude (complexity-audit follow-up).*

Issue #1525, opened 2026-09-01
Three coexisting optimistic-update strategies in the frontend


brendanlong:
## Problem

The frontend uses three different optimistic-update patterns:

1. **Timestamp-tracking concurrency tracker** (`src/lib/hooks/useEntryMutations.ts`, ~130 lines of bookkeeping): a per-entry `Map` of `{pendingCount, winningState, original*}`, where the response with the newest server `updatedAt` wins and is applied only if the cache isn't newer. Exists because React Query's default (`onMutate` + `cancelQueries` + rollback) is unusable here (cancelling aborts content fetches).
2. **Classic `onMutate`/rollback** for `subscriptions.delete` (optimistic removal + `onSuccess` absolute counts).
3. **No optimistic phase + shared handler** for `subscriptions.create` (`handleSubscriptionCreated`, shared with the SSE path, duplicate-safe).

Each pattern has different failure modes a contributor must know. Two observations on pattern 1 specifically (from a code survey, not observed in production):

- The tracking map is cleaned up only when `pendingCount` reaches 0; a mutation that neither resolves nor rejects leaks its entry.
- `recordMutationResult` treats a missing tracking entry as "single mutation, this is the winner", so a lost `startTracking` silently degrades to unguarded last-write-wins rather than failing loudly.

## Suggestion

Consolidate on one blessed pattern (realistically: the timestamp-tracking one, since it's the only one that composes with SSE + multi-client), or at minimum document in `src/FRONTEND_STATE.md` which pattern applies where and why, so new mutations don't pick a fourth. `applyWinningStateToCache` also does two full list-cache passes per entry (`updateEntriesReadStatus` + `updateEntryStarredStatus`) that could be one.

---
*Filed by Claude (complexity-audit follow-up).*

Issue #1524, opened 2026-09-01
Demo maintains a parallel router/state layer duplicating the SPA


brendanlong:
## Problem

The demo (`src/app/(public)/demo/`) reimplements the app's shallow-routing and state layer against static data: `DemoRouter.tsx` (~276 lines) + `DemoStateContext.tsx` (~324 lines), plus demo-specific wrappers (`DemoSidebar`, `DemoArticleView`, `DemoEntryListSSR`, `DemoLayoutContent`). It imports ~27 modules from core (`EntryList`, `EntryArticle`, `EntryStateButtons`, `LayoutShell`, `SummaryCard`, keyboard/gesture hooks, …) but drives them through this parallel layer instead of the real one.

This has already caused a real workflow hazard (recorded in the design-system work): a fix to a shared component's *usage* in the real app does not necessarily show on `/demo`, so screenshotting the demo to "verify" a real-app fix can verify the wrong code. It also means routing/state behavior changes (e.g. `?entry=` handling, keyboard nav) must be maintained twice.

## Suggestion

Drive the demo through the *real* state layer with a canned data source instead of a parallel one. The unit-test harness already proves the shape: `renderWithTrpc` (`tests/utils/component-test-helpers.tsx`) wraps components in the real `trpc.Provider` + `QueryClientProvider` with a terminating mock link that resolves each procedure from a handler map. A demo-mode tRPC link resolving `entries.list`, `entries.get`, `tags.list`, etc. from the static article fixtures would let the demo render the same `AppRouter`/component tree as the app, deleting `DemoRouter`/`DemoStateContext` and the drift class with them.

Worth checking before starting: whether the demo's mutation behaviors (mark read/star in-memory) map cleanly onto mock-link handlers + the normal cache-update paths (they should — they're plain cache writes), and whether the SSR'd demo entry list (`DemoEntryListSSR`) survives the switch.

---
*Filed by Claude (complexity-audit follow-up).*

Issue #1102, opened 2026-07-08
sync.events: mark_all_read keyset sentinel can skip an entry written in the same instant


brendanlong:
Latent low-severity edge found while reviewing the #1080 keyset fix (PR #1099). **Pre-existing** — the pre-keyset timestamp-only cursor had the same blind spot — but now made explicit by the `MAX_UUID` sentinel, so filing to track it.

### Mechanism
`mark_all_read` carries no single entry id, so the client advances its entries keyset cursor to `(T, MAX_UUID)`, where `T` is the mark-all-read `updated_at` (`markAllEntriesRead` in `src/server/services/entries.ts` — `const updatedAt = new Date()`, **millisecond** precision) and `MAX_UUID = "ffffffff-…"` (`src/lib/events/cursors.ts`).

A later catch-up `sync.events` then selects only rows with `GREATEST(entries.updated_at, user_entries.updated_at) > T`, because the keyset's tie branch `GREATEST = T AND entries.id > MAX_UUID` is **always false**. So any entry whose `GREATEST == T` **exactly** is permanently excluded from catch-up.

A different, unrelated entry can land on `GREATEST == T` if its `entries.updated_at` or `user_entries.updated_at` was written in the same millisecond as the mark-all-read (both come from JS `new Date()`, µs = `.000`, so a collision needs only the same *millisecond*, not the same microsecond — the code comment I added slightly understates this).

### When it actually bites
Only when **all** of these coincide:
1. A new entry's `GREATEST` equals `T` to full precision (same-millisecond write).
2. That entry's live `new_entry` SSE is missed — e.g. it's published to `feed:{feedId}:events` before the client's *dynamic* channel subscription for a just-subscribed feed completes, while the `mark_all_read` on `user:{userId}:events` is received.
3. The entry is genuinely new (not re-surfaced by a later change that would bump `GREATEST > T`).

### Mitigating factor
`mark_all_read`'s client handler already invalidates `entries.list` (`src/lib/cache/event-handlers.ts:144`), so in most orderings the missed entry reappears on the invalidation-triggered refetch. The residual window is: the entry's row becomes visible *after* that refetch completes but is stamped at `T`, so neither the refetch nor catch-up ever surfaces it until the next unrelated `entries.list` invalidation (navigation, another event, etc.).

### Options to consider
[...]

PR #1527, opened 2026-09-01
Deduplicate cross-surface declarations: counter rule, MCP/tRPC schemas, SSE event union, deploy-window fallbacks

brendanlong:
Four deduplication fixes from the complexity audit, one commit each:

## 1. Counter contribution rule: 5 copies → 1 SQL function (migration 0108)
The rule for which `user_entries` rows count toward which denormalized counter was written in each of the three trigger functions plus twice in `reconcileCounters`, kept in sync by a comment. It now lives in `user_entry_counts_toward_unread` / `user_entry_counter_contrib` (inlinable `LANGUAGE sql IMMUTABLE`), called by the triggers and both reconcile queries — reconcile mirrors the triggers **by construction**. Expand-only migration; counter semantics unchanged; the opt-in perf benchmark still passes.

## 2. MCP ↔ tRPC: shared entry output + tag name schemas
`stripGreaderIds` was a hand-rolled re-statement of the entry response shape, and the MCP tag-name constraints had already drifted (tRPC trims, MCP didn't). Entry output schemas now live in `src/server/trpc/entry-output.ts`, used as tRPC `.output()` schemas **and** parsed through by the MCP `list_entries`/`get_entry` tools; `tagNameSchema` moves to shared validation. One contract per shape, `stripGreaderIds` deleted.

## 3. SSE event union: 4 declarations → 1 base per event
The event union existed as Redis wire schemas (pubsub.ts), the client parse union, the polling union, and hand-written unvalidated JSON literals in the SSE route. Each event now has one base schema in `src/lib/events/schemas.ts`; the polling union uses bases directly, the client union derives (timestamp defaults / updatedAt fallbacks), the Redis wire derives by extending with routing fields, and the route builds typed `ServerSyncEvent` values the compiler checks. Wire shapes unchanged (the entry_updated literal's `feedType` was never populated by any publisher and never parsed by the client). The announcement-level enum also collapses to one definition.

## 4. Wire protocol version instead of per-field deploy-window fallbacks
Auditing every `.optional()` blamed on deploy windows showed most are actually **semantic** (spam entries carry no list payload; counts omitted on count-query failure; catch-up `subscription_created` can't compute counts) — comments corrected. The one pure deploy-window artifact was `mark_all_read.entryId`. `publishToChannel` now stamps `SYNC_PROTOCOL_VERSION` on every event; the client answers a version-mismatched live event with one coalesced catch-up sync instead of parsing it. `entryId` becomes required (deleting the `MAX_UUID` cursor fallback, #1102), and future event-shape changes bump the version rather than growing independently-tested `.optional()` branches. Policy documented at `SYNC_PROTOCOL_VERSION` and in `FRONTEND_STATE.md`.

## Testing
- `pnpm typecheck`, `pnpm test:unit` (3167), `pnpm test:integration` (783) green
- Full `pnpm test:e2e` (65 passed) against local services — realtime specs exercise the versioned pipeline end-to-end in a real browser
- Counter perf benchmark (`RUN_PERF_TESTS=1 … counters-perf.test.ts`) passes with the rewritten triggers

Related issues filed from the same audit: #1524 (demo parallel router), #1525 (optimistic-update strategies), #1526 (greader id confinement).


brendanlong:
An Opus-level reviewer subagent went over the full diff (it independently re-ran typecheck/lint/knip/unit, loaded migration 0108 into real Postgres 17 and 18, and compared the old FILTER-based counter aggregates against the new contrib-based ones over 5,000 randomized rows with injected NULLs — 0 mismatches, and EXPLAIN confirms the SQL functions fully inline, so plans are byte-identical to before).

**Verified safe:** migration semantics + expand/contract compatibility, `entry-output.ts` exactly matching the service shapes, Redis wire shapes unchanged except the three deliberate deltas, dropping `entry_updated.feedType` (never populated, never read), required `mark_all_read.entryId` (every publisher already always sends it), the full rolling-deploy matrix for the version gate, and cross-user channel isolation.

**Findings, addressed in the two follow-up commits:**
- Unusable-event syncs are now rate-limited (10s floor) — the first deploy of the version gate makes every pre-release event version-less, so an old worker's publish burst would otherwise chain back-to-back `sync.events` queries.
- A current-version Zod parse failure now also triggers a (rate-limited) catch-up sync instead of a silent drop — covers an old app machine forwarding a newer worker's payload reshaped to its older schema while propagating the worker's `v`.
- The e2e `refetchProcedures` filter now counts `sync.*`, so a regression in `v` stamping would fail the realtime tests instead of silently passing via the fallback (full suite re-run: 65 passed).
- `schema.sql` is replayable again (`dump-schema.ts` keeps `SET check_function_bodies = false`; verified with `psql -v ON_ERROR_STOP=1 -f`) and the PG-18 dump artifacts are reverted so the file only carries the intended function changes.
- Smaller items: misplaced `tagColorSchema` JSDoc fixed, the SSE-only-event deploy-window tradeoff documented at `SYNC_PROTOCOL_VERSION`, and the 8 hand-written copies of the feed-type enum folded into one exported `feedTypeSchema`.

— Claude

PR #1533, opened 2026-09-08
feat: Let users see and revoke their OAuth grants in Settings (#1520)

### LINKED ISSUE #1520: OAuth consent screen promises revocation in Settings, but no grant-revocation UI exists
Found during the security review of #1519 (pre-existing, not introduced there).

The OAuth consent screen tells users: *"Only authorize applications you trust. You can revoke access at any time in Settings."* (`src/app/(spa)/oauth/consent/ConsentForm.tsx`). But there is no UI (or tRPC endpoint) anywhere in `src/` for listing or revoking `oauth_consent_grants` / issued OAuth tokens — the only revocation surface is the RFC 7009 `/oauth/revoke` endpoint, which requires the *client's* credentials, not the user's.

Consequences:

1. The consent-screen claim is false today.
2. `hasConsent` keys on `(userId, clientId)` + scopes, so once granted, re-authorization is silent — the user never sees the consent screen again for that client.
3. CIMD (#1519) makes this more significant: a CIMD client's `redirect_uris` live in a document its *operator* can change at any time, and with a standing consent grant, `/oauth/authorize` will issue codes to whatever the document currently lists with no user interaction. Users should have an off switch.

Suggested scope: a Settings section listing active OAuth grants (client name/host, scopes, granted date, last used) with a revoke button that deletes the consent grant and revokes outstanding tokens for that `(user, client)` pair. Until then, consider softening the consent-screen copy.

— Claude

brendanlong:
Closes #1520.

The OAuth consent screen told users *"You can revoke access at any time in Settings"*, but no such UI (or endpoint) existed — the only revocation surface was the RFC 7009 `/oauth/revoke` endpoint, which authenticates the **client**, not the user. Since `hasConsent` keys on `(userId, clientId)`, a granted client is re-authorized silently forever, and a CIMD client's `redirect_uris` live in a document its operator can change at any time. Users needed an off switch.

## What's here

- **`oauthGrants` router** (session-only — an MCP or `reader:full-access` token must never manage grants): `list` returns active consent grants with their scope descriptions, grant date and last access-token use; `revoke` takes a `clientId`.
- **Revocation does both halves** — drops the consent grant *and* revokes the user's outstanding access/refresh tokens for that client. Either alone leaves a way back in: live tokens keep refreshing themselves, and a standing consent grant makes `/oauth/authorize` mint codes with no user interaction.
- **Settings → Connected Apps** page. Registered clients show our own name; CIMD clients show their `client_id`'s hostname (their name is self-asserted, and re-fetching a metadata document per row to render a list isn't worth it) — matching what the consent screen leads with.
- Consent-screen copy now names the exact place: *"Settings → Connected Apps"*.

## Screenshots

| Light | Dark |
| --- | --- |
|  |  |

## Tests

- `tests/integration/oauth-consent-grants.test.ts` (9 tests): naming for registered vs CIMD clients, last-used reporting, cross-user isolation on both list and revoke, revocation killing access + refresh tokens while leaving other clients alone, `NOT_FOUND` for an unauthorized client.
- `tests/e2e/connected-apps.spec.ts`: the full user path — see the app, click Revoke, confirm the grant is gone from the UI and the access token is revoked in the DB.

Integration (791), unit (3189), full e2e (75), typecheck, lint, knip and `check:colors` all pass locally.


brendanlong:
Ran an independent review pass over this branch and fixed what it found. Two of the findings were real holes in the revocation itself:

- **An unredeemed authorization code survived the revoke.** A code minted moments before the user clicked Revoke stayed redeemable for the rest of its 10-minute life, buying a fresh hour of access and a 30-day refresh chain — silently undoing the revocation. Revocation now burns the user's unredeemed codes for that client, and the code exchange re-checks `hasConsent`.
- **A refresh rotation could race the token sweep.** Under READ COMMITTED, a rotation in flight commits its successor tokens after the sweep's statement snapshot, so a live refresh chain could survive. `rotateRefreshToken` now takes a `requireUserConsent` opt-in checked inside the claim transaction, so such a successor is born unusable. It's opt-in rather than unconditional because the Wallabag password grant shares that code path and has no consent grant to check.

The grant is revoked *before* the token sweep, which is what makes both re-checks fail closed for anything racing it.

Smaller fixes: dropped a dead branch in the app card (`clientHost` and `clientName` are never both set), shared the CIMD host-vs-name display rule between the consent screen and the settings list instead of spelling it twice, moved the OAuth client fixture into the shared integration factories, and asserted the session-only invariant on `oauthGrants` in the token-scope suite.

New coverage: the revoke-kills-pending-code path, the rotation consent gate (and that Wallabag-style rotation is unaffected), and an e2e test that reaches the token endpoint's consent check the only way production can — a code that outlived the sweep.

Verified the reviewer's caching question separately: there is no Redis or in-process cache of OAuth access-token validation, so a revoked token stops working on the very next request everywhere.

Integration (797), unit (3189), e2e (76), typecheck, lint, knip and check:colors all pass.

— Claude

Issue #952, opened 2026-07-02
Data integrity: non-atomic entry→user_entries fanout and subscription writes can lose data / hide entries


brendanlong:
From a multi-agent design review (2026-07-02, verified at commit 0976051, post-#918). Grouped: multi-step writes that aren't atomic, plus check-then-act races that violate the repo's own upsert convention.

Note: #918 already fixed the ordering bug where `publishNewEntry` fired before the `user_entries` fanout — the fanout durability gap below is a *different* issue and remains.

## Non-atomic / non-retryable writes

- [ ] **Entry insert → `user_entries` fanout is non-atomic and non-retryable on all three write paths** (high). Feed path: fanout is driven by in-memory `isNew` after the loop (`entry-processor.ts` `createUserEntriesForFeed`); a worker crash between entry insert and fanout means the next fetch sees the entry already exists with matching `content_hash`, marks it `isNew:false`, and never fans it out — subscribers permanently lose it. Email path is worse: the Mailgun retry hits the duplicate-Message-ID early return (`process-inbound.ts:389-402`), so the newsletter is permanently invisible. Saved path (`services/saved.ts:208-241`) has the same two-statement pattern. Fix: wrap email/saved in `db.transaction`; on the feed path, make fanout state-driven (pass all fetched entry IDs to the already-idempotent `createUserEntriesForFeed` so it self-heals).
- [ ] **`createSubscription` is multi-step non-transactional** (`services/subscriptions.ts` ~440-514) — subscription upsert → `subscription_feeds` insert → `user_entries` populate; a crash mid-way leaves an active subscription with no `subscription_feeds` row.
- [ ] **Subscribe/fetch race** (`services/subscriptions.ts:379` read, `:506-514` insert) — `createSubscription` reads `feedRecord` early, then inserts `user_entries` matching `e.last_seen_at = feedRecord.lastEntriesUpdatedAt`. A fetch completing in between bumps `last_seen_at`, so zero rows match and the new subscriber sees an empty feed until the next new entry. Re-read `last_entries_updated_at` inside the INSERT's subquery.
[...]

PR #1534, opened 2026-09-08
Don't dump an archive re-announcement into unread (#1500)

### LINKED ISSUE #1500: Fetch pulled a number of old articles
A recent fetch of the feed https://thezvi.wordpress.com/feed/ backfilled articles over the last 4 years including 628 articles. I’m not sure if something happened on the feed but LionReader reports the feed is healthy. I’ve also noticed this feed tends to double post articles and picks up updates as though they were new articles.

Image

brendanlong:
Fixes #1500.

## What happened

The Zvi's WordPress blog **bulk-modified 606 posts on 2026-08-04** — every post in the archive has `modified: 2026-08-04` in the WordPress.com API, with its original 2022–2026 `date` intact. Six days later the issue was filed reporting ~628 backfilled articles spanning four years.

The feed itself only ever lists 10 items, and Lion Reader is younger than those posts, so they can't have accumulated from polling. They arrived as *announcements*: the blog advertises a WebSub hub (`?pushpress=hub`), and re-saving a published post fires the hub. Each announcement carried a post we'd never seen, so `processEntries` created it — with its original publication date — and fanned it out unread to every subscriber.

Nothing failed, which is why the feed still reported healthy.

## The fix

`isBackfilledEntry` classifies a **first sighting** as a backfill when the article was published well before `feeds.last_fetched_at` as of the previous fetch: if we pulled the whole feed at time T and the article was already old then but wasn't in it, it's history, not a release. Because the replay can arrive one push at a time, the signal has to be per-entry — a "this fetch delivered 600 new entries" heuristic never fires.

The verdict is **persisted as `entries.is_backfill`** (migration 0108, additive `NOT NULL DEFAULT false`), and every path that grants visibility copies it into `user_entries.read`: the fetch fanout, its #952 self-heal, and the subscribe-time populate. Deriving the read state from one stored fact rather than from per-call context is what makes it survive a crash between writes, a duplicate GUID inside one document, and a subscriber who joins during a WebSub feed's 24h backup-poll window — when the replay is still inside the current generation because a push deliberately doesn't advance `last_entries_updated_at`.

Backfilled entries also publish no `new_entry` event, aren't charged against the hub's push-reliability tally, and don't spend the per-fetch full-content budget the genuinely new articles need. Nothing is hidden from queries: they stay in the feed's list and in search, they just don't reach an unread badge.

The threshold is a month — wide enough to clear the noise floor of ordinary syndication (stale CDN copies, clock skew, mild backdating), far short of the multi-year replays it exists to contain. The guard is inert where old-but-new is legitimate:

- a feed's **first fetch** (`last_fetched_at` is null — we weren't watching, so its whole current window is new to us however old it is);
- a **forced subscribe-time refresh** (suppressed, so the stale-feed and fresh-feed subscribe paths deliver the same thing);
- a **dormant feed polled again** (everything published during the gap is *after* the previous fetch);
- entries with **no publication date**.

The accepted risk — a feed whose dates are systematically wrong gets everything silently marked read — is named in `src/server/feed/CLAUDE.md` rather than left implicit.

## Not fixed here

The reporter also notes the feed "tends to double post articles and picks up updates as though they were new articles" — a GUID-stability problem, filed as #1535. A read-state guard can't help, and it wouldn't have prevented this flood anyway (we'd never seen those 2022 posts under *any* GUID).

## Testing

`pnpm typecheck`, `pnpm lint`, `pnpm test:unit` (3210 passed), `pnpm test:integration` (813 passed) all pass.

New coverage:
- `tests/integration/entry-processor.test.ts` — classifier boundaries incl. the exact 30-day edge; an archive re-announcement landing read while a same-fetch fresh article stays unread (asserting the subscription's trigger-maintained `unread_count` and that `read_changed_at` stays NULL, so a later user mark-unread still wins); old entries staying unread on a first fetch; a repeated GUID in one document not escaping as unread; the #952 self-heal healing a backfill to *read*; re-listing not flipping it back; no `new_entry` event.
- `tests/integration/subscriptions.test.ts` — a new subscriber granted a backfilled entry already read.
- `tests/integration/websub-notification.test.ts` (**new file**) — `ingestWebsubNotification` had no integration coverage at all despite being the #1500 vector. Covers a fresh push delivered unread + hub credited, an archive replay delivered read + hub *not* credited, the `last_fetched_at`/`last_entries_updated_at`/`body_hash` invariants, and an unparseable body.

- `tests/integration/fetch-feed-handler.test.ts` (**new file**) — `handleFetchFeed` had no tests anywhere, so the poll-path wiring was verified by inspection only. Drives the handler against a loopback origin (the `feed-discovery-redirect` pattern): both early returns; a first fetch storing entries, fanning them out unread and recording the validators the next poll conditions on; the guard inert on a first fetch and on a forced refresh, firing on an archive re-announcement; conditional headers sent and omitted; 304 and byte-identical bodies short-circuiting without churning the visibility generation; 404/5xx/unparseable bodies backing off without parking the feed.

Mutation-checked: zeroing `previousLastFetchedAt`, dropping the `forceReprocess` suppression, and sending validators on a forced refresh each fail exactly one of these and nothing else.


brendanlong:
Filed the GUID-stability half of the report as #1535 — that one needs identity matching rather than a read-state guard, and wouldn't have prevented this flood anyway (we'd never seen those 2022 posts under any GUID).

— Claude

PR #1536, opened 2026-09-08
fix: Validate sy:updatePeriod at the napi boundary instead of casting

brendanlong:
Follow-up to the observation raised while reviewing #1514.

## The problem

`toFeedParseResult` asserted `raw.updatePeriod as UpdatePeriod` on a value the native module types as a bare `string`:

```ts
// The native side already lowercased and validated the period value.
if (raw.updatePeriod) syndication.updatePeriod = raw.updatePeriod as UpdatePeriod;
```

The comment was accurate — `native/feed-parser/core/src/types.rs:74` holds the valid list and `rss.rs:237` / `atom.rs:183` lowercase and check against it — but the invariant was enforced on one side of the napi boundary and merely *asserted* on the other. A change to the Rust parser could put an arbitrary string into a field the rest of the app treats as a closed union, and nothing in TypeScript would notice.

Consequences are currently mild (`syndicationToSeconds` has a defensive `periodSeconds === undefined` fallback, so a bad value degrades to "no hint" rather than `NaN`), which is exactly why this could rot unnoticed.

## The change

- `isUpdatePeriod` type guard in `src/server/feed/types.ts` narrows the string; an unrecognized period is dropped, matching what the Rust parser does with one.
- The union was spelled out twice — inline in `SyndicationHints` and in `streaming/syndication.ts`. `feed/types.ts` is now the single source, sitting next to the interface that uses it, and the one-type `syndication.ts` module is deleted.
- `scheduling.ts` names `UpdatePeriod` directly instead of indexing back into the interface with `NonNullable<SyndicationHints["updatePeriod"]>`.

**No behavior change today** — the native parser cannot currently produce a value this rejects. This closes the boundary so it stays that way.

## Tests

`toFeedParseResult` had no test coverage at all, so `tests/unit/native-result-syndication.test.ts` is new: 5 cases covering a recognized period, an unrecognized one, an unrecognized period alongside a valid frequency, case-sensitivity, and no hints. Verified 3 of the 5 fail against the old cast.

Full suite green on Node 26: `typecheck`, `lint`, `format:check`, `knip`, and `test:unit` (161 files / 3215 tests).

## Relationship to #1514

Independent — branched from master, and these changes are clean under both knip 5 (current master) and knip 6 (#1514), so the two can merge in either order.

— Claude


brendanlong:
Review follow-ups pushed.

The one that matters: `UPDATE_PERIODS` duplicates `VALID_UPDATE_PERIODS` in `native/feed-parser/core/src/types.rs` with nothing keeping them in sync, and the failure mode is silent — if Rust ever accepts a period the guard lacks, the hint gets dropped rather than honored. So all five periods now go through the real `parseRss`, which catches drift in either direction. Confirmed it works by removing `"yearly"` from the TS list: `keeps yearly` fails. The coupling is also called out in the guard's docblock now.

Also added a case for a frequency arriving with no period at all (the RSS syndication module allows it — that's a different branch from "period unrecognized"), and switched `if (updatePeriod)` to an explicit `!== undefined` to match the line computing it.

Test file is 12 cases now; full suite 3222 tests. `typecheck`, `lint`, `knip`, `format:check` all clean.

Review also verified the "no behavior change" claim by confirming `toFeedParseResult` is the only producer of `SyndicationHints` anywhere — JSON Feed never sets it, OPML has no such concept, plugins don't touch it, and it's never persisted — and that the guard's five values match the Rust constant exactly. Docs were checked for references to the deleted `streaming/syndication.ts`; there are none, and the spec link from its docblock survives on `SyndicationHints`.

Two notes I did not act on, both deliberate:
- `scheduling.ts:100`'s `periodSeconds === undefined` check is dead under the type system, but it was equally dead before this PR via `NonNullable<SyndicationHints["updatePeriod"]>`. Pre-existing and harmless.
- Using a Zod enum for the guard would be inconsistent here — `z.enum` in this codebase is used at API/input boundaries, not for narrowing napi return values on the per-fetch parse path. The `(X as readonly string[]).includes(...)` idiom matches `src/lib/ai/model-ref.ts`.

— Claude

brendanlong:
Removed the dead `periodSeconds === undefined` fallback in `syndicationToSeconds`.

Worth being precise about why it is only *now* safe to delete: it was unreachable before this PR too, but only on paper. The sole producer cast an unvalidated native string into the union, so a Rust-side change could have put an unknown period in and made the branch live again. The guard in the first commit is what turns it from load-bearing-in-principle into genuinely redundant — `PERIOD_SECONDS` is a total `Record<UpdatePeriod, number>`, `updatePeriod` is narrowed by the preceding check, and `noUncheckedIndexedAccess` is off, so the lookup is typed `number`.

I also swept the rest of the repo rather than assume that was the only instance. Running knip with the dead-code categories the project's script leaves out (`--include files,exports,types,nsExports,nsTypes,enumMembers,duplicates`) turns up **no** unused files, unused value exports, duplicate exports, or unused enum members — the only hit is 83 unused exported *types*, mostly Drizzle `$inferSelect`/`$inferInsert` aliases in `src/server/db/schema.ts`. That category is deliberately excluded from `pnpm knip`, and clearing it is an unrelated repo-wide change, so I left it alone. Happy to take it on separately if you want it.

3222 tests pass; `typecheck`, `lint`, `knip`, `format:check` clean.

— Claude

PR #1577, opened 2026-09-08
ci: Gate merges on cargo test and cargo clippy

### LINKED ISSUE #1568: CI never runs cargo test or cargo clippy: 132 Rust tests are dead weight
No workflow in `.github/workflows/` invokes `cargo test` or `cargo clippy`. Verified: `grep -n "cargo" .github/workflows/*.yml` returns only two comment lines in `ci.yml` (about build caching). CI **builds** the native modules and stops there.

Meanwhile `grep -rc "#\[test\]" native/*/core/src/*.rs native/*/src/*.rs` counts **132** test functions across the four crates. None of them gate a merge.

This matters most for `native/sanitizer`, which `SECURITY.md` names as the primary XSS defense. Its Rust suite includes the allow-list and escaping tests — the ones that would catch a sanitizer regression — and today a PR can turn any of them red and still go green.

Two pieces of corroborating evidence that the gap is already costing something:
- PR #1567 (the SVG/MathJax depth cap) found **two pre-existing clippy lints** on `master` that nobody had seen: `clippy::needless_return` in `mathjax.rs::convert_element` and `clippy::bool_assert_comparison` in `scanner.rs:373`. Neither is harmful; both are evidence clippy has not run in a long time.
- The stack-overflow crash fixed in that PR (see #1540) was reachable from any feed and would not have been caught by a test suite that never executes.

Suggested: add `cargo test --workspace` and `cargo clippy --workspace --all-targets -- -D warnings` to the existing native-build job, which already has the toolchain and a warm cargo cache, so the marginal cost is small. The two pre-existing lints need fixing or an explicit `allow` first, or the clippy step lands red.

— Claude

brendanlong:
Fixes #1568.

CI built the native modules and stopped there, so the **144 Rust tests** across the four crates never gated a merge — including `native/sanitizer`'s allow-list and escaping tests, which `SECURITY.md` names as our primary XSS defense. Clippy had likewise not run in long enough for lints to accumulate on `master`.

## Changes

**New `rust-checks` CI job** running `cargo test --workspace` and `cargo clippy --workspace --all-targets -- -D warnings`.

It's a separate job rather than steps inside `build-native` (as the issue suggested) because `build-native` skips every step on a cache hit — a check that can be skipped isn't a gate. It gets its own cargo cache via `Swatinem/rust-cache`, and since `deploy.yml` gates on the CI workflow succeeding, this gates deploys too.

**`scripts/cargo-native.sh`** — each native module is its own standalone cargo workspace (see the `[workspace]` stanzas in `native/*/Cargo.toml`), so there's no repo-level `--workspace` that reaches all four. The script loops over them and keeps going after a failure so one run reports every crate's problems. Exposed as `pnpm test:native` / `pnpm lint:native` for local use.

**The three pre-existing lints**, which would otherwise land the clippy step red:
- `clippy::needless_return` — `mathjax.rs::convert_element` (the trailing `return` in the `mjx-*` branch; nothing follows it)
- `clippy::bool_assert_comparison` — `scanner.rs:373`, `assert_eq!(x, true)` → `assert!(x)`
- `clippy::unnecessary_map_or` ×2 — `atom.rs`, `rel.map_or(true, |r| r.is_empty())` → `rel.is_none_or(..)`

## Verification

- `pnpm test:native`: 144 passed, 0 failed (107 sanitizer, 26 markdown, 11 feed-parser)
- `pnpm lint:native`: clean across all four crates
- `pnpm test:unit`: 3191 passed — the sanitizer/feed-parser `.node` binaries rebuilt from the lint fixes behave identically
- `pnpm format:check`, `pnpm knip`: clean


brendanlong:
Pushed b1451edb addressing an independent review.

**Applied:**
- `cache-on-failure: true` on `rust-cache` — its `post-if` is `success() || CACHE_ON_FAILURE`, so without this a job whose whole purpose is to go red never warms its own branch cache while you're iterating on a failure.
- `cargo-native.sh` now prints `FAILED in: <crates>` at the end (it deliberately continues past failures, so an early one is easy to miss at the tail of a long log) and errors out explicitly when cargo is missing instead of running it four times.
- `mathjax.rs`: made the "stray non-MathJax element" case its own guard clause. Just deleting the `return` left the unknown-`mjx-*` branch as the only one in a long guard-clause chain relying on being physically last, and orphaned the trailing comment. Behavior identical; rebuilt `sanitizer.node` and re-ran the 122 TS sanitizer/MathJax/SVG tests.
- Doc trims per the repo's "say each thing once" rule: the workspace rationale now lives only in the script header (it was duplicated into CLAUDE.md and `ci.yml`); the `SECURITY.md` citation now attaches to the "primary XSS defense" claim it actually supports.

**Deferred to follow-ups** — both are pre-existing gaps the review surfaced, not regressions from this PR:
- #1579 — Rust MSRV is unpinned, so CI compiles with a *newer* rustc than the Docker build's Alpine `apk add rust`. Fixing it means deciding what to pin and where, which is its own change.
- #1580 — `cargo fmt` is gated nowhere (prettier skips `.rs`; `lint-staged` has no `*.rs` glob). Cheap to add, but needs someone with `rustfmt` installed to confirm the tree is clean first.

**Verified:** `pnpm test:native` 144 passed / 0 failed; `pnpm lint:native` clean; script exits 1 on a failing crate, 0 clean, 64 no-args, 69 no-cargo. All checks were green on the previous commit, including the new `Rust Tests & Clippy` job — its log confirms 11 + 26 + 107 tests actually executed and clippy ran on all four crates (~1.5 min cold).

— Claude

PR #1581, opened 2026-09-08
fix: Give the OPML import drop zone a visible focus indicator

### LINKED ISSUE #1573: OPML import file input has no visible focus indicator (opacity-0 erases the outline)
The OPML import drop zone at `src/components/settings/OpmlImportExport.tsx:292` uses `className="absolute inset-0 cursor-pointer opacity-0"` on its `<input type="file">`.

The input is focusable and keyboard-reachable — that part is correct, and it is the pattern PR #1572 copied to fix the *unreachable* upload drop zone in `FileUploadButton.tsx`. But `opacity-0` applies to the focus outline too, so the global `:focus-visible` rule is erased: a keyboard user can tab to the control and has no way to see that they have.

Found while implementing #1572. That PR verified the problem in Chromium and used `text-transparent file:hidden` instead, which hides the input's own chrome while letting the outline draw around the drop zone. It deliberately did not touch this file, so the latent issue remains here.

Fix is to apply the same treatment. Worth doing both together so the two drop zones stay consistent — right now they will differ until this lands.

— Claude

brendanlong:
Fixes #1573.

The `<input type="file">` covering the OPML import drop zone was `opacity-0`. It was focusable and keyboard-reachable, but `opacity-0` applies to the focus outline too, so the global `:focus-visible` rule was erased — a keyboard user could tab to the control with no way to see that they had.

It now uses the same treatment PR #1572 verified in Chromium for the saved-article upload zone: transparent text plus `file:hidden`, which hides the input's own chrome while letting the outline draw around the zone.

Since the point is that the two drop zones stay consistent, the class string moved into one shared constant (`src/components/ui/drop-zone-file-input.ts`) that both call sites import, with the rationale documented there once. The OPML input also gained the `aria-label` it was missing — its visible text lives in a `pointer-events-none` sibling, not a `<label>`, so it had no accessible name.

## Testing

- New `tests/unit/frontend/components/OpmlImportExport.test.tsx` mirrors the FileUploadButton accessibility tests: accessible name, no `opacity-0`/`hidden`, non-negative tabindex, focusable, and the input alone drives the import preview.
- `pnpm test:unit` (3200 passing), `pnpm typecheck`, `pnpm lint`, `pnpm knip` all clean.
- Verified Tailwind still emits `text-transparent` and the `file:` variant now that the class string lives in a `.ts` file.


brendanlong:
Ran an independent reviewer over the diff. Nothing blocking; it verified the CSS approach (including that `h-full w-full` is a real fix, not decoration — a file input is a replaced element, so `inset-0` alone left master's input at its intrinsic size with only part of the OPML zone clickable), confirmed no hit-testing regression from the input now covering the zone, and re-checked that Tailwind still emits `file:hidden` now that the string lives in a `.ts` file.

Three of its notes are now addressed in 08cc353:

- **Square outline on a rounded zone.** The input had no radius, so the focus outline cut across the `rounded-lg` corners. Added `rounded-[inherit]`.
- **Duplicated class assertions.** Both zones render one string, so what it must and must not contain now lives in a single test over the constant — including the "no `focus:` utilities" assertion `src/components/CLAUDE.md` asks of shared controls, which neither component test was making. The component tests keep what's actually theirs.
- **Docs.** `src/components/CLAUDE.md` named `focus:outline-none` as the trap but not `opacity-0`, which has now caused this bug twice.

The rest were pre-existing gaps in the OPML flow rather than anything this PR introduces, so I filed #1584 instead of widening the diff: selecting a file unmounts the focused input and drops focus to `<body>` with no live region announcing the change (now the next barrier for a keyboard user, since this PR is what gives them a way in), the size/extension hint isn't wired up with `aria-describedby`, and `fileInputRef` plus its reset in `handleReset` are dead code.

— Claude

PR #1583, opened 2026-09-08
build: Pin rustc to the version production compiles with (#1579)

### LINKED ISSUE #1579: Rust MSRV is unpinned: CI validates against a newer rustc than the Docker build uses
Split out of the review of #1577 (which added the `rust-checks` CI job).

No `rust-toolchain.toml` exists and no manifest sets `rust-version`, so the Rust compiler version floats independently in two places:

- **CI** (`.github/workflows/ci.yml`, `rust-checks` + `build-native`) uses `ubuntu-latest`'s bundled stable — currently ~1.97/1.98.
- **Production** (`Dockerfile:51-55`) is `node:26-alpine` + `apk add --no-cache rust cargo`, i.e. whatever Alpine ships for whichever Alpine `node:26-alpine` currently tracks (roughly 1.87 on 3.22, 1.91 on 3.23).

Two consequences:

1. **CI structurally cannot catch a too-new stdlib API.** It compiles with the *newer* toolchain, so a call stabilized after Alpine's rustc passes every check and then fails the Docker build (or, worse, fails only after the base image is rebuilt). #1577 introduced the first post-1.80 API in the tree (`is_none_or`, stabilized 1.82) — it happens to be safe, but nothing verified that.
2. **A clippy release can redden unrelated PRs**, since `rust-checks` runs `-D warnings` against a floating stable.

Setting `rust-version` in the manifests would turn on `clippy::incompatible_msrv` and make (1) a real gate. But the value to pin is the question: hardcoding today's Alpine rustc goes stale silently every time the base image moves. The alternatives are pinning the Docker toolchain explicitly (rustup in the `native-builder` stage instead of `apk add rust`, plus a `rust-toolchain.toml` shared with CI), or pinning the Alpine version in the `FROM`.

Not urgent — it's a latent build break, not a runtime one, and it surfaces loudly in the Docker build rather than silently in production.

— Claude

brendanlong:
Fixes #1579.

## Problem

The Rust compiler version floated in two places independently:

- **CI** used `ubuntu-latest`'s bundled stable (~1.97/1.98).
- **Production** used whatever `apk add rust` gives in `node:26-alpine` (Alpine 3.24 → rustc **1.96.1** today).

CI ran the *newer* compiler, so it structurally could not catch a call to an API stabilized after Alpine's rustc — that only failed later, in the Docker build. And a clippy release could redden unrelated PRs.

## Fix

- **`rust-toolchain.toml`** pins CI and local dev to `1.96` — Alpine's rustc. What CI compiles is now what production compiles, and clippy no longer floats. CI needs no setup step: the runner's rustup installs the pinned version on the first cargo call.
- **`Dockerfile`**: the `native-builder` stage pins `node:26-alpine3.24` and asserts apk's rustc still matches `rust-toolchain.toml`, so bumping that tag past the pinned compiler fails the build with a message naming the fix instead of drifting silently. Only that stage pins its Alpine version — the other stages keep floating so the runtime image still picks up base-image updates.

### Why not `rust-version` in the manifests?

The issue suggested it to enable `clippy::incompatible_msrv`. Pinning the toolchain is a strictly stronger gate: that lint only knows about stdlib APIs, while the pinned compiler also rejects newer language, edition and cargo features — and it keeps one version number instead of eight (four crates plus their cores).

## Testing

- `./scripts/cargo-native.sh test --workspace` and `clippy --workspace --all-targets -- -D warnings` pass under 1.96.1.
- `podman build --target native-builder .` succeeds: the guard passes and all four `.node` modules build.
- Exercised the guard's comparison directly: `1.96` vs `1.96.1`/`1.96` pass; `1.97.0`, `1.9`, and pin `1.96.1` vs `1.96.2` all trip it.

Note for review: if Dependabot later bumps the pinned Alpine tag, that PR's Docker build is expected to fail until `rust-toolchain.toml` moves with it — that is the intended signal, but it surfaces at deploy time rather than in PR CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
         uses: actions/cache@v6.1.0
         with:
           path: native/*/*.node
-          key: native-modules-${{ runner.os }}-${{ hashFiles('native/**') }}
+          key: native-modules-${{ runner.os }}-${{ hashFiles('native/**', 'rust-toolchain.toml') }}
 
       - uses: pnpm/action-setup@v6.0.10
         if: steps.native-cache.outputs.cache-hit != 'true'
@@ -150,6 +150,13 @@ jobs:
       - name: Run clippy
         if: ${{ !cancelled() }}
         run: ./scripts/cargo-native.sh clippy --workspace --all-targets -- -D warnings
+
+      # The stage is just `apk add rust` plus the check, so this costs a pull
+      # and no cargo build. Without it, a base-image bump that moves Alpine's
+      # rustc off the pin is green here and red in the deploy build.
+      - name: Check the pin still matches the image's rustc
+        if: ${{ !cancelled() }}
+        run: docker build --target rust-base .
 
   unit-tests:
     name: Unit Tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,8 @@ jobs:
     steps:
       - uses: actions/checkout@v7.0.1
 
+      # No toolchain setup step: the runner's rustup installs the version
+      # rust-toolchain.toml pins on the first cargo call.
       - uses: Swatinem/rust-cache@v2.9.2
         with:
           # Without this a red run saves nothing, so a PR iterating on a Rust
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,9 +38,9 @@ export PATH="$PWD/.node26/bin:$PATH"
 
 ## Rust Version
 
-Pinned in `rust-toolchain.toml`, which explains what it must stay equal to and
-when to bump it. Read it before changing the Rust toolchain or the Dockerfile's
-`native-builder` base image.
+Pinned to the rustc Alpine ships in the Dockerfile's `rust-base` stage
+(`rust-toolchain.toml`, which says why). Bump the two together — CI fails
+otherwise.
 
 ## Commands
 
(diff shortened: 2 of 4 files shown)


brendanlong:
Follow-up commit after review (f6a6b71).

The review found the drift this pin guards against arriving through a door CI doesn't watch: `.github/dependabot.yml` enables weekly `docker` updates, so when Alpine 3.25 lands Dependabot will propose `node:26-alpine3.24 → 3.25`; nothing in CI builds the image, so that PR would merge green and the *deploy* build would break.

- Split the toolchain into its own `rust-base` stage, and `rust-checks` now runs `docker build --target rust-base .` — a pull plus `apk add rust`, no cargo build. Verified locally with podman: the stage builds, and with the channel set to a mismatched value the build fails with `Alpine ships rustc 1.96.1 but rust-toolchain.toml pins 1.80. Set channel = "1.96" there …`.
- `build-native`'s cache key now hashes `rust-toolchain.toml` too. It previously keyed on `native/**` only, so on merge it would have restored `.node` files built by the *old* floating stable and handed them to the unit/integration/e2e jobs.
- The failure message suggests the minor version (`${actual%.*}`); echoing the full patch version back would have replaced the patch-tolerant pin with an exact one that the next Alpine rust patch bump breaks.
- `profile = "minimal"` in `rust-toolchain.toml` — neither job caches `~/.rustup`, so every run was downloading `rust-docs` just to use clippy.

Also worth recording from the review: `node:26-alpine` and `node:26-alpine3.24` are the same digest today, so the pin is currently a no-op relative to production, and the first CI run confirmed `rust-checks` compiled with 1.96.1 via the pin.

— Claude

PR #1585, opened 2026-09-08
Sweep for production-unreachable code with knip (#1551)

### LINKED ISSUE #1551: knip cannot see production-dead code that its own tests keep alive
`pnpm knip` passes cleanly, but a substantial amount of `src/` is unreachable from production and stays "used" only because its unit tests are knip entry points.

The largest instance — a closed 4-module TTS provider subtree of ~1050 lines (`tts-provider-factory`, `fallback-tts-provider`, `browser-tts-provider`, `voice-download`) — is being deleted in a separate PR. It survived because its three test files import it; nothing in `src/` does. The live path is `ArticleNarrator` (which talks to `speechSynthesis` directly) plus `piper-tts-provider` (which does its own downloading).

The deletion PR does not address the tail, which has the same cause:
- `ArticleNarrator.getRate` / `getPitch` / `getCurrentParagraphIndex` / `getCurrentParagraphText` / `jumpToParagraph`
- `StreamingAudioPlayer.currentSentenceEndCallback` / `getPosition` / `isSentenceBuffered`
- `useEnhancedVoices.refreshStatus`
- `createMemoizedAddParagraphIds`
- `splitIntoSentencesWithInfo`

Also in this class, found separately: `src/server/feed/entry-processor.ts`'s `processEntry` (a near-exact duplicate of the live `processEntryWithCache`, called only from tests — so those tests assert against dead code rather than the live path) and `generateEntrySummary` (dead, and encoding a *different* summary rule than the live `cleanEntryContent`, so its tests document behavior the app does not have).

The fix is a mechanism rather than another round of deletions: a knip config change, or a periodic "reachable from `src/` entry points only" sweep, so this class shows up on its own.

Found during a frontend review and a feed review.

— Claude

brendanlong:
Fixes #1551.

## The mechanism

`pnpm knip` treats every test file as an entry point, so anything a test imports looks used — which is how a ~1000-line TTS provider subtree survived until a human review found it.

`pnpm knip:production` (new, gating CI alongside `pnpm knip`) runs the same analysis from the entry points that actually ship: the Next app plus `scripts/{server,worker,discord-bot,migrate}.ts` and the MCP binary. `ignoreExportsUsedInFile` keeps the common "exported so a test can reach it, but called by its own module" seam quiet — `isPrivateAddress`, `calculateJitter` and friends aren't reported — so what's left is genuinely unreachable. An export that must stay for tests carries a `@testonly` JSDoc tag, which the sweep skips (3 of them today: two module-state resets and the demo article id list).

It's a separate config file rather than flags on the existing one because the narrower project scope and `ignoreExportsUsedInFile` would weaken the default sweep, which should keep reporting an export nothing imports at all.

## The 21 findings it reported

Deleted in the first commit. Beyond the tail listed in the issue (`ArticleNarrator.getRate`/`getPitch`/`getCurrentParagraphIndex`/`getCurrentParagraphText`/`jumpToParagraph`, `StreamingAudioPlayer.currentSentenceEndCallback`/`getPosition`/`isSentenceBuffered`, `useEnhancedVoices.refreshStatus`, `createMemoizedAddParagraphIds`, `splitIntoSentencesWithInfo` — the class members by hand, since knip can't see them):

- **Wrappers over the live path**: `feed/opml.ts`'s sync `parseOpml`, `feed/parser.ts`'s `parseFeedWithFormat` + `UnknownFeedFormatError` re-export, `scheduling.getNextFetchTime`, `cimd.parseClientMetadataDocument`, `narration/errors`' `getVoiceErrorMessage`/`isVoiceErrorRetryable`.
- **Duplicates of a live implementation**: `google-reader/id.parseFeedStreamId` (the API parses `feed/{n}` in `streams.parseStreamId`), `saved.createUploadedArticle` (a third upload path next to `createSavedFromUpload`/`uploadArticle`).
- **Never-used**: `int64ToShortHex`, `isApplePrivateRelayEmail`, the `Card{Header,Title,Description,Body,Footer}` sub-components, and 5 exported types.
- **Test-owned query helpers** moved into the test files that assert with them: `getJob`, `listJobs`, `findEntryByGuid`.

`processEntry`/`generateEntrySummary` from the issue are already gone.

Where a test covered real behavior through the dead lookalike, it now goes through the live path instead of losing coverage. The CIMD one was the sharpest: `parseClientMetadataDocument` mapped a non-JSON body to "rejected", while the live fetch treats it as a transport failure and falls back to the pinned document — so `it("rejects non-JSON bodies (e.g. a Cloudflare challenge page)")` was asserting the opposite of what the app does.

## Verification

`pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm knip`, `pnpm knip:production` all clean; `pnpm test:unit` (3137) and `pnpm test:integration` (818) pass.


brendanlong:
Post-review note, one thing worth flagging rather than fixing here.

Deleting `parseClientMetadataDocument` removes the only test that touched a non-JSON CIMD body — and that test asserted the opposite of production. The live path (`src/server/oauth/service.ts:253-256`) maps a `JSON.parse` failure to `{kind: "transport-failure"}`, which `selectClientMetadata` turns into the **pinned** document rather than a rejection. That branch of `selectClientMetadata` is covered directly (`tests/unit/oauth-cimd.test.ts`), but the "body isn't JSON ⇒ transport-failure" mapping lives inside the private `fetchClientMetadataLive` and has never been tested. So this PR doesn't lose real coverage there, but it does make an existing gap visible: whether a Cloudflare challenge page can resurrect a pinned client is security-relevant and currently unpinned by any test. Extracting the response→outcome classification into a pure function would make it testable — happy to do that here or in a follow-up, whichever you prefer.

— Claude

PR #1598, opened 2026-09-10
Revoke other sessions when linking/unlinking OAuth providers (#1560)

### LINKED ISSUE #1560: OAuth link/unlink procedures do not revoke other sessions
PR #1559 fixes `me.setPassword`, which added a credential without revoking other sessions. The same gap exists for the four OAuth provider procedures in `src/server/trpc/routers/auth.ts`, which were outside that PR's scope:

- `linkGoogle` (`:994`)
- `linkApple` (`:1050`)
- `linkDiscord` (`:1117`)
- `unlinkProvider` (`:1166`)

Verified: `grep -n "revokeOtherUserSessions" src/server/trpc/routers/auth.ts` returns nothing — none of the four calls it.

`SECURITY.md` §4 says "Password change (**and any credential change**) must revoke other sessions." Linking a new provider adds a way to log in; unlinking one removes a way to log in. Both read as credential changes to me, and `unlinkProvider` is the sharper case: a user removing a provider because they think that account is compromised does not get the attacker's existing session revoked.

Whether this is intended may hinge on whether these ever run inside an onboarding or account-linking sequence where logging other devices out would be disruptive. I did not trace the callers, so this needs a look before changing anything — filing rather than sending a PR for that reason.

Secondary inconsistency noticed while checking: the three `link*` procedures are `expensiveProtectedProcedure` but `unlinkProvider` is a plain `protectedProcedure`. Possibly deliberate (unlinking makes no outbound call), but worth a glance given it is the one with the security-relevant failure mode.

— Claude

brendanlong:
Fixes #1560.

`SECURITY.md` §4 requires any credential change to revoke the user's other sessions (`revokeOtherUserSessions`), but the four OAuth provider procedures in `src/server/trpc/routers/auth.ts` never called it. Linking a provider adds a way to sign in; unlinking removes one. The sharp case is `unlinkProvider`: a user removing a provider account they believe is compromised kept the attacker's existing session alive.

## What changed

- The revoke lives in the shared `linkOAuthAccount` / `unlinkOAuthAccount` services rather than in the four procedures, so a fourth provider can't forget it.
- It fires only when a credential **actually changed**:
  - a `"linked"` result, **not** an `"updated"` re-link. `linkGoogle` doubles as the incremental-authorization path (granting the Google Docs scope re-links the account the user already has) — revoking there would log every other device out for granting a permission, and it adds no new way in.
  - an unlink that removed a row, not an idempotent no-op unlink. A failed unlink (the "only auth method" guard) throws inside the transaction, so it revokes nothing either.
- The unlink revoke runs after the transaction commits, so it isn't issued on the connection holding the user's `FOR UPDATE` lock.
- `unlinkProvider` is now `expensiveProtectedProcedure`, matching the three link procedures and the password mutations (the secondary inconsistency the issue noted).

## Tests

`tests/integration/password-session-revocation.test.ts` → `credential-session-revocation.test.ts`, now covering all the credential changes in one place. Four new cases: a new link revokes others and keeps the caller's, a re-link does not, `auth.unlinkProvider` revokes, and a no-op unlink does not. The sessions are real rows validated through `validateSession`, so the Redis cache eviction is covered too.

Full integration suite green (875 passed), plus `typecheck`, `lint`, `knip`, `knip:production`.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -116,10 +116,11 @@ Entry bodies, saved articles, and AI summaries are rendered with
 - Session cookie is `HttpOnly`, `Secure` (prod), `SameSite=Lax`; tokens are 32
   random bytes, SHA-256 hashed at rest, never stored raw. **Keep these flags.**
 - Password change (and any credential change) must revoke other sessions
-  (`revokeOtherUserSessions`). That includes linking and unlinking a social
-  provider — each adds or removes a way to sign in — which
-  `src/server/services/oauth-accounts.ts` does for all three providers at once
-  (and says why re-linking an already-linked account doesn't count).
+  (`revokeOtherUserSessions`) in the same transaction as the change. Linking and
+  unlinking a social provider counts — each adds or removes a way to sign in
+  (`src/server/services/oauth-accounts.ts`). The one deliberate exception is the
+  email-match link inside `processOAuthCallback`, which is an ordinary sign-in;
+  it says why at the branch.
 - Password-accepting endpoints are rate-limited per-IP **and** per-account (the
   account bucket degrades to in-memory, not fully open, during a Redis outage).
 - Password-accepting endpoints (tRPC `auth.login`, Google Reader `ClientLogin`,
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -116,7 +116,10 @@ Entry bodies, saved articles, and AI summaries are rendered with
 - Session cookie is `HttpOnly`, `Secure` (prod), `SameSite=Lax`; tokens are 32
   random bytes, SHA-256 hashed at rest, never stored raw. **Keep these flags.**
 - Password change (and any credential change) must revoke other sessions
-  (`revokeOtherUserSessions`).
+  (`revokeOtherUserSessions`). That includes linking and unlinking a social
+  provider — each adds or removes a way to sign in — which
+  `src/server/services/oauth-accounts.ts` does for all three providers at once
+  (and says why re-linking an already-linked account doesn't count).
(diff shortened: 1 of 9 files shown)


brendanlong:
Pushed a second commit after a review pass. Two things worth flagging:

**The browser link flow doesn't go through `linkOAuthAccount`.** The settings "Link" button generates a `mode: "login"` auth URL, and every provider's redirect URI is the API route — so a browser link lands in `processOAuthCallback`'s email-match branch, and the three `link*` procedures are live on the REST/OpenAPI surface only. That branch deliberately does **not** revoke: it is indistinguishable from an ordinary first social sign-in, and revoking there would log every device out of an account whose owner just proved control of its verified email. The reason is now documented at the branch, and `SECURITY.md` no longer overclaims. The unlink half of the fix does cover the real browser path, which is the case the issue called sharp. Filed #1600 for the dead SPA callback page / dead `mode: "link"` branch / stale `.env.example` redirect URI.

**The revoke now runs in the credential change's transaction** (`revokeOtherUserSessions` takes an optional `DbOrTx`). Running it after the commit meant a failed revoke left the credential changed, and — because the revoke is gated on something having actually changed — the retry would find nothing to change and skip it for good.

Also trimmed the rationale down to one statement in the module header, documented that `exceptSessionId` must be a session id (under token auth `ctx.session.session.id` is an `api_tokens` id, which would match nothing and take the caller's own sessions down), and replaced two hand-rolled OAuth-link inserts in tests with a shared factory plus a refused-unlink case.

Integration suite 876 passed locally, plus typecheck/lint/knip/knip:production.

— Claude

PR #1605, opened 2026-09-10
Link a social provider to the signed-in account, not to its email (#1603)

### LINKED ISSUE #1603: Settings "Link" creates a second account when the provider email differs
Split out from #1600 (which is about the dead code) — this is the user-visible consequence. Pre-existing; #1602 doesn't change it.

## Repro

Sign in as `bob@work.com` (email/password account). Settings → Linked Accounts → **Link Google**, and authorize a Google account whose email is `bob@personal.com`.

**Expected:** Google is linked to the `bob@work.com` account.

**Actual:** nothing is linked. Depending on config, either:

- **Signup open for Google:** `processOAuthCallback` finds no user with `bob@personal.com`, so `createUser` makes a **brand-new account**, `createSessionResponse` issues a session cookie for it, and you're redirected to `/complete-signup`. You are now signed in as a different, empty account; the original still has no Google linked, and an orphan account exists.
- **Google is invite-only or not an allowed signup provider:** `createUser` throws `inviteRequired` / `signupProviderNotAllowed`, and you get an invite error from a button that had nothing to do with invites.

It only works when the provider's email already equals the account's email — in which case it takes the "email matches existing user → link" branch (`src/server/auth/oauth/callback.ts:169-203`).

## Why

The "Link" button doesn't use a link path. `LinkedAccounts.tsx` calls `auth.googleAuthUrl`, which hardcodes `mode: "login"` (`src/server/trpc/routers/auth.ts:417-427`), and `getRedirectUri` (`src/server/auth/oauth/config.ts:171-173`) sends every provider to `/api/v1/auth/oauth/{provider}/callback`. So the callback takes its **login** branch and identifies the account by the provider's email, not by your session.

`linkOAuthAccount` (`src/server/services/oauth-accounts.ts`) does the right thing — identifies by session, so the emails need not match, and revokes other sessions on a new link per SECURITY.md §4. It just isn't reachable from a browser; #1600 covers the dead wiring, removed in #1602.

## Fix sketch

Give the link flow its own mode again and land it somewhere that calls `linkOAuthAccount` with the session's user: either a `mode: "link"` branch in the callback route that calls the service directly, or a `linkProvider: true` input on the `*AuthUrl` queries that threads through the Redis state blob. Apple and Discord need the same treatment (Discord has no browser link path at all today — `auth.linkDiscord` has never had a frontend caller).

Worth deciding at the same time whether an unauthenticated visitor hitting the link callback should error rather than fall through to sign-in.

— Claude

brendanlong:
Fixes #1603.

## The bug

Settings → Linked Accounts → **Link** generated a `mode: "login"` auth URL, and every provider's redirect URI points at the same callback route, so a link from Settings landed in the **login** branch — which picks the account from the *provider's* email. Authorizing a Google account whose address differs from your Lion Reader address signed you into (or created) a second, empty account and linked nothing; on an invite-only instance it failed with an invite error instead.

## The fix

`mode` in the per-flow Redis state blob decides what the callback does, and `link` is back:

- **`auth.linkAuthUrl`** — session-only, one procedure for all three providers — mints the auth URL with `mode: "link"` and sets the state binding cookie (#1263; `SameSite=None` for Apple's form_post, matching the sign-in URLs).
- The three callback routes dispatch `mode: "link"` to **`createLinkResponse`**, which resolves the account from the **session cookie** and calls the existing `linkOAuthAccount`. The emails need not match, and a new link revokes the user's other sessions (SECURITY.md §4). Discord gains a browser link path it never had.
- **No session ⇒ an error redirect** (`/login?error=link_requires_login`), never a fallthrough to sign-in — falling through is exactly the account mix-up above.

The settings page already had `?linked=` / `?link_error=` handling (orphaned when #1602 deleted the unreachable SPA callback page); it's reachable again, plus codes for "a different account of this provider is already linked" and for a link that landed but couldn't revoke the other sessions.

Drive-by: the three `create*AuthUrl` functions now take options objects — the Google one had grown to four positional parameters, three of them usually `undefined` at the call sites.

## Tests

- `tests/integration/oauth-link-callback.test.ts` drives the **real callback routes** with stubbed provider endpoints: a Google link whose provider email differs lands on the session's user and creates no stray account; other sessions are revoked while the linking one survives; no session ⇒ `/login?error=link_requires_login` with nothing created; a provider account owned by someone else reports `link_error=already_linked`; Discord links the same way.
- `tests/unit/frontend/components/LinkedAccounts.test.tsx` pins the button to `auth.linkAuthUrl` (and asserts it never calls a sign-in auth URL).
- `pnpm typecheck`, `lint`, `test:unit`, `test:integration`, `knip`, `knip:production` all pass locally.


brendanlong:
Pushed `c4a2b7c5` after a review pass found a flow-breaking bug in the first commit.

**Apple links could never have worked.** `createLinkResponse` read the `session` cookie, which is `SameSite=Lax` — and Apple's callback is a cross-site POST, the very fact that forces the state cookie to `SameSite=None`. The cookie would never arrive, so every Apple link would have ended at `/login?error=link_requires_login`; that's also a regression for the matching-email case the sign-in branch used to handle.

**Fix:** a link flow now records *who it is for* up front. `auth.linkAuthUrl` stores an `OAuthLinkTarget` (user + session) in the per-flow Redis blob, and the callback attaches the provider account to that target — nothing about the account is read from the callback request. That also closes a smaller hole the reviewer raised: a session change inside the flow's 10-minute window (shared browser, second account in another tab) would otherwise have redirected the link to whoever was signed in when it landed.

Carrying the target replaces the `mode: "link"` value, so Apple and Discord no longer need a `mode` in their state blobs at all and Google's keeps exactly its previous meaning.

Also in this commit:
- The target's session must still be live (`isSessionActive`) — adding a way to sign in is a credential change, so a flow whose session was logged out in the meantime is refused, not applied.
- `validateOAuthCallback` rejects a link state outright: the tRPC callback mutations can't honour a target, and falling through would pick the account by email (the original bug). Code and `src/server/auth/CLAUDE.md` now agree.
- Don't echo an unrecognized `?linked=` value into the settings alert; drop the guess about *why* sign-in is needed from the login message.
- Doc dedup: the sign-in-vs-link explanation lives in `src/server/auth/CLAUDE.md`; the route/helper comments point at it.

New tests: all three providers, including **Apple through a real cross-site POST with no session cookie**; the link goes to the user who started the flow even when the browser carries another user's session; a revoked session refuses the link; a link state presented to `auth.googleCallback` is rejected.

Two review findings I deliberately left alone:
- **Rollout window.** In-flight blobs from the old release are handled (no target ⇒ a sign-in), but during a rolling deploy or a rollback an old instance can consume a *new* blob, ignore its target and take the sign-in branch — i.e. #1603 again, for up to the 10-minute state TTL, only for someone who clicks Link mid-deploy. Not worth a two-release dance.
- **Google link mode records only the base scopes.** Reachable only by re-linking an already-linked Google account from a stale tab; it clobbers a recorded Docs grant and costs one extra consent screen on the next private-Doc save. Pre-existing on the sign-in path too.

— Claude

PR #1613, opened 2026-09-16
Manage Mailgun, the registrar, Sentry and healthchecks.io in Terraform

brendanlong:
`terraform/` already owned the DNS that points at these services, but not the
services themselves. Three are adopted here by import, with live values read
from their APIs rather than guessed.

The through-line is one question per provider: **does omitting an attribute
adopt the live value, or assert a default?** Optional+Computed adopts; Optional
with a static default overwrites production. It differs per resource, and it is
what decides whether a short block is safe or a loaded gun.

## Mailgun (`mailgun.tf`) — assert-y, in both directions

`wgebis/mailgun` v0.10.0. The `app.lionreader.com` domain and the one inbound
route. Mailgun is **receive-only** for us — no `MAILGUN_API_KEY` in the app's
secrets, nothing in the codebase sends mail.

- **`wildcard` must be declared.** `RequiresReplace`, defaults to `false`, live
  is `true` — and `true` is the entire reason `in.app.lionreader.com` works,
  since it is not itself a registered Mailgun domain. Omitting it plans a
  *replace*: delete the domain, regenerate the DKIM keypair, break inbound mail.
  Hence `prevent_destroy`.
- **`dkim_selector`, `dkim_key_size`, `force_dkim_authority` must NOT be.** The
  provider never reads them back (`applyDomainResponse` doesn't touch them), so
  post-import they are null in state. All three are `RequiresReplace`, so
  writing down the true live values would destroy the domain to record a fact.

## Registrar (`registrar.tf`)

`aws_route53domains_registered_domain`. `name_server` derives from
`cloudflare_zone.lionreader.name_servers`, so the delegation cannot drift from
what Cloudflare actually serves.

**This is where review earned its keep.** The first version of this file
hardcoded the four `awsdns` nameservers, read from `dig +short NS` — which was a
resolver's *cached pre-cutover answer*. The `.com` parent had already been
delegated to Cloudflare (registry last-changed `2026-09-16T17:39:40Z`), about an
hour before that config was written. Applying it would have issued
`UpdateDomainNameservers` and silently reverted the migration, becoming a hard
outage the moment the runbook's step 5 deletes the old hosted zone. Fixed in
b8ca384; the file now says to ask the parent, not a resolver, because the
failure mode is a confident answer rather than an error.

`transfer_lock` is declared `false` to match live (registry status is `["active"]`,
no `clientTransferProhibited`) so adoption stays a no-op. **Turning it on is
worth doing as its own change** — it is the control that blocks an unauthorized
transfer-out, and it is currently off.

Contacts are undeclared (Optional+Computed → adopted, PII stays out of the repo).

## Sentry (`sentry.tf`)

`jianyuan/sentry`. Project + client key, DSN as a sensitive output so rotation is
a pipe into `flyctl secrets`. Short on purpose: this provider is almost entirely
Optional+Computed, so the minimal block really is a neutral adoption. The file
says so, since a short block is otherwise the shape of a bug in this module.

It also records what the provider *doesn't* reach: `data_scrubber` and
`scrub_ip_addresses` are absent from its schema at both org and project level.
FWIW both are currently off at the org level — Sentry is storing request IPs.

## healthchecks.io (`healthchecks.tf`)

The three dead-man's-switch checks, plus their ping URLs as an output keyed by
the env var each is set as.

This one needed a **read-write** API key, and the reason is the hazard.
`channels` is Optional but *not* Computed and the provider's Read fills it from
live, so an undeclared value applies as **"detach every notification
integration"** — three checks that stay green and silently stop telling anyone
when they go red. A read-only key omits `channels` entirely (and `uuid`, so
there is nothing to import with), so you cannot even see what you would be
detaching.

Declared as one `for_each` rather than three resources on purpose: the attribute
that must not be forgotten is the one all three share, so there is a single
place to forget it. The channel is a data-source lookup guarded by a
`precondition`, because `dataSourceHealthcheckChannelRead` returns an **empty
id rather than an error** when nothing matches, and the provider drops empty
strings from the list — so a lookup miss would arrive as a silent detach.

`grace` is declared for the same class of reason: the provider defaults it to
3600, longer than all three live values (900/1200/1800), so omitting it would
quietly slow every alert down rather than erroring.

## Also

Stale comments in `cloudflare.tf`: it claimed the ingest `MX` records lived in
the dashboard (they are declared twenty lines below), and that proxying the
tracking CNAME would break click/open tracking (both are off; we never send).
The migration runbook is marked as complete through step 4.

## Not verified here

- **The registrar.** This machine's IAM user has no `route53domains:*`, so
  `auto_renew` and the four privacy flags are provider defaults, not reads. Run
  the `get-domain-detail` and `list-tags-for-domain` commands in the runbook and
  reconcile before the first apply. (`tags` is Optional with no default on a
  transparently-tagged resource — undeclared, it plans as empty and applies
  `UntagResource`.)
(healthchecks.io **is** now included — see below.)


brendanlong:
Two things I found while reading live config that are **not** fixed in this PR,
because neither should ride along with an adoption commit:

**1. `lionreader.com` has no transfer lock.** The registry status list is
`["active"]` — no `clientTransferProhibited`. That is the control that blocks an
unauthorized transfer-out, and it is free. It's a one-line change once this
merges:

```diff
-  transfer_lock = false
+  transfer_lock = true
```

The comment in `registrar.tf` points at this, and `transfer_lock = false` is
declared purely so the adoption plan stays a no-op.

**2. Sentry is storing request IP addresses.** `scrub_ip_addresses` is `false`
and `data_scrubber` is `false` at the org level (`brendan-long`); the project has
`dataScrubber: true` but `scrubIPAddresses: false`. Worth a look given the
privacy policy's "no third-party tracking" framing. The Terraform provider has
no schema for either at any level, so this stays a dashboard change — which is
why `sentry.tf` says so in a comment rather than silently implying coverage.

**And why healthchecks.io isn't here:** the key was read-only, and
healthchecks.io withholds both `uuid` (needed for the import id) and `channels`
from read-only keys. `channels` is `Optional` but **not** `Computed`, and the
provider's Read populates it from live — so importing without declaring it plans
a removal and applies an empty channel list, clearing every notification
integration. The result is three checks that stay green and silently stop
alerting anyone, which is worse than not having them. Happy to do it with a
read-write key, though the provider has been stale since 2025-04 and the API is
a stable alternative.

— Claude

brendanlong:
A third finding, surfaced while fixing the DSN rotation recipe in the docs —
**client-side Sentry is off in production.**

`src/instrumentation-client.ts:17` gates on `process.env.NEXT_PUBLIC_SENTRY_DSN`,
which Next.js inlines at build time. It reaches the build only via
`[build.args]` in `fly.toml:31`, where it is **commented out**, and
`.github/workflows/deploy.yml:69-72` passes `--build-arg` for
`GIT_COMMIT_SHA` / `BUILD_TIME` / `NEXT_PUBLIC_BUILD_TIME` only. So the browser
bundle ships with no DSN and reports nothing.

Meanwhile `flyctl secrets list -a lion-reader` shows a deployed
`NEXT_PUBLIC_SENTRY_DSN` with the same digest as `SENTRY_DSN`. It's inert —
`fly.toml:22-23` already warns that a runtime var has no effect for these. So it
looks configured from the secrets list while doing nothing, which is the part
worth knowing.

Server-side Sentry is fine: `SENTRY_DSN` is read at runtime and is set.

Not fixed here — turning browser error reporting on is a decision, not a typo,
and it's a `fly.toml` change rather than a Terraform one. Two options if you want
it: uncomment `fly.toml:31` with the literal DSN (it's not a secret — it ships in
the bundle by design), or drop the inert secret and leave client reporting off
deliberately. Happy to do either.

— Claude

PR #1614, opened 2026-09-16
Lock the domain against transfer, and stop sending IPs to Sentry

brendanlong:
Stacked on #1613 — **base is `claude/323db231-...`, not `master`**, so merge that
one first. The two findings this fixes came out of reading the live config while
adopting it.

## Sentry setting: done

"Prevent Storing of IP Addresses" is now **on** at the org level
(`scrubIPAddresses: true`, verified via the API), which Sentry applies whether
it is set org-wide or per-project. So the privacy-policy sentence in this PR is
accurate and nothing here is blocked.

It is deliberately not in Terraform, because it cannot be: the provider has no
attribute for it on any of its 24 resources, and the maintainer's own feature
request for exactly these settings
([#248](https://github.com/jianyuan/terraform-provider-sentry/issues/248)) was
closed `not_planned` on 2026-09-01. `sentry.tf` records that it must stay on.

Unrelated leftover: org-level `dataScrubber` is still `false`. The project
overrides it to `true`, so nothing is exposed today — it only sets the default
for new projects.

## Transfer lock

`transfer_lock = false` → `true`. The registry status list was `["active"]` with
no `clientTransferProhibited`, so the lock was simply off. It was declared
`false` in the adoption commit so that import stayed a no-op — the right default
for an adoption, the wrong one to leave sitting there.

This is the first real change through the registrar resource, so the plan is
worth a careful read: it should touch `transfer_lock` and nothing else.

## Sentry IPs

**The first attempt at this was wrong and is reverted in `f8e8adc`** — see the
thread. Setting `dataCollection: { userInfo: false }` changed `userInfo` by
nothing (already the resolved value) while flipping six other categories from
the conservative baseline to the permissive one, because
`resolveDataCollectionOptions` re-bases to Sentry's `DEFAULTS` the moment
`dataCollection` is non-null — for every key, not just the one you set. It would
have started sending unfiltered `x-forwarded-for` and Anthropic prompts.

What's here now:

- **No `dataCollection` option.** The baseline it would replace is already the
  conservative one.
- **The posture is asserted in a test** instead, read back through the client's
  own `getDataCollectionOptions()`. I verified it fails on the exact regression
  above, and it will fail again when v11 removes the `sendDefaultPii` bridge.
  Watching the value is the right move; setting it was not.
- **The actual leak is fixed.** Sentry's HTTP server instrumentation sets
  `http.client_ip` from `x-forwarded-for` and `net.peer.ip` from the socket on
  every server transaction span, **ungated** by `dataCollection` — the sibling
  `httpHeadersToSpanAttributes` call is gated, these two are not — and
  "Prevent Storing of IP Addresses" only covers `user.ip_address`. At a 10%
  trace rate that's a real flow of user IPs. Now stripped in
  `beforeSendTransaction`. `net.host.ip` is kept: that's our own address.

Corrections to claims I made earlier in this PR: the browser SDK **does** signal
`infer_ip: "never"` to Relay when `userInfo` is false, so "the SDK cannot do this
for us" was wrong for the browser; and the privacy policy no longer says "or
other identifying details", which was broader than anything I verified (we
attach no user identity to Sentry at all — there are no `Sentry.setUser` calls).

## Checks

`pnpm typecheck`, `pnpm test:unit` (3238 passed), `eslint`, `knip:production`,
`terraform validate` + `fmt` all clean. I have no `route53domains` access from
here, so I have not planned the `transfer_lock` change.


brendanlong:
**Converting to draft — don't merge this yet. The Sentry half of it is a net
privacy regression, and I introduced it.**

I verified with the SDK's own resolver rather than by reading:

```
node -e "resolveDataCollectionOptions({})  vs  resolveDataCollectionOptions({dataCollection:{userInfo:false}})"

same     userInfo
CHANGED  cookies            {"deny":["forwarded","-ip","remote-","via","-user"]} -> true
CHANGED  httpHeaders        {request:{deny:[...]},response:{deny:[...]}}         -> {request:true,response:true}
CHANGED  httpBodies         []                                                   -> [4 categories]
CHANGED  urlQueryParams     {"deny":[...]}                                       -> true
CHANGED  genAI              {inputs:false,outputs:false}                         -> {inputs:true,outputs:true}
CHANGED  databaseQueryData  false                                                -> true
CHANGED  frameContextLines  7                                                    -> 5
```

`userInfo` is **unchanged** — it was already `false`, so the change bought
nothing — while six other categories flipped from the conservative baseline to
the permissive one. I spotted the `dataCollection != null` re-baselining trap
and then only applied it to the one field I was looking at.

Two of those are concretely harmful here:

- **`httpHeaders` deny list → `true`.** That deny list (`forwarded`, `-ip`,
  `remote-`, `via`, `-user`) is exactly what currently `[Filtered]`s
  `x-forwarded-for`, `fly-client-ip`, `x-real-ip` and `via` out of span
  attributes. Removing it sends them verbatim — the precise opposite of this
  PR's goal.
- **`genAI.inputs/outputs` → `true`.** The Anthropic integration is default-on
  in the auto-performance set, and we call Anthropic server-side with article
  text as the prompt. That would start shipping article content to Sentry,
  contradicting the sentence directly above the one this PR adds.

Two more corrections to my own claims:

- **The browser SDK *does* tell Sentry not to infer the IP.** `@sentry/browser`'s
  client sets `_metadata.sdk.settings.infer_ip = userInfo ? "auto" : "never"`,
  with the comment "Only allow IP inferral by Relay if the user opted in via
  dataCollection". My comment saying the SDK can't do this was wrong for the
  browser.
- **There's a server-side IP leak neither half of this PR touches.**
  `http.client_ip` is set from `x-forwarded-for` on every server transaction
  span in `server-subscription.js:205`, **ungated** by `dataCollection` (the
  `httpHeadersToSpanAttributes` call on the next line is gated; this line
  isn't), alongside `net.peer.ip`. With tracing at 10% in prod, that is a real
  IP flow, and `scrubIPAddresses` won't catch it — that targets
  `user.ip_address`, not an arbitrary span attribute.

So the privacy-policy sentence is still not true, for a different reason than I
said. Reworking: drop the `dataCollection` option entirely (the baseline it
replaces is already the conservative one), guard the v11 default-flip with a
test instead of runtime config, and strip the IP span attributes in
`beforeSendTransaction` — which is the part that actually makes the claim true.

— Claude

